Issue #2, opened 2016-04-03
Refer to RFC 7807 for the common error object


ePaul:
I just noticed (looking for the April fools RFCs) that there is a new [RFC 7807](http://tools.ietf.org/html/rfc7807) defining a JSON error object, which seems to be based on the draft to which we are referring in the common data objects section.

That section should be adapted to refer to this RFC instead.

The RFC defines it as `application/problem+json`, while our guidelines are using `application/x.problem+json`. I guess we need some guideline about transitioning between both, because old clients might expect the content type with `x` (and old servers send it with `x`), while new clients would expect it without `x` (and new servers send it without `x`).

hjacobs:
:+1:

mostlymagic:
@ePaul rather than filing an issue, why don't you submit a pull request?

ePaul:
> @ePaul <https://github.com/ePaul> rather than filing an issue, why don't
you submit a pull request?

I noted that on a weekend, when I don't work, and filed the issue so I
don't forget it. I will create a pull request as soon as I find some time
to do it (if noone else does it earlier).

PR #3, opened 2016-04-05
Issue #2: refer to RFC 7807 instead of the (long expired) Internet Draft

ePaul:
Fixes Issue #2:
Refer to RFC 7807 instead of the (long expired) Internet Draft, with some explanation.
I also added a guidance about `application/x.problem+json` vs. `application/problem+json`.
I also removed the link to the Pennybags documentation, which is not accessible without
authentication.

@whiskeysierra Do you want me to add something from that documentation? (I'm at home right now and can't see it.)

### DIFF
--- a/common-data-objects/CommonDataObjects.md
+++ b/common-data-objects/CommonDataObjects.md
@@ -90,15 +90,22 @@ cases. Use it in your APIs — and compatible extend it if necessary for your AP
 
 ## {{ book.must }} Use Common Error Return Objects
 
-`application/x.problem+json` is an  example of a common error return object derived from an
-[Internet draft specification](https://tools.ietf.org/html/draft-nottingham-http-problem-07) for
-error objects that have never reached registered
-[media type status](http://www.iana.org/assignments/media-types/media-types.xhtml#application).
-The media type name is prefixed with “x.” to denote its status. Hint: APIs may define custom problem types with extension properties, according to their specific needs.
+[RFC 7807](http://tools.ietf.org/html/rfc7807) defines the media type `application/problem+json`
+for a common error return object (and the format of the contained JSON object).
+Operations should return that (together with a suitable status code) when any problem
+occurred during processing and you can give more details than the status code itself
+can supply, whether it be caused by the client or the server (i.e. both for 4xx or 5xx errors).
+
+A previous version of this guideline (before the publication of that RFC and the
+registration of the media type) told to return `application/x.problem+json` in these
+cases (with the same contents).
+Servers for APIs defined before this change should pay attention to the `Accept` header sent
+by the client and set the `Content-Type` header of the problem response correspondingly.
+Clients of such APIs should accept both media types.
 
 APIs may define custom problems types with extension properties, according to their specific needs.
-[Go here](https://docs.pennybags.zalan.do/problems) for a more detailed explanation
-(IAM login required).
+
+Here is the definition in Open API 2.0 YAML format in the case you don't need any extensions:
 
     Problem:
      type: object


whiskeysierra:
This is basically what it says on that page. Not sure if there is something worth keeping.

# Problems

<html>
<blockquote>
<p>something that is difficult to deal with : something that is a source of trouble, worry, etc.</p>
<footer>
  <a href="http://www.merriam-webster.com/dictionary/problem"><cite title="Problem">Merriam-Webster</cite></a>
</footer>
</blockquote>

<blockquote>
<p>A matter or situation regarded as unwelcome or harmful and needing to be dealt with and overcome.</p>
<footer>
  <a href="http://www.oxforddictionaries.com/definition/english/problem"><cite title="Problem">Oxford Dictionaries</cite></a>
</footer>
</blockquote>
</html>

## application/problem+json

<html>
<blockquote>
<p>
   HTTP status codes are sometimes not sufficient to convey enough 
   information about an error to be helpful.  While humans behind Web 
   browsers can be informed about the nature of the problem with an
   HTML response body, non-human consumers of so-called <em>HTTP APIs</em>
   are usually not.
</p>

<p>
   [application/problem+json] defines simple JSON and XML document formats to 
   suit this purpose.  They are designed to be reused by HTTP APIs, 
   which can identify distinct <em>problem types</em> specific to their needs.
</p>
<footer>Mark Nottingham and Erik Wilde in
<a href="https://tools.ietf.org/html/draft-nottingham-http-problem-07"><cite title="Problem Details for HTTP APIs">Problem Details for HTTP APIs</cite></a>
</footer>
</blockquote>
</html>

[Problem Details for HTTP APIs](https://tools.ietf.org/html/draft-nottingham-http-problem-07) is an Internet-Draft for a common and extensible format to transport errors over HTTP. Even though it's expired as of January 29, 2015 it's the closest to an official standard one can find and it has a bunch of nice properties, one of which is the ability to define custom extensions using the ''type'' property as a discriminator which is a perfect match with inheritance in [Swagger](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#composition-and-inheritance-polymorphism) and Jackson's [Polymorphic Type Handling](http://wiki.fasterxml.com/JacksonPolymorphicDeserialization) - making it extremely pleasant to de/serialize custom and generic problem types in resource and [gateway](http://martinfowler.com/articles/microservice-testing/#anatomy-modules) implementations.

## Example

```yaml
{
  "type": "https://docs.pennybags.zalan.do/problems/insufficient-credit-account",
  "status": 422,
  "title": "Insufficient Credit Account",
  "detail": "Current balance is EUR 0.00 but needed EUR 20.00 or more.",
  "balance": {
    "amount": 0.00,
    "currency": "EUR"
  },
  "debit": {
    "amount": -20.00,
    "currency": "EUR"
  }
}
```

## Schema

```yaml
Problem:
  type: object
  properties:
    type:
      type: string
      format: uri
      description: |
        An absolute URI that identifies the problem type.  When dereferenced, it SHOULD 
        provide human-readable documentation for the problem type (e.g., using HTML).
      example: http://httpstatus.es/503
    title:
      type: string
      description: A short, human-readable summary of the problem type. It SHOULD NOT
      change from occurrence to occurrence of the problem, except for purposes of localisation.
      example: Service Unavailable
    status:
      type: integer
      format: int32
      description: |
        The HTTP status code generated by the origin server for this occurrence of the problem.
      example: 503
    detail:
      type: string
      description: |
        An human readable explanation specific to this occurrence of the problem.
      example: Connection to database timed out
    instance:
      type: string
      format: uri
      description: |
        An absolute URI that identifies the specific occurrence of the problem. 
        It may or may not yield further information if dereferenced.
  required:
    - type
    - title
    - status
```

## Generic problems

There are status codes which convey actually enough information to the client to be handled correctly. Examples would include ''404 Not Found'', ''412 Precondition Failed'' and ''415 Unsupported Media Type''. Then there are status codes that wouldn't justify a specific, custom problem type but may you still want to attach some additional details, e.g. ''414 Request-URI Too Long''. ''type'', ''title'' and ''status'' in this case can all be infered by the status code alone. The ''type'' property points to the specification of the status code over at [http://httpstatus.es/] and ''title'' contains the reason phrase as defined by [RFC 7231](https://tools.ietf.org/html/rfc7231#section-6.1).

```yaml
{
  "type": "http://httpstatus.es/414",
  "title": "Request-URI Too Long",
  "status": 414,
  "detail": "Maximum length is 8,192 characters."
}
```

ePaul:
The RFC tells to use type `about:blank` (which is also the default value) for `type` if your type has not more specific information than the status code itself. This would mean not to use the `httpstatus.es` links anymore in those cases.

whiskeysierra:
> This would mean not to use the httpstatus.es links anymore in those cases.

Exactly. I already proposed changes to our problem library to implement this. https://github.com/zalando/problem/pull/20

ePaul:
So I guess we can leave it the way it currently is.

PR #5, opened 2016-04-07
adding content location header description

peletomi:
I would like to add this section to provide a background, why we require to add a content location header in the response of mutating requests. There were quite a few comments and questions regarding this during our reviews.

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -11,5 +11,6 @@
 * [Hypermedia](hyper-media/Hypermedia.md)
 * [Data Formats](data-formats/DataFormats.md)
 * [Common Data Objects](common-data-objects/CommonDataObjects.md)
-* [Common Headers](common-headers/CommonHeaders.md)
+* [Common Headers](headers/CommonHeaders.md)
+* [Proprietary Headers](headers/ProprietaryHeaders.md)
 * [API Discovery](api-discovery/ApiDiscovery.md)
--- a/headers/CommonHeaders.md
+++ b/headers/CommonHeaders.md
@@ -1,0 +1,19 @@
+# Common Headers
+
+This section describes a handful of headers, which we found raised the most questions in our daily usage.
+
+## Content-Location
+
+This header is used in the response of either a GET, HEAD, PUT, POST or PATCH request. In the case of the GET requests it
+points to a location where an alternate representation of the entity in the response body can be found. In this case one
+has to set the Content-Type header as well. For example:
+
+    GET /product-images/123
+    Content-Type: image/png
+    Content-Location: /product-images/123?format=png
+
+In the case of mutating HTTP methods the Content-Location header is used, when there is a response body. If the Content-Location
+is the same than the Location header it means, that the returned body is indeed the current representation of the entity, making
+a subsequent GET operation from the client side not necessary.
+
+More details in [rfc7231](https://tools.ietf.org/html/rfc7231#section-3.1.4.2)
--- a/headers/CommonHeaders.md
+++ b/headers/CommonHeaders.md
@@ -4,16 +4,29 @@ This section describes a handful of headers, which we found raised the most ques
 
 ## Content-Location
 
-This header is used in the response of either a GET, HEAD, PUT, POST or PATCH request. In the case of the GET requests it
-points to a location where an alternate representation of the entity in the response body can be found. In this case one
+This header is used in the response of either a successful read (GET, HEAD) or successful write operation (PUT, POST or PATCH).
+
+In the case of the GET requests it points to a location where an alternate representation of the
+entity in the response body can be found. In this case one
 has to set the Content-Type header as well. For example:
 
-    GET /product-images/123
-    Content-Type: image/png
-    Content-Location: /product-images/123?format=png
+```http
+GET /products/123/images HTTP/1.1
 
-In the case of mutating HTTP methods the Content-Location header is used, when there is a response body. If the Content-Location
-is the same than the Location header it means, that the returned body is indeed the current representation of the entity, making
-a subsequent GET operation from the client side not necessary.
+HTTP/1.1 200 OK
+Content-Type: image/png
+Content-Location: /products/123/images?format=raw
+```
+
+In the case of mutating HTTP methods, the Content-Location header can be used when there is a response body,
+and then it indicates that the included response body can be found at the location indicated in the header.
+
+If the header value is the same as the location of the created resource (as indicated by the Location header
+after POST) or the modified resource (as indicated by the request URI after PUT / PATCH), then the returned
+body is indeed the current representation of the entity making a subsequent GET operation from the client side
+not necessary.
+
+If your API returns the new representation after a PUT, PATCH, or POST you should include the Content-Location header
+to make it explicit, that the returned resource is an up-to-date version.
 
 More details in [rfc7231](https://tools.ietf.org/html/rfc7231#section-3.1.4.2)
--- a/http/Http.md
+++ b/http/Http.md
@@ -90,7 +90,7 @@ different HTTP methods on resources.
 | Code | Meaning | Methods |
 | --   | --      | --                 |
 | 200  | OK - this is the standard success response | All |
-| 201  | Created - Returned on successful entity creation. You are free to return either an empty response or the created resource in conjunction with the Content-Location header. (More details found [here](../headers/CommonHeaders.html).) *Always* set the Location header. | POST, PUT |
+| 201  | Created - Returned on successful entity creation. You are free to return either an empty response or the created resource in conjunction with the Content-Location header. (More details found in the [Common Headers section](../headers/CommonHeaders.html).) *Always* set the Location header. | POST, PUT |
 | 202  | Accepted - The request was successful and will be processed asynchronously. | POST, PUT, DELETE, PATCH |
 | 204  | No content - There is no response body | PUT, DELETE |
 
--- a/http/Http.md
+++ b/http/Http.md
@@ -90,7 +90,7 @@ different HTTP methods on resources.
 | Code | Meaning | Methods |
 | --   | --      | --                 |
 | 200  | OK - this is the standard success response | All |
-| 201  | Created - Returned on successful entity creation. You are free to return either an empty response or the created resource in conjunction with the Content-Location header.  Always set the Location header. | POST, PUT |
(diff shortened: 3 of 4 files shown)


http/Http.md
```diff
@@ -90,7 +90,7 @@ different HTTP methods on resources.
 | Code | Meaning | Methods |
 | --   | --      | --                 |
 | 200  | OK - this is the standard success response | All |
-| 201  | Created - Returned on successful entity creation. You are free to return either an empty response or the created resource in conjunction with the Content-Location header.  Always set the Location header. | POST, PUT |
+| 201  | Created - Returned on successful entity creation. You are free to return either an empty response or the created resource in conjunction with the Content-Location header. (More details found [here](../headers/CommonHeaders.html).) *Always* set the Location header. | POST, PUT |
```
fmueller: After renaming file to ProprietaryHeaders the link should be broken.

headers/CommonHeaders.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Content-Location
+
+This header is used in the response of either a GET, HEAD, PUT, POST or PATCH request. In the case of the GET requests it
+points to a location where an alternate representation of the entity in the response body can be found. In this case one
+has to set the Content-Type header as well. For example:
+
+    GET /product-images/123
+    Content-Type: image/png
+    Content-Location: /product-images/123?format=png
```
whiskeysierra: Shouldn't we promote content negotiation for this?

headers/CommonHeaders.md
```diff
@@ -0,0 +1,19 @@
...
+
+This section describes a handful of headers, which we found raised the most questions in our daily usage.
+
+## Content-Location
+
+This header is used in the response of either a GET, HEAD, PUT, POST or PATCH request. In the case of the GET requests it
+points to a location where an alternate representation of the entity in the response body can be found. In this case one
+has to set the Content-Type header as well. For example:
+
+    GET /product-images/123
```
whiskeysierra: Github support `http` style formatting.

headers/CommonHeaders.md
```diff
@@ -0,0 +1,19 @@
...
+    Content-Type: image/png
+    Content-Location: /product-images/123?format=png
+
+In the case of mutating HTTP methods the Content-Location header is used, when there is a response body. If the Content-Location
+is the same than the Location header it means, that the returned body is indeed the current representation of the entity, making
```
fmueller: small typo: then not 'than' and -it

fmueller:
+1

headers/CommonHeaders.md
fmueller: Well, this is the typical example on the web for this. Maybe /product/123/images?format=png would be better in this case.

whiskeysierra: > Well, this is the typical example on the web for this.

I know, but most examples on the web are not the best. Another example, the HATEOAS example on wikipedia: https://en.wikipedia.org/wiki/HATEOAS

headers/CommonHeaders.md
whiskeysierra: > Github support http style formatting.

I'd write it as:

```http
GET /products/123/images HTTP/1.1

HTTP/1.1 200 OK
Content-Type: image/png
Content-Location: /products/123/images?format=raw
```

fmueller: +1 for this example

headers/CommonHeaders.md
ePaul: My understanding of Content-Location with Content-Negotiation:

When the variant which was retrieved with (server-side) content negotiation also has an individual URI (where you always get this variant, independent of headers), the Content-Location header then can indicate this URI.

whiskeysierra: > When the variant which was retrieved with (server-side) content negotiation also has an individual URI

Kind of violates the rule that there should be *one* canonical URL for one resource.

http/Http.md
ePaul: I would write it as `More details found in the [Common Headers section](../headers/CommonHeaders.html).`
"Here" is a bad link text.

SUMMARY.md
```diff
@@ -11,5 +11,6 @@
 * [Hypermedia](hyper-media/Hypermedia.md)
 * [Data Formats](data-formats/DataFormats.md)
 * [Common Data Objects](common-data-objects/CommonDataObjects.md)
-* [Common Headers](common-headers/CommonHeaders.md)
+* [Common Headers](headers/CommonHeaders.md)
+* [Proprietary Headers](headers/ProprietaryHeaders.md)
```
ePaul: That might be a different discussion, but what is our stance on ["Cool URIs don't change"](https://www.w3.org/Provider/Style/URI.html)?

For example, this makes http://zalando.github.io/restful-api-guidelines/common-headers/CommonHeaders.html a broken link, introducing a new URI containing this same content.

Is there a way to create a redirect with Gitbook?

headers/CommonHeaders.md
```diff
@@ -0,0 +1,19 @@
+# Common Headers
+
+This section describes a handful of headers, which we found raised the most questions in our daily usage.
+
+## Content-Location
+
+This header is used in the response of either a GET, HEAD, PUT, POST or PATCH request. In the case of the GET requests it
+points to a location where an alternate representation of the entity in the response body can be found. In this case one
```
ePaul: Hmm, "an alternate representation of the entity in the response body" sounds wrong. Or I totally misunderstood the meaning of the header.

As I understand [RFC 7231](https://tools.ietf.org/html/rfc7231#section-3.1.4.2), it should be more like this:

> In the case of GET requests, it can point to an alternate location where the same representation in the response body can be found.

Though maybe we should also add a link to the RFC here?

headers/CommonHeaders.md
ePaul: We have several resources here:

* The image for product 123
    * URI /product-images/123, content-negotiated → response can have a Content-Location header with one of the next two URIs.
* The PNG image for product 123
    * URI /product-images/123?format=png
* The JPEG image for product 123
    * URI /product-images/123?format=jpeg

Each of these three resources has one canonical URI.

whiskeysierra: > We have several resources here:

I disagree. It's **one** resource with multiple representations. A resource deserves it's own URL, a representation doesn't.

headers/CommonHeaders.md
```diff
@@ -0,0 +1,19 @@
...
+has to set the Content-Type header as well. For example:
+
+    GET /product-images/123
+    Content-Type: image/png
+    Content-Location: /product-images/123?format=png
+
+In the case of mutating HTTP methods the Content-Location header is used, when there is a response body. If the Content-Location
+is the same than the Location header it means, that the returned body is indeed the current representation of the entity, making
+a subsequent GET operation from the client side not necessary.
```
ePaul: The first sentence sounds a bit misleading. If the response body is just something like "Your request succeeded and took 32 milliseconds on the server, the created document is at URI XY", there is no point in using a Content-Location header.

Alternative proposal:

> In the case of mutating HTTP methods, the Content-Location header can be used when there is a response body, and then indicates that the included response body can be found at the location indicated in the header.

> If the header value is the same as the location of the created resource (as indicated by the Location header after POST) or the modified resource (as indicated by the request URI after PUT/PATCH), the returned body is indeed the  current representation of the entity, making a subsequent GET operation from the client side not necessary.

> If your API returns the new representation after a PUT/PATCH/POST, you should include the Content-Location header to make this explicit.

headers/CommonHeaders.md
whiskeysierra: > Though maybe we should also add a link to the RFC here?

:+1:

SUMMARY.md
whiskeysierra: > Is there a way to create a redirect with Gitbook?

Good question. I guess we could introduce a new page layout and then using javascript to fake a redirect:

    window.location = "http://www.andereseite.com/";

This won't work for google, but at least users will have a still working link.

headers/CommonHeaders.md
ePaul: It seems like the RFC has this as the only actual use case of `Content-Location` for GET, though. Do you want to say "The RFC is wrong here, don't use Content-Location with GET"?

headers/CommonHeaders.md
ePaul: Actually there is already one at the end.

headers/CommonHeaders.md
whiskeysierra: > It seems like the RFC has this as the only actual use case of Content-Location for GET, though. Do you want to say "The RFC is wrong here, don't use Content-Location with GET"?

Unfortunately the RFC is not very clear here. It states

> The "Content-Location" header field references a URI that can be used
   as an identifier for **a specific resource corresponding to the
   representation** in this message's payload.

It doesn't say that the URI points to a different representation. The example further down is also not 100% clear on that:

> For example, a purchase transaction made via a
      POST request might include a receipt document as the payload of
      the 200 (OK) response; the Content-Location field-value provides
      an identifier for retrieving a copy of that same receipt in the
      future.

It talks about a *purchase transaction* and a *receipt document*. Those sound like related resources to me and one could be seen as a representation of the other, which makes the whole thing kind of vague.

Until now *representation* for me was the message format, e.g. *png*, *jpeg* and *raw*. While the resource, *an image* could be transported in any of those formats. The way this RFC is using the term feels different.

ePaul: The canonical example for how several URIs can point to the same "thing" is a weather service, with resources being e.g.

* the weather now in Berlin
* the weather at 2016-04-07 in Berlin

On 2016-04-07 both would return the same representation, and the first one could return a Content-Location header pointing to the second one.

headers/CommonHeaders.md
peletomi: it is not: if - then, but same - than

headers/CommonHeaders.md
peletomi: I have changed the first sentence a bit.

I have changed the last paragraph to this one (bit of cosmetic fixing).

What I don't like is the: make _this_ explicit part. It would be nice to name _this_ concretely. I would suggest:


> If your API returns the new representation after a PUT, PATCH, or POST you should include the Content-Location header
> to make it explicit, that the returned resource is an up-to-date version.

headers/CommonHeaders.md
ePaul: same as

PR #18, opened 2016-04-11
Improving tenancy header wording

fmueller:
I want to make it more clear that the tenant header should be used by new platform services. Not only that it's currently only used by new platform - also that it should be used when designing new platform services.


whiskeysierra:
:+1:  (as soon as merge conflicts are resolved)

PR #23, opened 2016-04-13
Introducing resources chapter for resource-specific rules

fmueller:
Some editorial changes, suggested by @tfrauenstein 

All resource-related rules are grouped in own chapter.

### DIFF
--- a/resources/Resources.md
+++ b/resources/Resources.md
@@ -1,0 +1,57 @@
+# Resources
+
+## {{ book.must }} Avoid Actions — Think About Resources
+
+REST is all about your resources, so consider the domain entities that take part in web service interaction, and aim to model your API around these using the standard HTTP methods as operation indicators. For instance, if an application has to lock articles explicitly so that only one user may edit them, create an article lock with PUT or POST instead of using a lock action.
+
+Request:
+
+    PUT /article-locks/{article-id}
+
+The added benefit is that you already have a service for browsing and filtering article locks.
+
+## {{ book.should }} Define *useful* resources
+
+As a rule of thumb resources should be defined to cover 90% of all its client's use cases. A *useful* resource should
+contain as much information as necessary, but as little as possible. A great way to support the last 10% is to allow
+clients to specify their needs for more/less information by supporting filtering and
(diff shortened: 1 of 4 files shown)


resources/Resources.md
```diff
@@ -0,0 +1,57 @@
...
+    /carts/1681e6b88ec1/cart-items/1
+
+However, if the resource can be accessed directly via its unique id, then the API should expose it as a top-level resource. For example, customer is a collection for sales orders; however, sales orders have globally unique id and some services may choose to access the orders directly, for instance:
+
+    /customer/1681e6b88ec1
+    /sales-order/5273gh3k525a
+
+## {{ book.should }} Limit of Resources
+
+To keep maintenance manageable, and to avoid violations of the “separation of concern” principle, an API should not expose more than 16 resources. If you exceed 16 resources, first check if you can split them into separate subdomains with distinct APIs.
```
whiskeysierra: I haven't seen that rule before. It's kind of a random magic number, isn't it? Why 16?

fmueller: Right, this sounds strange. Any historical background on this by @tfrauenstein @mostlymagic or @ePaul?

ePaul: I don't remember having seen this (nor the next) rule before.

Looking into the history, this was introduced in commit a68803b1ed ("Merging all pending changes from the internal Google doc") from you. I suppose someone proposed it in the Google doc and it was then simply copied here?

whiskeysierra: > I suppose someone proposed it in the Google doc and it was then simply copied here?

That makes sense.

fmueller: Ok, then I'll drop it.

whiskeysierra:
I created #28 for the discussion/change about the limit.

resources/Resources.md
whiskeysierra: > Ok, then I'll drop it. 

Sorry, I saw that too late.

> Nevertheless we should say sth about separation of concerns and single responsibility of APIs. Do we have a rule for this?

Not yet, but :+1:

mostlymagic: I agree with the rule in general, but I don't think it makes sense to use a hard number

ePaul:
Parent-Issue: #41.

PR #25, opened 2016-04-13
Added rules for how to use PATCH

whiskeysierra:


http/Http.md
```diff
@@ -24,12 +24,15 @@ Be compliant with the standardized HTTP method semantics summarized as follows:
...
-- partial resource representation passed as payload has either resource content type with optional fields (to be
-  updated) or a custom content type that also may include instructions of how to change the resource
-- usually accepted only by single resources, because the semantics for PATCH on a collection resource
-  are very hard to define
 - usually not robust against non existence of the entity
-
+- since implementing PATCH correctly is a bit tricky we strongly suggest to go for the following:
+
+1. use PUT with a custom media type and only if that isn't sufficient anymore
+2. use PATCH with [JSON Merge Patch](https://tools.ietf.org/html/rfc7386), a specialized media type
```
ePaul: That RFC is obsoleted by https://tools.ietf.org/html/rfc7396 (which defines about the same syntax).

I fear that it is difficult to actually describe this in OpenAPI, though (mainly due to the `null` values, which OpenAPI 2.0 doesn't allow, though many implementations ignore that fact).

http/Http.md
```diff
@@ -24,12 +24,15 @@ Be compliant with the standardized HTTP method semantics summarized as follows:
...
 - usually not robust against non existence of the entity
-
+- since implementing PATCH correctly is a bit tricky we strongly suggest to go for the following:
+
+1. use PUT with a custom media type and only if that isn't sufficient anymore
+2. use PATCH with [JSON Merge Patch](https://tools.ietf.org/html/rfc7386), a specialized media type 
+   `application/merge-patch+json` that is a partial resource representation. If updates are too complex to be expressed
+   with a merge patch, you should
+3. use PATCH with [JSON Patch](http://tools.ietf.org/html/rfc6902), a specialized media type 
+   `application/json-patch+json` that includes instructions on how to change the resource
```
ePaul: From an editorial point of view, could you indent this numbered list by four spaces so it gets put under the `since` bullet point?

fmueller:
I really appreciate this proposed change. Can we have one or two more +1 on this, to be able to merge. @ePaul @dehora @tfrauenstein @mostlymagic

http/Http.md
```diff
@@ -21,24 +21,26 @@ Be compliant with the standardized HTTP method semantics summarized as follows:
...
-- usually accepted only by single resources, because the semantics for PATCH on a collection resource
-  are very hard to define
 - usually not robust against non existence of the entity
+- since implementing PATCH correctly is a bit tricky we strongly suggest to go for the following:
+    1. use PUT with a custom media type and only if that isn't sufficient anymore
+    2. use PATCH with [JSON Merge Patch](https://tools.ietf.org/html/rfc7396), a specialized media type 
+       `application/merge-patch+json` that is a partial resource representation. If updates are too complex to be expressed
+       with a merge patch, you should
+    3. use PATCH with [JSON Patch](http://tools.ietf.org/html/rfc6902), a specialized media type 
+       `application/json-patch+json` that includes instructions on how to change the resource
```
mostlymagic: looks good, but I would say any one endpoint should implement only one of these, right?

fmueller: Thought this is clear. Then maybe "to go for one of the following (per endpoint)".

whiskeysierra: Can you define endpoint?

ePaul: Path item, as OpenApi calls it? (I.e. one URI, or a group or URIs identified by a path pattern.)

ePaul:
:+1: if the "only one of these" issue is resolved.

http/Http.md
fmueller: 👍

dehora:
This could be phrased as guidance on when to use PUT -

```
use PUT with a custom media type and only if that isn't sufficient anymore
```

we could say  - 

```
If you want to create, or fully replace the state of resource, use PUT and not PATCH.
```

dehora:
```md
2. use PATCH with [JSON Merge Patch](https://tools.ietf.org/html/rfc7396), a specialized media type   `application/merge-patch+json` that is a partial resource representation. If updates are too complex to be expressed with a merge patch, you should
3. use PATCH with [JSON Patch](http://tools.ietf.org/html/rfc6902), a specialized media type 
  `application/json-patch+json` that includes instructions on how to change the resource
```

This guidance makes sense for the method. The [errata](https://www.rfc-editor.org/errata_search.php?rfc=5789) for rfc5789 lean that way -

```
Section 2 says:

   If  the operation does not modify the resource identified by the Request-
   URI in a predictable way, POST should be considered instead of PATCH
   or PUT.

It should say:

   If the operation does not modify the resource identified by the Request-
  URI in a predictable way that's defined by the semantics of the PATCH 
  media type, POST should be considered instead of PATCH or PUT.

...

RFC5789 does not explicitly tie PATCHing semantics to the media type of 
the request. This was well understood in the discussions around the document, 
and can be read between the lines in it, but it doesn't come out and say it.
```

Fwiw, I agree with the errata's suggestion to use `POST`. Personally I've not been convinced (in general) by the value of `PATCH` given the effort clients and servers need to expend to implement PATCH-centric media types. I am wondering then if we should add a fourth bullet about what to do when you don't have a patch-oriented media type, cribbed from the errata -

```md
4.   If the PATCH request does not modify the resource in a way defined by the 
semantics of the media type, POST should be used instead.
```

But I'd be happy to discuss that in another pr and not hold this one up, so in the meantime, :shipit:

http/Http.md
whiskeysierra: > looks good, but I would say any one endpoint should implement only one of these, right?

What if an endpoint going through the list evolutionary? Due to our rules on backwards compatibility one could end up with a single endpoint implementing multiple of the listed patterns.

whiskeysierra:
> If the operation does not modify the resource [..] in a predictable way

What would be an example for such an operation?

ePaul:
@dehora 

> This could be rephrased as guidance on when to use PUT -
> 
> ```
> use PUT with a custom media type and only if that isn't sufficient anymore
> ```
> 
> we could say  - 
> 
> ```
> If you want to create, or fully replace the state of resource, use PUT and not PATCH.
> ```

I think the point of the current wording is that even for changes of just a few fields, it is often easier to use PUT (after retrieving the old object and changing the fields in there) instead of PATCH, and PUT should be preferred in most cases.

dehora:
> What would be an example for such an operation?

For me, it's whatever's defined as predictable by the media type - eg an ["add"](https://tools.ietf.org/html/rfc6902#section-4.1) operation as defined by JSON Patch in rfc6902, which has a definition of "operation". But, rfc5789 itself doesn't clearly define what "operation" means - you can read it as refering to HTTP methods and to the act on the resource implied by the media type sent with PATCH in different places. Which is in part why the bullet 4 suggestion doesn't use the term "operation" :)

Bullets 2 and 3 move in the direction of rfc5789's intent (use specific media types designed for updates with PATCH), so :+1: for me. 

(What I'm expecting to see in real world API designs is using PATCH instead of POST, but without patch-oriented media types, ie submitting the same media type you'd send for a PUT or receive on a GET).

dehora:
"""
I think the point of the current wording is that even for changes of just a few fields, it is often easier to use PUT (after retrieving the old object and changing the fields in there) instead of PATCH, and PUT should be preferred in most cases.
"""

Got it. The 3 points as they are give better guidance than what we have now, so 👍

whiskeysierra:
I incorporated the suggested changes, please have a look.

mostlymagic:
LGTM

ePaul:
:+1:

PR #33, opened 2016-04-14
Remove duplicated ISO guidance.

dehora:
Looks like the ISO guidance is repeated. This removes `Use Standard Data Formats` and leaves `Use Standards for Country, Language and Currency Codes` in place.


ePaul:
Parent: #41. (This undoes one of the changes in the commit to be reviewed here.)

Issue #24, opened 2016-04-13
Add comparison of different link standards


whiskeysierra:
Right now the guidelines state:

> Should: Use Standards for Hypermedia Link Types
>
> To represent hypermedia links in payload results, we should consider using a standard format like HAL, JSON API, JSON-LD with Hydra, or Siren. (Hint: Read Kevin Sookocheff’s post, for instance, to learn more about the hypermedia types.)

We probably should not allow all link standards in the long term, since we are going for a consistent API look and feel. To move that discussion forward, maybe we could start adding a comparison of different standards.

fmueller:
Good point. @zmeda already did a comparison of different HATEOAS implementations: http://zmeda.github.io/mc-hateoas-evaluation/#/ We can use this as starting point of our comparison.

tfrauenstein:
Our plan was: 
step 1) restrict guideline to HAL and json:api for the time being. 
step 2) Align on Pros and Cons (based on separate doc, for the time being) <-- initial draft from Boris
step 3) Agree on some guideance update which one to be preferred (in which situation)

I propose to proceed like this -- step 1) immediately; step 2) @fmueller with Boris -- as input for next Guild JF

PR #34, opened 2016-04-14
Creating a references section

peletomi:
To be able to see in one page on, which documents we base our guidelines
I have collected the references from the guidelines. In this move I have
updated one reference to an obsolete RFC and made some markdown
formatting consistent with other sections in the document.

### DIFF
--- a/data-formats/DataFormats.md
+++ b/data-formats/DataFormats.md
@@ -2,7 +2,8 @@
 
 ## {{ book.must }} Use JSON as the Body Payload
 
-JSON-encode the body payload. The JSON payload must follow RFC-7158 and be compatible with RFC-4627 clients by having (if possible) a serialized object or array as the top-level object.
+JSON-encode the body payload. The JSON payload must follow [RFC-7159](https://tools.ietf.org/html/rfc7159) by having
+(if possible) a serialized object or array as the top-level object.
 
 ## {{ book.must }} Use Standard Date and Time Formats
 
--- a/http/Http.md
+++ b/http/Http.md
@@ -129,7 +129,7 @@ different HTTP methods on resources.
 | 501 | Not Implemented -  server cannot fulfill the request (usually implies future availability, e.g. new feature). | All |
 | 503 | Service Unavailable - server is (temporarily) not available (e.g. due to overload) -- client retry may be senseful. | All |
 
-All error codes you can find at [W3C](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html) and [Wikipedia](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) or via https://httpstatuses.com/<error_code>.
+All error codes you can find in [RFC7231](https://tools.ietf.org/html/rfc7231#section-6) and [Wikipedia](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) or via https://httpstatuses.com/<error_code>.
 
 ## {{ book.must }} Providing Error Documentation
 
--- a/references/References.md
+++ b/references/References.md
@@ -1,0 +1,30 @@
+# References
+
+This section collects links to documents to which we refer, and base our guidelines on.
+
+## RFCs
+
+* [RFC 3339: Date and Time on the Internet: Timestamps](https://tools.ietf.org/html/rfc3339)
+* [RFC 5988: Web Linking](https://tools.ietf.org/html/rfc5988)
+* [RFC 7230: Hypertext Transfer Protocol (HTTP/1.1): Message Syntax and Routing](https://tools.ietf.org/html/rfc7230)
+* [RFC 7231: Hypertext Transfer Protocol (HTTP/1.1): Semantics and Content](https://tools.ietf.org/html/rfc7231)
+* [RFC 7807: Problem Details for HTTP APIs](https://tools.ietf.org/html/rfc7807)
+
+## Dissertations
+
+* [Roy Thomas Fielding - Architectural Styles and the Design of Network-Based Software
+  Architectures](http://www.ics.uci.edu/~fielding/pubs/dissertation/top.htm)
+
+## Books
+
+* [REST in Practice: Hypermedia and Systems
+  Architecture](http://www.amazon.de/REST-Practice-Hypermedia-Systems-Architecture/dp/0596805829)
+* [Build APIs You Won't
+  Hate](https://leanpub.com/build-apis-you-wont-hate)
+* [InfoQ eBook - Web APIs: From Start to
+  Finish](http://www.infoq.com/minibooks/emag-web-api)
+
+## Blogs
+
+- [Lessons-learned blog: Thoughts on RESTful API
+  Design](http://restful-api-design.readthedocs.org/en/latest/)
--- a/references/References.md
+++ b/references/References.md
@@ -2,13 +2,23 @@
 
 This section collects links to documents to which we refer, and base our guidelines on.
 
+## OpenAPI Specification
+
+* [OpenAPI Specification](https://github.com/OAI/OpenAPI-Specification/)
+
 ## RFCs
 
 * [RFC 3339: Date and Time on the Internet: Timestamps](https://tools.ietf.org/html/rfc3339)
 * [RFC 5988: Web Linking](https://tools.ietf.org/html/rfc5988)
+* [RFC 7159: The JavaScript Object Notation (JSON) Data Interchange Format](https://tools.ietf.org/html/rfc7159)
 * [RFC 7230: Hypertext Transfer Protocol (HTTP/1.1): Message Syntax and Routing](https://tools.ietf.org/html/rfc7230)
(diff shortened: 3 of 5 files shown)


references/References.md
```diff
@@ -0,0 +1,30 @@
+# References
```
whiskeysierra: What about ISO standards? Currency, Country Code, Language, Locale?

mostlymagic: also, HATEOAS

references/References.md
```diff
@@ -0,0 +1,30 @@
...
+* [RFC 3339: Date and Time on the Internet: Timestamps](https://tools.ietf.org/html/rfc3339)
+* [RFC 5988: Web Linking](https://tools.ietf.org/html/rfc5988)
+* [RFC 7230: Hypertext Transfer Protocol (HTTP/1.1): Message Syntax and Routing](https://tools.ietf.org/html/rfc7230)
+* [RFC 7231: Hypertext Transfer Protocol (HTTP/1.1): Semantics and Content](https://tools.ietf.org/html/rfc7231)
+* [RFC 7807: Problem Details for HTTP APIs](https://tools.ietf.org/html/rfc7807)
+
+## Dissertations
+
+* [Roy Thomas Fielding - Architectural Styles and the Design of Network-Based Software
+  Architectures](http://www.ics.uci.edu/~fielding/pubs/dissertation/top.htm)
```
ePaul: Maybe add a note that "this is the definition of REST"?

references/References.md
```diff
@@ -0,0 +1,30 @@
+# References
+
+This section collects links to documents to which we refer, and base our guidelines on.
+
+## RFCs
```
ePaul: The "Data Formats" section has a reference to a JSON RFC (or two of them, actually).

references/References.md
ePaul: We should also refer to the [OpenAPI specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md), I guess.

references/References.md
peletomi: What do you mean?

mostlymagic: add a note to the effect that this dissertation actually defines what Rest is

references/References.md
peletomi: Thanks! I have added 7159, and removed 7159, and 4627 as they are obsolete.

references/References.md
ePaul: @mostlymagic I guess for HATEOAS references we should wait for #24 to be resolved (or add them with any pull requests resulting from that). Or do you have any specific references for this in mind?

peletomi: @mostlymagic which documents do you suggest?

mostlymagic: @peletomi I'm fine with @ePaul's suggestion

data-formats/DataFormats.md
```diff
@@ -2,7 +2,8 @@
 
 ## {{ book.must }} Use JSON as the Body Payload
 
-JSON-encode the body payload. The JSON payload must follow RFC-7158 and be compatible with RFC-4627 clients by having (if possible) a serialized object or array as the top-level object.
+JSON-encode the body payload. The JSON payload must follow [RFC-7159](https://tools.ietf.org/html/rfc7159) by having
+(if possible) a serialized object or array as the top-level object.
```
whiskeysierra: *Must* and *if possible* feels wrong. Is there valid reason to use something other than an array or object?

references/References.md
```diff
@@ -2,13 +2,23 @@
 
 This section collects links to documents to which we refer, and base our guidelines on.
 
+## OpenAPI Specification
+
+* [OpenAPI Specification](https://github.com/OAI/OpenAPI-Specification/)
+
 ## RFCs
```
whiskeysierra: Since we included ISO standards now, that headline doesn't really fit anymore, does it?

data-formats/DataFormats.md
mostlymagic: in theory, some endpoints could just return boolean, string or integer. But I also think we should discourage that

references/References.md
```diff
@@ -2,13 +2,23 @@
 
 This section collects links to documents to which we refer, and base our guidelines on.
 
+## OpenAPI Specification
```
whiskeysierra: Can we find a more generic term, because right now I'd suspect that we never add an item to that particular list, since it's too specific. Maybe *API Description Languages*?

data-formats/DataFormats.md
whiskeysierra: > in theory, some endpoints could just return boolean, string or integer

I'd always use a single-property object for that, otherwise you loose any possibility for future changes. A single primitive value also doesn't really feel like a resource to me.

mostlymagic: I agree, but we should explicitly say so

ePaul: The previous text suggested a reason to have an object/array: only that was defined in the older RFC 4627, so some old readers might not support primitives at the top level.

I guess another reason is the extensibility, as mentioned by @whiskeysierra. A proposal mentioning both:

> JSON-encode the body payload. The JSON payload must follow [RFC-7159](https://tools.ietf.org/html/rfc7159), and must have an object or array of objects as the top-level object. (By this, it will also be compatible to the earlier [RFC 4627](https://tools.ietf.org/html/rfc4627), and you can extend your API in the future by adding more fields.)

I don't like the "have an object ... as the top level object", as here "object" has two different meanings. What other word could be used for the second one?

references/References.md
```diff
@@ -0,0 +1,40 @@
...
+* [RFC 3339: Date and Time on the Internet: Timestamps](https://tools.ietf.org/html/rfc3339)
+* [RFC 5988: Web Linking](https://tools.ietf.org/html/rfc5988)
+* [RFC 7159: The JavaScript Object Notation (JSON) Data Interchange Format](https://tools.ietf.org/html/rfc7159)
+* [RFC 7230: Hypertext Transfer Protocol (HTTP/1.1): Message Syntax and Routing](https://tools.ietf.org/html/rfc7230)
+* [RFC 7231: Hypertext Transfer Protocol (HTTP/1.1): Semantics and Content](https://tools.ietf.org/html/rfc7231)
+* [RFC 7807: Problem Details for HTTP APIs](https://tools.ietf.org/html/rfc7807)
+* [ISO 8601: Date and time format](https://en.wikipedia.org/wiki/ISO_8601)
+* [ISO 3166-1 alpha-2: Two letter country codes](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2)
+* [ISO 639-1: Two letter language codes](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)
+* [ISO 4217: Currency codes](https://en.wikipedia.org/wiki/ISO_4217)
```
ePaul: We should either move the ISO standards out to their own section, or change the title of this section.

references/References.md
ePaul: * [Roy Thomas Fielding - Architectural Styles and the Design of Network-Based Software
      Architectures](http://www.ics.uci.edu/~fielding/pubs/dissertation/top.htm).
      This is the text which defines what REST is.

data-formats/DataFormats.md
mostlymagic: "as return type" or "as response body"?

references/References.md
ePaul: (Done now.)

data-formats/DataFormats.md
ePaul: Okay, new try:

    JSON-encode the body payload. The JSON payload in the response body must follow
    [RFC-7159](https://tools.ietf.org/html/rfc7159), and must be an object or an
    array of objects, not a primitive value. (By this restriction, it will also be
    compatible to the earlier [RFC 4627](https://tools.ietf.org/html/rfc4627),
    and you will be able to extend your API in the future by adding more fields.)

whiskeysierra: > body payload

Those are the same thing to me.

> JSON-encode the body payload.

I think the whole paragraph would work as well without that sentence.

http/Http.md
```diff
@@ -129,7 +129,7 @@ different HTTP methods on resources.
 | 501 | Not Implemented -  server cannot fulfill the request (usually implies future availability, e.g. new feature). | All |
 | 503 | Service Unavailable - server is (temporarily) not available (e.g. due to overload) -- client retry may be senseful. | All |
 
-All error codes you can find at [W3C](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html) and [Wikipedia](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) or via https://httpstatuses.com/<error_code>.
+All error codes you can find in [RFC7231](https://tools.ietf.org/html/rfc7231#section-6) and [Wikipedia](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) or via https://httpstatuses.com/<error_code>.
```
whiskeysierra: > All error codes you can find

All error codes can be found?

whiskeysierra:
:+1:

whiskeysierra:
I'm going to merge this now. Any further change can be done in a separate pull-request.

PR #36, opened 2016-04-15
RFC: Nakadi Event Guidelines

dehora:
Candidate guidelines for using events with Nakadi.

### DIFF
--- a/events/events.md
+++ b/events/events.md
@@ -1,0 +1,160 @@
+# Events
+
+Zalando’s architecture centers around decoupled microservices and in that context we favour asynchronous event driven approaches. The [Nakadi](https://github.com/zalando/nakadi) framework provides an event publish and subscribe system, abstracting exchanges through a RESTful API. 
+
+Nakadi is a key element for data integration in the architecture, and the standard system for inter-service event messaging. The guidelines in this section are for use with the Nakadi framework and focus on how to design and publish events intended to be shared for others to consume. Critically, once events pass service boundaries they are considered part of the service API and are subject to the API guidelines. 
+
+## {{ book.must }} Treat Events as part of the service interface
+
+Events are part of a service’s interface to the outside world equivalent in standing to a service’s REST API. Services publishing data for integration must treat their events as a first class design concern, just as they would an API. For example this means approaching events with the "API first" principle in mind [as described in the Introduction](../Introduction.md) and making them available for review.
+
+## {{ book.must }} Events should define useful business resources
+ 
+ Events are intended to be used by other services including business process/data analytics and monitoring. They should be based around the resources and business processes you have defined for your service domain and adhere to its natural lifecycle (see also "Should: Define useful resources" in the [General Guidelines](../general-guidelines/GeneralGuidelines.md)).
+ 
+ As there is a cost in creating an explosion of event types and topics, prefer to define event types that are abstract/generic enough to be valuable for multiple use cases, and avoid publishing event types without a clear need.
+
+## {{ book.must }} Event Types must conform to Nakadi's API
+
+Nakadi defines a structure called an _EventType_, which describes details for a particular kind of event. The EventType declares standard information, such as a name, an owning application (and by implication, an owning team), a well known event category (business process or data change), and a schema defining the event payload. It also allows the declaration of validation and enrichment strategies for events, along with supplemental information such as how events are partitioned in the stream. 
+
+An EventType is registered with Nakadi via its _Schema Registry API_. Once the EventType is created, individual events that conform to the type and its payload schema can be published, and consumers can access them as stream of Events. [Nakadi's Open API definitions](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml) include the definitions for two main categories, business events (BusinessEvent), and data change events (DataChangeEvent), as well as a generic 'undefined' type. 
+
+The service specific data defined for an event is called the _payload_. Only this non-Nakadi defined part of the event is expected to be submitted with the EventType schema.  When defining an EventType's payload schema as part of your API and for review purposes, it's ok to declare the payload as an object definition using Open API, but please note that Nakadi currently expects the EventType's schema to be submitted as JSON Schema and not as Open API JSON or YAML. Further details on how to register EventTypes are available in the Nakadi project's documentation.
+
+## {{ book.must }} Use the Business Events to signal steps and arrival points in business processes
+
+Nakadi defines a specific event type for business processes, called [BusinessEvent](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml#/definitions/BusinessEvent). When publishing events that represent steps in a business process, event types must be based on the BusinessEvent type.
+
+All your events of a single business process will conform to the following rules:
+
+- Business events must contain a specific identifier field (a business process id or "bp-id") similar to flow-id to allow for efficient aggregation of all events in a business process execution.
+
+- Business events must contain a means to correctly order events in a business process execution. In distributed settings where monotically increasing values (such as a high precision timestamp that is assured to move forwards) cannot be obtained, Nakadi provides a `parent_eids` data structure that allows causal relationships to be declared between events.
+
+- Business events should only contain information that is new to the business process execution at the specific step/arrival point.
+
+- Each business process sequence should be started by a business event containing all relevant context information.
+
+- Business events must be published reliably by the service.
+
+At the moment we cannot state whether it's best practice to publish all the events for a business process using a single event type and represent the specific steps with a state field, or whether to use multiple event types to represent each step. For now we suggest assessing each option and sticking to one for a given business process.
+
+## {{ book.must }} Use the Data Change Event structure to signal mutations
+
+Nakadi defines an event for signalling data changes, called a [DataChangeEvent](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml#/definitions/DataChangeEvent). When publishing events that represents created, updated, or deleted data, change event types must be based on the DataChangeEvent category.
+
+- Change events must identify the changed entity to allow aggregation of all related events for the entity.
+
+- Change events should contain a means of ordering events for a given entity (such as created or updated timestamps that are assured to move forwards with respect to the entity, or version identifiers provided by some databases). Note that basing events on data structures that can be converged upon (such as [CRDTs](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type) or [logical clocks](https://en.wikipedia.org/wiki/Logical_clock)) in a distributed setting are outside the scope of this guidance.
+
+- Change events must be published reliably by the service.
+
+- When the `hash` partition strategy is used, change events should identify a partition key that allows all related events for the entity to be consistently to assigned to a partition.
+
+## {{ book.should }} Data Change Events should match API representations
+
+A data change event's representation of an entity should correspond to the REST API representation. 
+
+There's value in having the fewest number of published structures for a service. Consumers of the service will be working with fewer representations, and the service owners will have less API surface to maintain. In particular, you should only publish events that are interesting in the domain and abstract away from implementation or local details - there's no need to reflect every change that happens within your system.
+
+There are cases where it could make sense to define data change events that don't directly correspond to your API resource representations. Some examples are -
+ 
+ - Where the API resource representations are very different from the datastore representation, but the physical data are easier to reliably process for data integration.
+
+ - Publishing aggregated event data that rolls up a set of changes. 
+ 
+ - Using events to indicate changes in relationship between entities, for example indicating items have had a tag applied to them, or have been classified in some way. 
+
+ - Events that are the result of a computation, such as a matching algorithm, or the generation of enriched data, and which might not be stored as entity by the service. 
+
+
+## {{ book.must }} Event Types must indicate ownership
+
+Event definitions must have clear ownership - this can be indicated via the `owning_application` field of the EventType. 
+
+Typically there is one producer application, which owns the EventType and is responsible for its definition, akin to how RESTful API definitions are managed. However, the owner may also be a particular service from a set of multiple services that are producing the same kind of event. Please note indicating that a specific application is producing or consuming a specific event type is not yet defined explicitly, but a subject of Nakadi service discovery support functions.
+
+## {{ book.must }} Event Payloads must be defined in accordance with the overall Guidelines
+
+Events must be consistent with other API data and the API Guidelines in general. 
+
+Everything expressed in the [Introduction to these Guidelines](../Introduction.md) is applicable to event data interchange between services. This is because our events, just like our APIs, represent a commitment to express what our systems do and designing high-quality, useful events allows us to develop new and interesting products and services. 
+
+What distinguishes events from other kinds of data is the delivery style used, asynchronous publish-subscribe messaging. But there is no reason why they could not be made available using a REST API, for example via a search request or as a paginated feed, and it will be common to base events on the models created for the service’s REST API. 
+
+The following existing guidelines for API data are applicable to events -
+
+[General Guidelines](../general-guidelines/GeneralGuidelines.md) -
+
+- Must: API First: Define APIs using OpenAPI YAML or JSON
+- Must: Write in U.S. English
+
+[Naming](../naming/Naming.md) -
+
+- Must: JSON field names must be snake_case (never camelCase)
+
+[Data Formats](../data-formats/DataFormats.md) -
+
+- Must: Use JSON as the Body Payload
+- Must: Use Standard Date and Time Formats
+- Could: Use Standards for Country, Language and Currency Codes
+- Could: Use Application-Specific Content Types
+
+[Common Data Objects](../common-data-objects/CommonDataObjects.md) -
+
+- Should: Use a Common Money Object
+- Should: Use Common Address Fields
+
+[Hypermedia](../hyper-media/Hypermedia.md) - 
+
+- Should: Use HATEOAS. The use of hypermedia techniques are equally applicable to events.
+- Could: Use URIs for Custom Link Relations
+- Should: Consider Using a Standard for linked/embedded resources. 
+
+There are a couple of technical considerations to bear in mind when defining event schema -
+
+ - Nakadi currently requires JSON Schema format, and does not yet support Open API YAML. It's ok to define your events for peer review in YAML to avoid maintaining duplicate representations.
+
+ - Open API's use of JSON Schema varies from the JSON Schema draft-4 specification. An example of this is Open API allowing extension structures such as `x-extensible-enum` that aren't directly expressible in JSON Schema. 
+
+These are best understood as point in time guidance around using Nakadi - as the project develops it may support consuming Open API (or other) structures directly, and the guidance here will be updated. 
+
+## {{ book.must }} Maintain backwards compatibility for Events
+
+Changes to events must be based around making additive and backward compatible changes. This follows the guideline, "Must: Don’t Break Backward Compatibility" from the [Compatibility guidelines](../compatibility/Compatibility.md). 
+
+In the context of events, compatibility issues are complicated by the fact that producers and consumers of events are highly asynchronous and can’t use content-negotiation techniques that are available to REST style clients and servers. This places a higher bar on producers to maintain compatibility as they will not be in a position to serve versioned media types on demand. 
+
+For event schema, these are considered backward compatible changes, as seen by consumers  -
+
+- Adding new optional fields to JSON objects.
+- Changing the order of fields (field order in objects is arbitrary).
+- Changing the order of values with same type in an array.
+- Removing optional fields.
+- Removing an individual value from an enumeration.
+
+These are considered backwards-incompatible changes, as seen by consumers  -
+
+- Removing required fields from JSON objects.
+- Changing the default value of a field.
+- Changing the type of a field, object, enum or array.
+- Changing the order of values with different type in an array (also known as a tuple).
+- Adding a new optional field to redefine the meaning of an existing field (also known as a co-occurrence constraint).
+- Adding a value to an enumeration (note that `x-extensible-enum` is not available in JSON Schema).
+
+## {{ book.must }} Event identifiers must be unique
+
+The `eid` (event identifier) value of an event must be unique.
+
+The `eid` property is part of the standard metadata for an event and gives the event an identifier. Producing clients must generate this value when sending an event and it must be guaranteed to be unique from the perspective of the owning application. In particular events within a given event type's stream must have unique identifiers. This allows consumers to process the `eid` to assert the event is unique and use it as an idempotency check. 
+
+Note that uniqueness checking of the `eid` is not enforced by Nakadi at the event type or global levels and it is the responsibility of the producer to ensure event identifiers do in fact distinctly identify events. A straightforward way to create a unique identifier for an event is to generate a UUID value.
+
+## {{ book.must }} Event Type names must follow conventions
+
+Event types can follow these naming conventions (each convention has its own should, must or could conformance level) -
+ 
+ - Event type names must be url-safe. This is because the event type names are used by Nakadi as part of the URL for the event type and its stream.
+
+ - Event type names should be lowercase words and numbers, using hypens, underscores or periods as separators. 
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -17,5 +17,6 @@
 * [Common Headers](headers/CommonHeaders.md)
 * [Proprietary Headers](headers/ProprietaryHeaders.md)
 * [API Discovery](api-discovery/ApiDiscovery.md)
+* [Events](events/events.md)
 * [References](references/References.md)
 * [Tooling](tooling/Tooling.md)


events/events.md
```diff
@@ -0,0 +1,107 @@
...
+
+## {{ book.must }} Event envelopes must conform to Nakadi's API
+
+[Nakadi](https://github.com/zalando/nakadi) defines general purpose event structures based on Open API. They acts as envelopes for carrying data specific to the type of event and are required to be able to work with Nakadi. 
+
+The definition is available as part of the Nakadi open source project, [here](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml). 
+
+## {{ book.must }} Event data must be based on Open API
+
+The specific data contained by the event must also be based on Open API. This schema is defined and owned by the team, and results in a new EventType that can be published to multiple consumers.
```
whiskeysierra: EventType -> event type

whiskeysierra: I'm not sure if event types should be owned by specific teams. Teams can change.

events/events.md
```diff
@@ -0,0 +1,107 @@
...
+
+[Hypermedia](http://zalando.github.io/restful-api-guidelines/hyper-media/Hypermedia.html) - 
+
+- Should: Use HATEOAS. The use of hypermedia techniques are equally applicable to events.
+- Could: Use URIs for Custom Link Relations
+- Should: Consider Using a Standard for linked/embedded resources. 
+
+## {{ book.must }} Treat Events as part of the service interface
+
+Events are part of a service’s interface to the outside world equivalent in standing to a service’s REST API, and are owned by the service that produces them. Services publishing data for integration treat their must as a first class design concern, just as they would an API - for example this means newly defined events should be made available for review.
```
whiskeysierra: > Services publishing data for integration treat their **must** as a first class design concern, just as they would an API

I think it's missing a word there

whiskeysierra: > and are owned by the service that produces them

I wouldn't couple APIs and services like that. Clients shouldn't care about a service. A service is an implementation. APIs are our abstraction. There will be cases where a service is superseded by a new one, implementing the same API. Or two services creating the same event type. Clients shouldn't know nor care.

events/events.md
```diff
@@ -0,0 +1,107 @@
...
+
+Teams creating internal events for use within a service boundary may treat this guideline as a _Should_.
+
+## {{ book.must }} Use the Data Change Event structure to signal mutations
+Nakadi defines a specific event for signalling data changes, called a [DataChangeEvent]DataChangeEvent](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml#/definitions/DataChangeEvent). When publishing events that represents created, deleted or updated data, event types must be based on the DataChangeEvent type.
```
whiskeysierra: Nitpicking, but I'd assume to have 

> created, deleted or updated

in chronological order, so

> created updated or deleted

events/events.md
```diff
@@ -0,0 +1,107 @@
...
+
+For event schema, these are considered backward compatible changes, as seen by consumers  -
+
+- Adding optional fields to JSON objects.
+- Changing the order of fields (field order in objects is arbitrary).
+- Changing the order of values in an array.
+- Changing the length or format of object IDs or other opaque strings. 
+- Removing optional fields.
+- Removing an individual value from an enumeration.
+- Adding new link relations where the relations are not defined by an enum.
```
whiskeysierra: Link relations should be strings, either well defined ones (http://www.iana.org/assignments/link-relations/link-relations.xhtml) or custom URIs (there is a rule in the guidelines for this).

events/events.md
```diff
@@ -0,0 +1,107 @@
...
+For event schema, these are considered backward compatible changes, as seen by consumers  -
+
+- Adding optional fields to JSON objects.
+- Changing the order of fields (field order in objects is arbitrary).
+- Changing the order of values in an array.
+- Changing the length or format of object IDs or other opaque strings. 
+- Removing optional fields.
+- Removing an individual value from an enumeration.
+- Adding new link relations where the relations are not defined by an enum. 
+- Changing the length or format of object IDs or other opaque strings.
```
whiskeysierra: Duplicate

events/events.md
```diff
@@ -0,0 +1,107 @@
...
+Teams creating internal events for use within a service boundary may treat this guideline as a _Should_.
+
+In the context of events, compatibility issues are complicated by the fact that producers and consumers of events are highly asynchronous and can’t use content-negotiation techniques that are available to REST style clients and servers. This places a higher bar on producers to maintain compatibility as they will not be in a position to serve different or versioned media types on demand. 
+
+For event schema, these are considered backward compatible changes, as seen by consumers  -
+
+- Adding optional fields to JSON objects.
+- Changing the order of fields (field order in objects is arbitrary).
+- Changing the order of values in an array.
+- Changing the length or format of object IDs or other opaque strings.
```
whiskeysierra: Tricky. If a producer specifies a string format, clients should be able to rely on this. If it can change clients shouldn't use them. Why have them in the first place then?

events/events.md
```diff
@@ -0,0 +1,107 @@
...
+- Changing the length or format of object IDs or other opaque strings. 
+- Removing optional fields.
+- Removing an individual value from an enumeration.
+- Adding new link relations where the relations are not defined by an enum. 
+- Changing the length or format of object IDs or other opaque strings. 
+
+These are considered backwards-incompatible changes, as seen by consumers  -
+
+- Removing required fields from JSON objects.
+- Adding an individual value to an enumeration.
```
whiskeysierra: Unless specified as `x-extensible-enum` (there is a rule for it, somewhere).

events/events.md
tkrop: I think, it is true that an event type in general is owned by a single publisher and should preferably only emitted by one location/service. Such an event type is normally part of the service API and the event type name should refers to this location.

Anyhow it might be possible that multiple publisher agree on using the same event type from multiple locations. Under this circumstances event type should be owned by separate domain/entity and be owned by a different API. Under this circumstances the event description in the API of the using publishers might best refer to the original API for event description.

events/events.md
tkrop: I'm indifferent on how to define that. I think, it is true that not every service consumer is although an event consumer and the other way around. Thus, it is not necessary to strictly bind Service API versions to Event API versions. A service deployment just is compatible against one or multiple versions of both APIs. May be we should state that Event and Service APIs should be defined in different documents that are related in name?

events/events.md
dehora: The idea here is to avoid clients depending on the internal details of string identifiers, for example that the string happens to be a UUID structure, can be coerced to an number, or is a specific length.

whiskeysierra: > The idea here is to avoid clients depending on the internal details of string identifiers

I 100% agree, but if that's what we are going for, why not just adding a rule that forbids declaring any format for id fields altogether?

dehora: @whiskeysierra right. If we do that, I'd rather add it as general guidance elsewhere. I'm going to pull this item out of the list in the meantime, and maybe spin up another pr for identifiers.

events/events.md
dehora: @tkrop +1. I've updated the text to make the point about multiple publishers using one event definition.

dehora: > I'm not sure if event types should be owned by specific teams. Teams can change.

@whiskeysierra +1. I've used the term publisher here instead.

events/events.md
tkrop: +1 for forbidding formats for id fields.

events/events.md
```diff
@@ -0,0 +1,107 @@
...
+
+
+## {{ book.must }} Define Events based on domain resources
+
+Events intended for use by other services, included data analytics, should be based around the resources you have defined for your services domain and their natural lifecycle (see also "Should: Define useful resources" in the [General Guidelines](http://zalando.github.io/restful-api-guidelines/general-guidelines/GeneralGuidelines.html)).
+
+There is a downside at the platform level to creating an explosion of event types and topics that might not be generally useful. In particular, you should avoid publishing fine grained or highly specific events that you wouldn't naturally express in your API or that are internal details to your service. 
+
+Teams creating internal events for use within a service boundary may treat this guideline as a _Could_, but it may be worth spending time defining and clarifying events as they might be promoted to general use in the future. 
+
```
tkrop: I suggest to remove all references to internal events. If events are strictly internal, they are not part of the (open) API but only of the service internal specification. Thus, it is only responsibility to the team how to communicate this specification (autonomy). Making suggestions about how this standard should be used for internal events the document unnecessary complex.

events/events.md
tfrauenstein: Yes, we should be more accurate on "ownership" of events:
Event type (incl. event scheme) is assigned to an application which is owned by team. (Of course, team ownership may change, but application assignment not.) Assigned application defines a service boundary and event is part of its API. Typically, the application is a microservice that provides REST APIs on his business resources and notifies other interested parties about business resource changes as single event producer. However, in other design situations assigned application may also be an event consumer -- just an asynchronous event service input interface. Here we may have multiple different service applications that produce the event, but do not own it.

whiskeysierra: > Event type (incl. event scheme) is assigned to an application

An application is an implementation. Why is it useful to expose that to anybody?

> Of course, team ownership may change, but application assignment not.

What if the team decides to rewrite an application completely? Now a different application *own* the event.

> Typically, the application is a microservice that provides REST APIs on his business resources

I'm still opposed to the idea that we should couple the notion of APIs and Events with an application. APIs (and that includes Events from now on) are the only thing that clients care about. If a team decides to split up a bigger service and serve parts of its API with three different applications from now on, nobody should care as long as the API stays the same, right?

events/events.md
dehora: @tkrop ok, tried removing the concept in 2ecad350887172f5a6e7ebb3c855ca0d02cca5ea

events/events.md
```diff
@@ -0,0 +1,98 @@
+# Events
+
+The [Nakadi](https://github.com/zalando/nakadi) project supports event-driven applications by abstracting exchanges through a RESTful API. The guidelines in this section are for use with Nakadi and focus on how to generally design and publish events intended to be shared for others to consume.
+
+## {{ book.must }} Event envelopes must conform to Nakadi's API
+
+[Nakadi](https://github.com/zalando/nakadi) defines general purpose event structures based on Open API. They acts as envelopes for carrying data specific to the type of event and are required to be able to work with Nakadi.
```
ePaul: Grammar: `acts` → `act`

events/events.md
```diff
@@ -0,0 +1,98 @@
+# Events
+
+The [Nakadi](https://github.com/zalando/nakadi) project supports event-driven applications by abstracting exchanges through a RESTful API. The guidelines in this section are for use with Nakadi and focus on how to generally design and publish events intended to be shared for others to consume.
+
+## {{ book.must }} Event envelopes must conform to Nakadi's API
+
+[Nakadi](https://github.com/zalando/nakadi) defines general purpose event structures based on Open API. They acts as envelopes for carrying data specific to the type of event and are required to be able to work with Nakadi. 
+
+The definition is available as part of the Nakadi open source project, [here](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml).
```
ePaul: Do not use "here" as a link title.

    [The envelope definition](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml) is
    available as part of the Nakadi open source project.

events/events.md
```diff
@@ -0,0 +1,98 @@
...
+
+## {{ book.must }} Event envelopes must conform to Nakadi's API
+
+[Nakadi](https://github.com/zalando/nakadi) defines general purpose event structures based on Open API. They acts as envelopes for carrying data specific to the type of event and are required to be able to work with Nakadi. 
+
+The definition is available as part of the Nakadi open source project, [here](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml). 
+
+## {{ book.must }} Event data must be based on Open API
+
+The specific data contained by the event must also be based on Open API. This schema is defined and owned by the publisher, and results in a new event type that can be published to multiple consumers.
```
ePaul: I guess we should say explicitly how these data schemas are to be defined?
I.e. some extension like `x-events` in the OpenAPI file, or include them simply in the `definitions` section?

events/events.md
```diff
@@ -0,0 +1,98 @@
...
+
+Publishers defining event content must avoid copying the Nakadi structures in their own definitions to reduce the amount of boilerplate in their Open API definitions. You can use these URIs to reference Nakadi definitions:
+
+- [#/definitions/Event](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml#/definitions/Event)
+
+- [#/definitions/BusinessEvent](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml#/definitions/BusinessEvent)
+
+- [#/definitions/DataChangeEvent](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml#/definitions/DataChangeEvent)
+
+_todo: establish final permalinks_
```
ePaul: Hmm, maybe I'm misunderstanding this, but how would one of those objects be used in an API definition?
Are we supposed to extend from one of those in our definition (using `allOf`), or should we simply define just our "payload" schema?

events/events.md
```diff
@@ -0,0 +1,98 @@
...
+
+[Nakadi](https://github.com/zalando/nakadi) defines general purpose event structures based on Open API. They acts as envelopes for carrying data specific to the type of event and are required to be able to work with Nakadi. 
+
+The definition is available as part of the Nakadi open source project, [here](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml). 
+
+## {{ book.must }} Event data must be based on Open API
+
+The specific data contained by the event must also be based on Open API. This schema is defined and owned by the publisher, and results in a new event type that can be published to multiple consumers. 
+
+It might be possible that multiple publishers agree on using the same event type from multiple locations. Under this circumstances event type should clearly belong to a particular API and the event description in the API of the using publishers can refer to the original API definition for the event description.
```
tkrop: Oh? Does this mean event types are kind of inherited by reference? In this case, we should explicitly state that the reference must be to an API explicitly including the major and minor version number, to make sure that referenced API does never automatically change the contract.

events/events.md
```diff
@@ -0,0 +1,98 @@
...
+- [#/definitions/DataChangeEvent](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml#/definitions/DataChangeEvent)
+
+_todo: establish final permalinks_
+
+## {{ book.must }} Events must be defined in accordance with the overall Guidelines
+Events can and must be consistent with other data in the Platform and the API Guidelines in general. 
+
+Everything expressed in the [Introduction to these Guidelines](http://zalando.github.io/restful-api-guidelines/index.html) is applicable to event data interchange between services. This is because our events, just like our APIs, represent a commitment to express what our systems do and designing high-quality, useful events allows us to develop new and interesting products and services. 
+
+What distinguishes events in the Platform from other kinds of data the delivery style used asynchronous, publish-subscribe. But there is no reason why they could not be made available using a REST API, for example via a search request or as a paginated feed, and it may be common to define an event using the same definition created for the service’s REST API.
```
tkrop: data the delivery -> data *is* the delivery

events/events.md
tkrop: Good question. Couldn't we handle the Nakadi details just a like a kind of container details and refer only to the payload? In this case, a DataChangeEvent of Nakadi is just like an example for an explicitly referenced event type used from multiple locations.

tkrop: To continue the discussion: It seems as we need a concept to specify event container (envelop) formats and event payload definitions. Is it possible to restrict the number of container formats to one, or do we need to allow define multiple nested containers? Similar, does a container format allow multiple payloads or can it be restricted to a single payload? And finally, how to express container parameters that must be defined in a fixed way as event type?

If I think about all these implications, it might be better to just recommend the use of standard event containers and supporting a reference to them, but to encourage the definition of all event aspects. Otherwise we may suggest best practices that may end up in ambiguities and misconceptions.

events/events.md
```diff
@@ -0,0 +1,98 @@
...
+- Changing the type of a field, object, enum or array.
+- Adding a new optional field to redefine the meaning of an existing field (also known as a co-occurrence constraint)
+
+
+## {{ book.must }} Define Events based on domain resources
+
+Events intended for use by other services, included data analytics, should be based around the resources you have defined for your services domain and their natural lifecycle (see also "Should: Define useful resources" in the [General Guidelines](http://zalando.github.io/restful-api-guidelines/general-guidelines/GeneralGuidelines.html)).
+
+There is a downside at the platform level to creating an explosion of event types and topics that might not be generally useful. In particular, you should avoid publishing fine grained or highly specific events that you wouldn't naturally express in your API or that are internal details to your service. 
+
```
tkrop: The discussion up to this point and especially the last paragraph is strongly focused on data change events. Effectively, it strongly discourages the use of business (process) events. In my opinion this neglects the value of business process events.

Business process events open up a new world for dynamic business analytics and are as least as important as for business process intelligence (BPI, aka business activity monitoring) as data change events are for business intelligence (BI). While RESTful APIs and data change events provide a view on the state of domain resources for BI, service communication logs and business process events provide a view on processes and activities for BPI.

Therefore, I suggest to define the guidelines much more open to (or better to encourage) business process events (in addition to data change events), as they provide huge benefits for ex-post analyzing and real-time monitoring of the effective business processes and activities.

dehora: > Effectively, it strongly discourages the use of business (process) events. In my opinion this neglects the value of business process events.

@tkrop absence of guidance is not discouragement, but I see your point.

It would be great add guidance on business process events, but it's not clear (to me) what that is. For example - when an order is placed, could/should that fire two events, one for the data change event for a new order and one for the business process to begin fulfilling that order. Similarly we might to clarify delivery concerns (dropping business process event seems problematic, either for fulfilment or anomaly detection). All that said, if we have guidance to give, we should add it.

events/events.md
dehora: :+1:

events/events.md
dehora: 👍

events/events.md
dehora: 👍

events/events.md
dehora: I've seen them in added to the definitions section so far. I'd like some consensus/feedback from teams on what's working for them, unless it's clear to us what the right way to do this is.

events/events.md
dehora: @tkrop I'm unsure what that would look like, or even if Swagger/OpenAPI supports that. Is there a way to specify a version of a defintion with a ref other than embedding it in the URI? (`$ref` is a string afaik).

The point for the guidance is that there should be one canonical owner for an event type (especially because event type names are mapped to resources in Nakadi). If we need to be very literal on how references are passed around and structured we should step back and think it through as this seems like OpenAPI level advice on sharing media types and media types are not a strong point of the project.

dehora: And thinking about it some more, I'm unsure reusing definitions is advisable at this point. I might strike this guidance.

events/events.md
dehora: > it might be better to just recommend the use of standard event containers and supporting a reference to them, but to encourage the definition of all event aspects. Otherwise we may suggest best practices that may end up in ambiguities and misconceptions.

@tkrop  I think we're in rough agreement. The initial thinking is that people should focus on defining their event data, and avoid vendoring Nakadi definitions into their project as part of their definition.  If you have alternative text to get that across, go for it, otherwise I'll take a pass later.

events/events.md
tkrop: @dehora thanks for pointing out, that my feedback was not specific enough. So let me try to reformulate the last section:

## {{ book.must }} Events must define useful business resources

Events are intended to be use by other services including business process/data analytics and monitoring. They should be based around the resources and business processes you have defined for your service domain and adhere to its natural lifecycle (see also "Should: Define useful resources" in the [General Guidelines](http://zalando.github.io/restful-api-guidelines/general-guidelines/GeneralGuidelines.html)).

As there is a risk in creating an explosion of event types and topics, event types should be abstract/generic enough to be valuable for multiple use cases. In particular, you should avoid publishing very fine grained or highly specific events. In addition, definition and publishing should be avoided unless a consumer is foreseeable and the use case is really valuable.

events/events.md
tkrop: @dehora agreed, go for it.

events/events.md
tkrop: @dehora Yes, I see. Nakadi has no concept of version and allows updates to event types. Thus its URI cannot be used to refer to an event schema. In addition, I see now that you sentence although reflects that in Nakadi an event type bound to an application instead of binding an event to a publisher/application-end-point. Anyhow, I still think, that event reuse is a valid concept that adds some value for certain kind of business process events - especially if it is defined generic enough to be of use for multiple consumers.

events/events.md
```diff
@@ -0,0 +1,98 @@
+# Events
+
+The [Nakadi](https://github.com/zalando/nakadi) project supports event-driven applications by abstracting exchanges through a RESTful API. The guidelines in this section are for use with Nakadi and focus on how to generally design and publish events intended to be shared for others to consume.
```
tfrauenstein: editorial: story line should be...
- we favour asynchronous event driven architectures
- latest when events pass service boundaries they are part of the service API and are subject of this guidelines
- with Nakadi we support an event pub/sub and discovery system as core infrastructure REST service -- quite analog to HTTP / REST infrastructure
- Nakadi is mandatory element for Data Integration and standard system for inter team event messaging
- so the guidelines below focus on the use of events with Nakadi framework

events/events.md
```diff
@@ -0,0 +1,98 @@
...
+
+## {{ book.must }} Event envelopes must conform to Nakadi's API
+
+[Nakadi](https://github.com/zalando/nakadi) defines general purpose event structures based on Open API. They act as envelopes for carrying data specific to the type of event and are required to be able to work with Nakadi.
```
tfrauenstein: "envelope" is not a term of the Nakadi API. I think we should very concrete relate to the Nakadi API terminology -- an event access API is defined via Event Type definition. Event Type has a name, assigned application (and by this also an owning team), event type class and a schema defining the event payload, i.e. event data representation -- supplemented by some information to control partitioning and event validation. We will define here rules for event naming, event type class selection, schema definition and event design --> to be mentioned in the intro above.

This rule here, is related to the event schema definition. 
The next rule too: "Event Schema Definition must..." -- more specifically on OpenAPI object scheme.

events/events.md
```diff
@@ -0,0 +1,98 @@
...
+
+## {{ book.must }} Event envelopes must conform to Nakadi's API
+
+[Nakadi](https://github.com/zalando/nakadi) defines general purpose event structures based on Open API. They act as envelopes for carrying data specific to the type of event and are required to be able to work with Nakadi. 
+
+[The envelope definition](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml) is available as part of the Nakadi open source project.
+
+## {{ book.must }} Event data must be based on Open API
+
+The specific data contained by the event must also be based on Open API. This schema is defined and owned by the publisher, and results in a new event type that can be published to multiple consumers.
```
tfrauenstein: One aspect is, how schemes have to be defined -- see above. 
Other aspects are ownership and pub/sub usage of event types by applications -- should be separated. 
Wrt. to the latter aspect I did not really understood message of next sentences -- What is "locations" and "particular API" and "original API definition"?

I propose to state the following:
- Event Types define an event application API for involved producer and consumer applications
- Event Type definition must have clear ownership, indicated via "owning application" 
- Typically there is one event producer application -- and it owns the Event Type. However, the owner may also be one dedicated application from a set of multiple applications producing the same event or even a specific consumer application. 
- Whether a specific application produces or consumes a specific event type is not (yet) defined explicitly, but a subject of Nakadi service discovery support functions.

events/events.md
```diff
@@ -0,0 +1,98 @@
...
+- Should: Use a Common Money Object
+- Should: Use Common Address Fields
+
+[Hypermedia](http://zalando.github.io/restful-api-guidelines/hyper-media/Hypermedia.html) - 
+
+- Should: Use HATEOAS. The use of hypermedia techniques are equally applicable to events.
+- Could: Use URIs for Custom Link Relations
+- Should: Consider Using a Standard for linked/embedded resources. 
+
+## {{ book.must }} Treat Events as part of the service interface
```
tfrauenstein: editorial: shift to top as 1st rule -- very generic. 
Add reference to API first principle which now also includes event type definition.

events/events.md
```diff
@@ -0,0 +1,98 @@
...
+These are considered backwards-incompatible changes, as seen by consumers  -
+
+- Removing required fields from JSON objects.
+- Adding an individual value to an enumeration, unless specified as x-extensible-enum.
+- Changing the default value of a field.
+- Changing the type of a field, object, enum or array.
+- Adding a new optional field to redefine the meaning of an existing field (also known as a co-occurrence constraint)
+
+
+## {{ book.must }} Define Events based on domain resources
```
tfrauenstein: editorial: shift to top as 2nd rule -- quite a generic event design rule

dehora: 👍 moved up to 2nd

events/events.md
dehora: 👍 moved up to 1st

events/events.md
dehora: updated intro text in 912c52bd102ecbc7ce5033f21d076d347833a9d9 to cover points

events/events.md
dehora: removed envelope jargon in 912c52bd102ecbc7ce5033f21d076d347833a9d9. will cover the rest in a followup.

events/events.md
dehora: > Wrt. to the latter aspect I did not really understood message of next sentences -- What is "locations" and "particular API" and "original API definition"?

I replaced that text and added a new guideline, "Event Types must indicate ownership", that uses the points above. Thinking on it more, reusing event type definitions such that multiple services can publish to the same stream is something we should gather experience on (there are cases where it make it can make sense such as log collation or user interface events).

events/events.md
dehora: @tkrop added in 2d812b1b3d11505fa9cb1079bd840d227755f7a1

events/events.md
```diff
@@ -0,0 +1,120 @@
...
+
+The `eid` (event identifier) value of an event must be unique.
+
+The `eid` value gives each event an identifier. Producing clients must generate this value when sending an event and it must be guaranteed to be unique from the perspective of the owning application. In particular events within a given event type's stream must have unique identifiers. 
+
+Note that uniqueness checking of the `eid` is not enforced by Nakadi at the event type or global levels and it is entirely the responsbility of the producer to ensure event identifiers do in fact distinctly identify events. This allows consumers to process the `eid` to assert the event is unique and use it as an idempotency check. The simplest way to create a unique identifier for an event is to generate a UUID value.
+
+Nakadi gives each event sent to a stream an `offset` that can be seen by clients via a `Cursor`, but this does cover the cases where an event is sent twice to a stream. The Cursors offset mechanism is instead intended to allow consumer client software to manage its position within the event type's stream.
+
+## {{ book.must }} Event Type names must follow conventions
```
dehora: The naming convention guidelines here are worth reviewing. They're derived in part from comments in Nakadi's yaml file.

```
- Event type must be url-safe.
```

Seems to be sensible given the names are used in URLs as path components

```
- Event type names should be lowercase words and numbers, using hypens, underscores or periods as separators. 
```

is roughly consistent with our guidance elsewhere, but more relaxed. Nakadi examples use a mix of, uppercase/lowercase, and period/hypen separators, and Nakadi itself has minimal name enforcement (which makes sense for the open source project). I'm thinking lowercase for overall consistency and also to avoid any unanticipated weirdness with case sensitivity in URIs.

```
- For data change events, the name should end in `change-event`, eg `order-change-event`

- For business events, the action part of name should be in the past tense, eg `order-received`
```

The naming conventions for business events in the past tense and data change events having a common suffix  are more stylistic and come from Nakadi's own suggestions.

There seems to be an open question about whether to put the owning app name into the event type name. I can see how they'll get out of date as teams/project names change, and it's not something we do for APIs, but also see why the namespacing could help.


@ePaul @fmueller @tfrauenstein @tkrop any thoughts on this, and/or whether we need to give any guidance at all here right now?

tkrop: @dehora I agree on url-safety and restricting names to patterns similar defined for path segments as they are used in this context and event in general similar to REST resources. 

As Nakadi re-uses event type names as primary key as well as resource/queue name while enforcing schema conformance, it makes sense to propose additional rules to guide a sensible name selection. For me the most important aspect would be to organize names according to functional and organizational domains. E.g. an order exists in many functional and organizational domains, and thus an order-received event should be named very specific representing its functional and organizational domain.

I would suggest something like <organizational-domain> "." <functional-domain>|<service-name>|<application-name> "." <sub-functional-domain>|<resource-name> "." <event-type>

Anyhow, naming conventions according to event types is - in my opinion - the least important aspect of a naming convention.

whiskeysierra: > E.g. an order exists in many functional and organizational domains, and thus an order-received event should be named very specific representing its functional and organizational domain.

This is the reason why *order* itself is discouraged as a business entity. We have *Sales Order*, *Purchase Order*, *Return Order*, *Shipment Order*, ...

dehora: > This is the reason why order itself is discouraged as a business entity.

@whiskeysierra noted, I'll change the examples

events/events.md
```diff
@@ -0,0 +1,120 @@
...
+
+The `eid` (event identifier) value of an event must be unique.
+
+The `eid` value gives each event an identifier. Producing clients must generate this value when sending an event and it must be guaranteed to be unique from the perspective of the owning application. In particular events within a given event type's stream must have unique identifiers. 
+
+Note that uniqueness checking of the `eid` is not enforced by Nakadi at the event type or global levels and it is entirely the responsbility of the producer to ensure event identifiers do in fact distinctly identify events. This allows consumers to process the `eid` to assert the event is unique and use it as an idempotency check. The simplest way to create a unique identifier for an event is to generate a UUID value.
+
+Nakadi gives each event sent to a stream an `offset` that can be seen by clients via a `Cursor`, but this does help in the case where an event is submitted multiple times to a stream, as each submitted item will be given its own offset value. The Cursors offset mechanism is instead intended to allow consumer client software to manage its position within the event type's stream.
```
ePaul: this does help → this does not help

events/events.md
```diff
@@ -0,0 +1,120 @@
...
+
+## {{ book.must }} Maintain backwards compatibility for Events
+
+Changes to events must be based around making additive and backward compatible changes. This follows the guideline, "Must: Don’t Break Backward Compatibility" from the [Compatibility guidelines](http://zalando.github.io/restful-api-guidelines/compatibility/Compatibility.html). 
+
+In the context of events, compatibility issues are complicated by the fact that producers and consumers of events are highly asynchronous and can’t use content-negotiation techniques that are available to REST style clients and servers. This places a higher bar on producers to maintain compatibility as they will not be in a position to serve different or versioned media types on demand. 
+
+For event schema, these are considered backward compatible changes, as seen by consumers  -
+
+- Adding optional fields to JSON objects.
```
ePaul: I guess adding required fields, or making an optional field required is also compatible (doesn't break consumers).

events/events.md
tkrop: Thinking about line 106, I actually would vote to delete it. Explaining the cursor and offset should be done be Nakadi. This aspect is not really related to the event guideline.

events/events.md
tkrop: @ePaul Updating a producer together with the event definition change is much more complicated than you anticipate, if event schema validation is active (which as I think should be best practice [do we need to have a section on that?]).

With event schema validation you either need (1) to stop the world for redeployment, or(2) to stop schema validation while redeploying is in progress, or (3) a migration strategy as adding first an optional field, second updating producers to always providing it, and third making the field required. It will even be more complicated, if multiple sources are allowed to produce the same event type.

I'm not sure whether we should suggest adding/updating to required fields without mentioning the potential trouble and necessary processes in place.

ePaul: I didn't think about schema validation (which happens on Nakadi, right?), you are right.

For new optional fields we need to first update the event type definition and then start producing them.

For making an optional field required, we first start to always produce it, and then update the event type definition to make them required. Same applies to removing optional fields (first stop producing, then remove from definition) and removing values from an enumeration (first stop producing the value, then remove it from the definition).

For introducing a new required fields, we need both steps.

events/events.md
```diff
@@ -0,0 +1,120 @@
...
+- Changing the order of fields (field order in objects is arbitrary).
+- Changing the order of values in an array.
```
ePaul: JSON Schema allows to define different schemas for consecutive values of an array. In this case changing the order of them is a quite breaking change, also for validation.

In the other case (all have the same type) it is no change at all, right?

tkrop: @ePaul I aggree...  unless we generally like to discourage arrays with different types, we should change this to:

- Changing the order of values with same type in an array.

And add a similar sentence for breaking changes:

- Changing the order of values with different type in an array.

events/events.md
```diff
@@ -0,0 +1,120 @@
...
+
+## {{ book.must }} Events must define useful business resources
+
+Events are intended to be use by other services including business process/data analytics and monitoring. They should be based around the resources and business processes you have defined for your service domain and adhere to its natural lifecycle (see also "Should: Define useful resources" in the [General Guidelines](http://zalando.github.io/restful-api-guidelines/general-guidelines/GeneralGuidelines.html)).
+
+As there is a risk in creating an explosion of event types and topics, you should define event types that are abstract/generic enough to be valuable for multiple use cases. In addition, you should avoid or delay definition and publishing of event types unless a consumer is foreseeable and the use case is really valuable.
+
+## {{ book.must }} Use the Data Change Event structure to signal mutations
```
tkrop: @dehora hi bill, I like to suggest the below section for business events. In the end I have attached slightly corrected version for data change events.

## {{ book.must }} Use the Business Event structure to signal steps and arrival points in business process

Nakadi defines a specific event type for signaling steps and arrival points in business processes, called [BusinessEvent] (https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml#/definitions/BusinessEvent). When publishing events that represent steps in a business process, event types must be based on the BusinessEvent type.

All your events of a single business process should be aligned according to the following rules to be minimal and most valuable:

- Business events must contain a specific identifier field (business-process-id/bp-id) similar to flow-id to allow aggregation of all events in business process execution.
- Business events must contain a source specific high precision time stamp (created) that reflects the moment of creation to allow for correct ordering of events in a business service execution.
- Business events should only contain only information that is new to the business process execution at the specific step/arrival point.
- All business events should be designed with respect to the underlying business process considering the possible sequences of events and sub-processes to minimize the exposed information at each step.
- Each business process sequence should be started by a business event containing all relevant context information.

At the current point in time we cannot state, whether it should be best practice to publish all business events of a business process using a single event type representing the specific steps in a state field, or using multiple event types each representing exactly one step. For now we only suggest to stick to one of the before mentioned principles.

Note: business events are explicitly assumed to be published reliable.

## {{ book.must }} Use the Data Change Event structure to signal mutations

Nakadi defines a specific event for signaling data changes, called a [DataChangeEvent](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml#/definitions/DataChangeEvent). When publishing events that represent creating, updating, or deleting of data, event types must be based on the DataChangeEvent type.

@dehora can you provide a similar guideline for design of data change events?

Note: data change events are explicitly assumed to be published reliable.

tkrop: @dehora by the way, thanks bill. You are doing a great job writing this chapter!!

events/events.md
dehora: > JSON Schema allows to define different schemas for consecutive values of an array

@Paul tuple type validation? Right. We should really consider disallowing that feature, it's fragile. For now I'll rephrase this.

dehora: > I'll rephrase this.

done, based it on @tkrop's text

events/events.md
dehora: rm'd it

events/events.md
dehora: @tkrop 

I've added the business section, working on the corresponding change section now. Take a look at https://github.com/zalando/restful-api-guidelines/pull/36/commits/98803bd2fe6771e1a3c1135177d2abdbcdbc2a74 which qualifies the timestamp guidance to allow for distributed business process execution, where services might not have a global clock to coordinate on, and mentions a nakadi feature that might be useful.

events/events.md
dehora: @ePaul @tkrop I'm thinking there's no guidance here right now (we might want to provide rules of play for schema evolution elsewhere when Nakadi gets used in anger)

events/events.md
dehora: change section is in, rebased into 880a157abb0e926b96a8358cb1c5e20e47cf86f4

events/events.md
```diff
@@ -8,11 +8,25 @@ Nakadi is a key element for data integration in the architecture, and the standa
 
 Events are part of a service’s interface to the outside world equivalent in standing to a service’s REST API. Services publishing data for integration must treat their events as a first class design concern, just as they would an API. For example this means newly defined events should be made available for review, and we approach events with the "API first" principle in mind [as described in the Introduction](http://zalando.github.io/restful-api-guidelines/index.html).
 
-## {{ book.must }} Events must define useful business resources
+## {{ book.must }} Use the Business Events to signal steps and arrival points in business processes
```
tkrop: @dehora oh, I see. Sorry for expressing my idea not clear enough. I was suggesting to add an additional section. This would create three sections:

1. Events must define useful business resources => as is as general introduction
2. Use the Business Events to signal steps and arrival points in business processes => according to my suggestion
3. Use the Data Change Event structure to signal mutations => the version with your extensions in your last [commit](https://github.com/zalando/restful-api-guidelines/commit/880a157abb0e926b96a8358cb1c5e20e47cf86f4)

events/events.md
```diff
@@ -8,11 +8,25 @@ Nakadi is a key element for data integration in the architecture, and the standa
...
+
+- Business events must contain a source specific high precision time stamp (created) that reflects the moment of creation to allow for correct ordering of events in a business service execution.
+
+- Business events should only contain only information that is new to the business process execution at the specific step/arrival point.
+
+- All business events should be designed with respect to the underlying business process considering the possible sequences of events and sub-processes to minimize the exposed information at each step.
+
+- Each business process sequence should be started by a business event containing all relevant context information.
+
+- Business events are published reliably by the service.
```
tkrop: @dehora In my contribution I emphasized the necessity of publishing events reliable, because using Nakadi as is would not be sufficiently. I you like it to state as rule, I would suggest to state it as follows:

- Business events *must be* published reliably by the service.

events/events.md
tkrop: @dehora "gets us in danger"? Agreed! Should not be in this section.

events/events.md
dehora: @tkrop adjusted in 5f4d491e73e35c7d9368e22f9075d0b3c7ce9787

dehora:
@tkrop @ePaul I think most of the feedback's been covered, thanks so much for the detailed review so far - I'll rebase shortly.

events/events.md
dehora: @tkrop missed this, updating.

dehora: @tkrop added third section back in 42d7fb38f8ecf400e844fd7c18b2ab676e1a00b4

dehora:
> we should write "to allow for efficient aggregation", 

@tkrop done in 2c70875

events/events.md
```diff
@@ -0,0 +1,149 @@
...
+
+## {{ book.must }} Treat Events as part of the service interface
+
+Events are part of a service’s interface to the outside world equivalent in standing to a service’s REST API. Services publishing data for integration must treat their events as a first class design concern, just as they would an API. For example this means newly defined events should be made available for review, and we approach events with the "API first" principle in mind [as described in the Introduction](http://zalando.github.io/restful-api-guidelines/index.html).
+
+## {{ book.must }} Events should define useful business resources
+ 
+ Events are intended to be use by other services including business process/data analytics and monitoring. They should be based around the resources and business processes you have defined for your service domain and adhere to its natural lifecycle (see also "Should: Define useful resources" in the [General Guidelines](http://zalando.github.io/restful-api-guidelines/general-guidelines/GeneralGuidelines.html)).
```
fmueller: Please use relative link.

dehora: :+1: fixed in 18c7a1f

fmueller:
👍

dehora:
@ePaul @fmueller  @tfrauenstein @tkrop @whiskeysierra rebased with feedback to date incorporated.

events/events.md
```diff
@@ -0,0 +1,152 @@
...
+
+- Change events must identify the changed entity to allow aggregation of all related events for the entity.
+
+- Change events should contain a means of ordering events for a given entity (such as an created or updated timestamps that are assured to move forwards with respect to the entity, or version identifiers provided by some databases). Note that basing events on advanced data structures that can be converged upon (such as [CRDTs](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type) or [logical clocks](https://en.wikipedia.org/wiki/Logical_clock)) in a distributed setting are outside the scope of this guidance.
+
+- Change events must be published reliably by the service.
+
+- Change events should identify a partition key fields to allow all related events for the entity to be consistently to assigned to a partition when the  `hash` partition strategy is used.
```
tkrop: fields -> field or identify a partition key -> identify partition key
consistently to assigned -> consistently assigned

events/events.md
```diff
@@ -0,0 +1,152 @@
...
+
+Nakadi defines a specific event type for signaling steps and arrival points in business processes, called [BusinessEvent](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml#/definitions/BusinessEvent). When publishing events that represent steps in a business process, event types must be based on the BusinessEvent type.
+
+All your events of a single business process should be aligned according to the following rules to be minimal and most valuable:
+
+- Business events must contain a specific identifier field (business-process-id/bp-id) similar to flow-id to allow for efficient aggregation of all events in a business process execution.
+
+- Business events must contain a means to correctly order events in a business service execution (such as a high precision timestamp that is assured to move forwards). In distributed settings where monotically increasing values cannot be obtained, Nakadi provides a `parent_eids` data structure that allows causal relationship to be declared between events.
+
+- Business events should only contain only information that is new to the business process execution at the specific step/arrival point.
```
tkrop: only container only information -> only contain information

events/events.md
```diff
@@ -0,0 +1,152 @@
...
+- Change events must identify the changed entity to allow aggregation of all related events for the entity.
+
+- Change events should contain a means of ordering events for a given entity (such as an created or updated timestamps that are assured to move forwards with respect to the entity, or version identifiers provided by some databases). Note that basing events on advanced data structures that can be converged upon (such as [CRDTs](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type) or [logical clocks](https://en.wikipedia.org/wiki/Logical_clock)) in a distributed setting are outside the scope of this guidance.
+
+- Change events must be published reliably by the service.
+
+- Change events should identify a partition key fields to allow all related events for the entity to be consistently to assigned to a partition when the  `hash` partition strategy is used.
+
+
+## {{ book.must }} Event Types must conform to Nakadi's API
```
tkrop: I'm not sure, but I have the feeling that this section should move up two sections. It introduces the special event types for BusinessEvent and DataChangeEvent. Other opinions?

events/events.md
```diff
@@ -0,0 +1,152 @@
...
+
+An EventType is registered with Nakadi via its _Schema Registry API_. Once the EventType is created, individual events that conform to the type and its payload schema can be published, and consumers can access them as stream of Events. [Nakadi's Open API definitions](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml) include the definitions for two main categories, business events (BusinessEvent), and data change events (DataChangeEvent). Both of these categories are extended to define event types.  
+
+Only the variable non-Nakadi API part of an event is expected to be submitted with the EventType schema, and this is different for the DataChangeEvent and BusinessEvent types -
```
tkrop: I have the feeling that this detailed explanation of differences is not necessary here. Referencing the event types in my opinion implies that there are differences, that one has to be compliant to.

events/events.md
dehora: 👍 done in 6b535fd096479b12a16607aeeca3e9de35e45d01

events/events.md
dehora: done in 241caabfebc817bd8e9ae976eac667049f90b56d

events/events.md
dehora: done in 241caabfebc817bd8e9ae976eac667049f90b56d

events/events.md
dehora: @tkrop I think we might need to cover this, so here's some rationale for the section. An instance of an event is a mashup of schema from two different owners - the custom extensions defined by an owning application and the standardized parts defined by the Nakadi API. Plus what gets sent to Nakadi as the event type schema and what get sent as an actual event don't always match (the event type schema alone can't be used to validate an event). It's fairly different to what we do elsewhere (even allowing for Problem JSON). It means there are different possible approaches to schema definition and supporting API First needs like peer review. There's at least four ways we could describe a schema -

 1. Copy the Nakadi definitions into local definitions. Problem: the local and Nakadi's definitions drift over time.
 
 1. Reference the Nakadi definitions using refs. Problem: Nakadi needs to support stable and very probably version locked URIs because refs are strings and can't express content negotiation the way a HTML `a` tag with a `media` attribute could [1].
 
 1. Reference the Nakadi definitions using refs plus discriminators. Much the same as the above, but plus discriminators which begs an Open API style discussion.

 1. Define only the custom parts of the event type as objects with the awareness that defining a business event with `metadata` is something we don't allow. Problem: this restricts what automated tools can do (eg something like twintip can't understand a given object is an event)

The chances of all teams converging on only one these I suspect is very small. That alone would break the common hand principle.

The guidance as it is now targets option 4, as it seems to involve the fewest assumptions with good room to evolve our approaches and nothing outside the guidelines needs to change for things to "work".   

----

[1] This is an Open API design flaw around refs, not anything to do with Nakadi, but we're stuck with it. Incidentally [HAL's](http://stateless.co/hal_specification.html) design allows for media type hinting, it just isn't in the spec

dehora: @tkrop all that said, I had a go at reducing the text down by unifying this guide with `Event Payloads must conform to Open API` in d24c3ee - let me know what you think.

tkrop: @dehora bill, I absolutely agree with your above rational. My above statement was only directed against the original sentence and the two paragraphs below. Sorry, that I didn't made it clear enough. By the way you rephrasing addressed my original concerns.

events/events.md
```diff
@@ -0,0 +1,141 @@
...
+
+An EventType is registered with Nakadi via its _Schema Registry API_. Once the EventType is created, individual events that conform to the type and its payload schema can be published, and consumers can access them as stream of Events. [Nakadi's Open API definitions](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml) include the definitions for two main categories, business events (BusinessEvent), and data change events (DataChangeEvent). Both of these categories are extended to define event types.  
+
+Only the variable non-Nakadi defined part of an event is expected to be submitted with the EventType schema. The service specific data defined for an event is called the _payload_.  When defining a event type's payload schema as part of your API and for review purposes, it's sufficient to declare the payload as an object definition using Open API. Note however that Nakadi currently expects the EventType to be submitted in JSON syntax. Further details on how to register Event Types are available in the Nakadi project's documentation.
```
tkrop: defining a event type's payload -> defining an event type's payload
how to register Event Types -> how to register event types

events/events.md
```diff
@@ -0,0 +1,141 @@
...
+
+## {{ book.must }} Event Type names must follow conventions
+
+Event types can follow these naming conventions (each convention has its own should, must or could conformance level) -
+ 
+ - Event type must be url-safe. This is because the event type names are used by Nakadi as part of the URL for the event type and its stream.
+
+ - Event type names should be lowercase words and numbers, using hypens, underscores or periods as separators. 
+
+ - For data change events, the name should end in `change-event`, eg `order-change-event`
```
tkrop: @dehora I agree on url-safety and restricting names to patterns similar defined for path segments as they are used in this context and event in general similar to REST resources. Giving some guidance on event names is surely helpful.

Anyhow, as Nakadi re-uses event type names as primary key as well as resource/queue name providing only a common global namespace, it makes sense to propose additional rules to guide a sensible name selection. For me the most important aspect would be to organize names according to functional and/or organizational domains:

I would suggest something like <organizational-domain> "." <functional-domain>|<service-name>|<application-name> "." <sub-functional-domain>|<resource-name> "." <event-type>

Naming conventions according to event types are - in my opinion - the least important aspect of a naming convention.

PS: In my last company such a more detailed naming pattern was even enforced by the review/registration process.

tkrop: @dehora sorry forgot to escape in my first suggestion. please read again.

events/events.md
dehora: @tkrop fixed in 3d2a099. Went with `Event Types -> EventTypes`, as there'd already been feedback around using Nakadi terminology in this part

events/events.md
dehora: > Naming conventions according to event types are - in my opinion - the least important aspect of a naming convention.

@tkrop No real argument against that from me. These last two points are taken from Nakadi's own suggestions in its API definition, the difference here being aligning the case/hypen structures with what we say elsewhere. 

> I would suggest something like 
> ```
> <organizational-domain> "." <functional-domain>|<service-name>|<application-name> "." <sub-functional-domain>|<resource-name> "." <event-type>
>```


Nakadi has a slightly different take on this - https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml#L1094-L1105

```
          The name for the EventType SHOULD follow the pattern, but is not enforced
          'stups_owning_application.event-type', for example 'gizig.price-change'.
          The components of the name are:
          * Organization: the organizational unit where the team is located; can be omitted.
          * Team name: name of the team responsible for owning application; can be omitted.
          * Owning application: SHOULD match the field owning_application; indicates
          * EventType name: name of this EventType; SHOULD end in ChangeEvent for DataChangeEvents; MUST be
          in the past tense for BusinessEvents.
          (TBD: how to deal with organizational changes? Should be reflected on the name of the EventType?
          Isn't it better to omit [organization:team] completely?)
```

I'm going to suggest we remove the last two bullet points of guidance for now and event type naming conventions in can be done in a separate pr/issue that focused on just that (we're getting close to 100 comments on this pr, it might be easier to work through things in a dedicated issue).

tkrop: @dehora agreed!

dehora: nice one @tkrop, removed in 56a909b. Will rebase shortly.

events/events.md
```diff
@@ -0,0 +1,138 @@
...
+
+## {{ book.must }} Treat Events as part of the service interface
+
+Events are part of a service’s interface to the outside world equivalent in standing to a service’s REST API. Services publishing data for integration must treat their events as a first class design concern, just as they would an API. For example this means newly defined events should be made available for review, and we approach events with the "API first" principle in mind [as described in the Introduction](../Introduction.md).
+
+## {{ book.must }} Events should define useful business resources
+ 
+ Events are intended to be use by other services including business process/data analytics and monitoring. They should be based around the resources and business processes you have defined for your service domain and adhere to its natural lifecycle (see also "Should: Define useful resources" in the [General Guidelines](../general-guidelines/GeneralGuidelines.md)).
```
bocytko: "to be use" => "to be used"?

events/events.md
```diff
@@ -0,0 +1,138 @@
...
+
+- Change events must identify the changed entity to allow aggregation of all related events for the entity.
+
+- Change events should contain a means of ordering events for a given entity (such as an created or updated timestamps that are assured to move forwards with respect to the entity, or version identifiers provided by some databases). Note that basing events on advanced data structures that can be converged upon (such as [CRDTs](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type) or [logical clocks](https://en.wikipedia.org/wiki/Logical_clock)) in a distributed setting are outside the scope of this guidance.
+
+- Change events must be published reliably by the service.
+
+- Change events should identify a partition key to allow all related events for the entity to be consistently to assigned to a partition when the  `hash` partition strategy is used.
```
bocytko: Duplicated space in 

```
the `hash`
```

events/events.md
```diff
@@ -0,0 +1,138 @@
...
+
+- Each business process sequence should be started by a business event containing all relevant context information.
+
+- Business events must be published reliably by the service.
+
+At the moment we cannot state whether it should be best practice to publish all business events of a business process using a single event type representing the specific steps in a state field, or using multiple event types each representing exactly one step. For now we only suggest to stick to one of the before mentioned principles.
+
+## {{ book.must }} Use the Data Change Event structure to signal mutations
+
+Nakadi defines a specific event for signalling data changes, called a [DataChangeEvent](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml#/definitions/DataChangeEvent). When publishing events that represents created, updated, or deleted data, change event types must be based on the DataChangeEvent type.
```
bocytko: I would include a recommendation here to keep the representation of the entities as close as possible to the exposed REST API for it.

dehora: :+1: 5a06f83

events/events.md
dehora: :+1: 5a06f83

events/events.md
dehora: :+1: 5a06f83

PR #37, opened 2016-04-18
Added hint to pagination and ux related issues

fmueller:
After an internal discussion about an article which states that ux researches shows load more and next/previous page navigation is favoured by users and results in better conversions than page number navigation, this change adds reference to this in pagination chapter.

### DIFF
--- a/pagination/Pagination.md
+++ b/pagination/Pagination.md
@@ -13,6 +13,10 @@ There are two page iteration techniques:
   unique key element identifies the first page entry (see also
   [Facebook’s guide](https://developers.facebook.com/docs/graph-api/using-graph-api/v2.4#paging))
 
+The technical conception of pagination should also consider user experience related issues. As mentioned
+in this [article](https://www.smashingmagazine.com/2016/03/pagination-infinite-scrolling-load-more-buttons/),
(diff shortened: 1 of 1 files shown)


pagination/Pagination.md
```diff
@@ -13,6 +13,10 @@ There are two page iteration techniques:
   unique key element identifies the first page entry (see also
   [Facebook’s guide](https://developers.facebook.com/docs/graph-api/using-graph-api/v2.4#paging))
 
+The technical conception of pagination should also consider user experience related issues. As mentioned
+in this [article](https://www.smashingmagazine.com/2016/03/pagination-infinite-scrolling-load-more-buttons/),
```
whiskeysierra: Since the pr for the references section is merged now, what is our rule for putting links there? Should this be there as well?

whiskeysierra:
:+1:

pagination/Pagination.md
fmueller: Good point, the question for me is, does reference section mean 'official reference documents' like RFC and api documentations or links in general? What's the consensus on this?

whiskeysierra: > 'official reference documents' like RFC and api

I think we can go with that for now. So no changes needed.

PR #38, opened 2016-04-19
Added temporary recommendation for hal and jsonapi as hateoas impls

fmueller:
As mentioned in #24 we temporarily recommend hal and json api. In next api guild meeting, a comparison of both is presented + experiences of merchant center teams with hateoas. So that we can come up with a conclusion on hateaos.

### DIFF
--- a/hyper-media/Hypermedia.md
+++ b/hyper-media/Hypermedia.md
@@ -19,6 +19,7 @@ Additional resources: [Wikipedia](http://en.wikipedia.org/wiki/HATEOAS), [The RE
 Related or even embedded (sub-) resources should be linked via “meta link attributes” within the response payload; use here the following [HAL](http://stateless.co/hal_specification.html) compliant syntax:
 
     {
+      ...
       “_links”: {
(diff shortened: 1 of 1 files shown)


hyper-media/Hypermedia.md
```diff
@@ -19,23 +19,43 @@ Additional resources: [Wikipedia](http://en.wikipedia.org/wiki/HATEOAS), [The RE
 Related or even embedded (sub-) resources should be linked via “meta link attributes” within the response payload; use here the following [HAL](http://stateless.co/hal_specification.html) compliant syntax:
 
     {
+      ...
```
whiskeysierra: I'd move this after the _links property. Just my personal preference though, since I tend to declare links up front and traditional properties later.

whiskeysierra:
:+1:

hyper-media/Hypermedia.md
fmueller: It's a matter of taste and because this is only a temporary solution until we've decided on HATEOAS in our guild, I'll leave it as it is.

Issue #42, opened 2016-04-26
Don't specify all possible HTTP status codes


ePaul:
The guidelines currently say (in section "HTTP", rule "[Must: Providing Error Documentation](http://zalando.github.io/restful-api-guidelines/http/Http.html#must-providing-error-documentation)"):

> The OpenAPI specification shall include definition of all error status codes and error descriptions
> that possibly will be returned; they are part of the interface definition and provide important
> information for service clients to handle exceptional situations and support troubleshooting.

I think we should instead encourage to not document common HTTP errors which have no special meaning for this API, but just the default HTTP meaning, and can happen always (like 500, 503, ...).

Clients should be prepared for unexpected response codes, and in case of doubt handle them like they would handle the corresponding x00 code. Adding new response codes (specially error responses) should be considered a compatible API evolution.

(This might also affect the compatibility section.)

ePaul:
(This text arrived in the guidelines with  a68803b, thus parent issue: #41.)

tfrauenstein:
I think we should not mix both aspects.
a) API spec: intention of the guideline was: "explicitly define the returned errors as part of the interface spec". IMO it should be valid also for common HTTP errors. 
However, a bad practise is: defining error codes in the API that are not yet returned, but look like good candidates to be returned in some future with more experience in implementation operation.
b) Compatible API changes: define adding newly returned errors in API as compatible extension. Hence, clients should be prepared for unexpected response error codes -- at least they should provide analog handling as for corresponding x00 code as default.

P.S: We discussed it in a more recent API Guild meeting and actually @whiskeysierra accepted an action item to provide PR for corresponding update of both guidelines.

ePaul:
@tfrauenstein I'm not sure how to distinguish "error codes that are not yet returned" from "error codes that are returned" in an API first setting (i.e. before any implementation is there).

What triggered this issue was the question whether we (as API designers) need to add a '500' (internal Server error) and '503' (temporary not available) response for all operations, because one can never be certain that these don't happen, or if we can simply omit them from the description, and expect the client to handle them anyways.

whiskeysierra:
- differentiation between technical and functional errors
- technical errors don't need to be specified (covered by frameworks, e.g. 401, 403, 406, 415. doesn't add any value to the API)
- functional problems need to be specified, that includes a 404 if a resource wasn't found (if and only if it's actually a special case)
- Should include meta information, e.g. Retry-After, problem body, etc. to help the client handle it gracefully
[...]

Issue #41, opened 2016-04-26
Review the "merge all pending changes from google doc" commit


ePaul:
In commit a68803b1, a lot of changes were merged into the guideline without proper discussion. Those originated as change proposals in our internal Google doc, and had different states of approval at this point.

I was already surprised by several of those changes since then. I guess we should go systematically over all those changes and discuss whether they are good, and then possibly roll back the bad ones, or replace the text with better text.

This issue is a meta issue, we should create separate issues for individual points.

ePaul:
Apart from #42 all the sub-issues are now closed, so I'm closing this one too.

PR #43, opened 2016-04-26
Restrict error definition guidance to response entities.

dehora:
This qualifies the guidance on complete description of error
documentation in definitions to be limited to response entities, and avoids
an interpretation that all HTTP response codes that a resource
might return are to be described in advance.

related: https://github.com/zalando/restful-api-guidelines/issues/42


ePaul:
Parent Issue: #41.

whiskeysierra:
👍

ePaul:
I'm not sure if this actually resolves #42, but this is at least better than the current version, so :+1: from me too.

PR #61, opened 2016-05-10
Bugfix/#52 fix review issues for common headers section

ePaul:
Fixes two things:

* use `ID`/`UID` instead of `Id`/`Uid`, as told by »Exceptions are common abbreviations like “ID.”« in the Naming section.
* use an UUID instead of the string `Zalando-Fashion-Store` as the example to the `X-Tenant-ID` header.


fmueller:
👍

PR #63, opened 2016-05-11
Updated rules for hypermedia and links

### LINKED ISSUE #24: Add comparison of different link standards

### LINKED ISSUE #54: Recommendations for links and IDs
In last API guild meetup we decided to update guidelines on links usage in payload and ids:

- all links should be present in _links field which can also be used multiple times in one payload
- favor simple IDs if in same namespace/system otherwise think about using URNs e.g. urn:warehouse:FG-AS-123-FD

whiskeysierra:
Fixes #24 
Fixes #54 

- deprecated Link headers
- favored HAL over other standards
- added hypermedia format comparison

### DIFF
--- a/hyper-media/Hypermedia.md
+++ b/hyper-media/Hypermedia.md
@@ -39,6 +39,12 @@ Interesting articles for comparisons of different hypermedia formats:
 * [Kevin Sookocheff’s On choosing a hypermedia type for your API](http://sookocheff.com/post/api/on-choosing-a-hypermedia-format/)
 * [Mike Stowe's API Best Practices: Hypermedia](http://blogs.mulesoft.com/dev/api-dev/api-best-practices-hypermedia-part-3/)
 
+## {{ book.must }} Do Not Use Link Headers with JSON entities
+
+We don't allow the use of the [`Link` Header defined by RFC 5988 ](http://tools.ietf.org/html/rfc5988#section-5) 
+in conjunction with JSON media types, and favor [HAL](#must-use-hal) instead. The primary reason is to have a consistent
+place for links as well as the better support for links in JSON payloads compared to the uncommon link header syntax.
+
 ## {{ book.could }} Use URIs for Custom Link Relations
 
 You should consider using a custom link relation if and only if standard [link relations](http://www.iana.org/assignments/link-relations/link-relations.xml)
--- a/hyper-media/Hypermedia.md
+++ b/hyper-media/Hypermedia.md
@@ -14,8 +14,7 @@ Links to other resources must be defined exclusively using [HAL](http://stateles
 preferably using standard [link relations](http://www.iana.org/assignments/link-relations/link-relations.xml).
 
 Clients and Servers are required to support `_links` with its `href` and `rel` attributes, not only at the root level
-but also in nested objects. To reduce the effort needed by clients to process hypertext data from servers it's not
-required to implement CURIEs, URI templates or embedded resources. Nor is it required to support the HAL media type
+but also in nested objects. To reduce the effort needed by clients to process hypertext data from servers it's not recommended to serve data with CURIEs, URI templates or embedded resources. Nor is it required to support the HAL media type
 `application/hal+json`.
 
 We opted for this subset of HAL after conducting a comparison of different hypermedia formats based on properties like:
--- a/hyper-media/Hypermedia.md
+++ b/hyper-media/Hypermedia.md
@@ -18,7 +18,7 @@ but also in nested objects. To reduce the effort needed by clients to process hy
 required to implement CURIEs, URI templates or embedded resources. Nor is it required to support the HAL media type
 `application/hal+json`.
 
-We opted for this subset of HAL after conduction a comparison of different hypermedia formats based on properties like:
+We opted for this subset of HAL after conducting a comparison of different hypermedia formats based on properties like:
 
 * Simplicity: resource link syntax and concepts are easy to understand and interpret for API clients.
 * Compatibility: introducing and adding links to resources is not breaking existing API clients.
@@ -41,7 +41,7 @@ Interesting articles for comparisons of different hypermedia formats:
 
 ## {{ book.must }} Do Not Use Link Headers with JSON entities
 
-We don't allow the use of the [`Link` Header defined by RFC 5988 ](http://tools.ietf.org/html/rfc5988#section-5) 
+We don't allow the use of the [`Link` Header defined by RFC 5988](http://tools.ietf.org/html/rfc5988#section-5) 
 in conjunction with JSON media types, and favor [HAL](#must-use-hal) instead. The primary reason is to have a consistent
 place for links as well as the better support for links in JSON payloads compared to the uncommon link header syntax.
 
--- a/hyper-media/Hypermedia.md
+++ b/hyper-media/Hypermedia.md
@@ -28,7 +28,8 @@ We opted for this subset of HAL after conducting a comparison of different hyper
 
 | Standard                                                       | Simplicity | Compatibility | Adoption | Primary Focus           | Docs |
 |----------------------------------------------------------------|------------|---------------|----------|-------------------------|------|
-| [HAL](http://stateless.co/hal_specification.html)              | ✓          | ✓             | ✓        | Links and relationships | ✓    |
+| HAL Subset                                                     | ✓          | ✓             | ✓        | Links and relationships | ✓    |
+| [HAL](http://stateless.co/hal_specification.html)              | ✗          | ✓             | ✓        | Links and relationships | ✓    |
 | [JSON API](http://jsonapi.org/)                                | ✗          | ✗             | ✓        | Response format         | ✓    |
 | [JSON-LD](http://json-ld.org/)                                 | ✗          | ✓             | ?        | Link data               | ?    |
 | [Siren](https://github.com/kevinswiber/siren)                  | ✗          | ✗             | ✗        | Entities and navigation | ✗    |
--- a/hyper-media/Hypermedia.md
+++ b/hyper-media/Hypermedia.md
@@ -4,13 +4,47 @@
 
 We prefer [REST Maturity Level 2](http://martinfowler.com/articles/richardsonMaturityModel.html#level2). Only if HATEOAS - as [Level 3](http://martinfowler.com/articles/richardsonMaturityModel.html#level3) - adds value to your API, consider using it. Because we are following API First, HATEOAS adds not much value for us in terms of self-descriptiveness. And we don't believe in generic HATEOAS clients which crawl and use APIs on their own. So, only use HATEOAS if your API really defines possible client actions (one part of hypermedia) and leverages other HATEOAS features.
 
-Decoupled from HATEOAS, adding links and relationships should be considered as useful addition to APIs as it adds context and helps navigation in APIs to decent degree.
+Decoupled from HATEOAS, adding [links and relationships](#must-use-hal) should be considered as useful addition to APIs as it adds context and helps navigation in APIs to decent degree.
 
 Additional resources: [Wikipedia](http://en.wikipedia.org/wiki/HATEOAS), [The RESTful CookBook](http://restcookbook.com/Basics/hateoas/)
 
+## {{ book.must }} Use a well-defined subset of HAL
+
+Links to other resources must be defined exclusively using [HAL](http://stateless.co/hal_specification.html) and
+preferably using standard [link relations](http://www.iana.org/assignments/link-relations/link-relations.xml).
+
+Clients and Servers are required to support `_links` with its `href` and `rel` attributes, not only at the root level
+but also in nested objects. To reduce the effort needed by clients to process hypertext data from servers it's not
+required to implement CURIEs, URI templates or embedded resources. Nor is it required to support the HAL media type
+`application/hal+json`.
+
+We opted for this subset of HAL after conduction a comparison of different hypermedia formats based on properties like:
+
+* Simplicity: resource link syntax and concepts are easy to understand and interpret for API clients.
+* Compatibility: introducing and adding links to resources is not breaking existing API clients.
+* Adoption: use in open-source libraries and tools as well as other companies
+* Docs: degree of good documentation
+
+<p></p>
+
+| Standard                                                       | Simplicity | Compatibility | Adoption | Primary Focus           | Docs |
+|----------------------------------------------------------------|------------|---------------|----------|-------------------------|------|
+| [HAL](http://stateless.co/hal_specification.html)              | ✓          | ✓             | ✓        | Links and relationships | ✓    |
+| [JSON API](http://jsonapi.org/)                                | ✗          | ✗             | ✓        | Response format         | ✓    |
+| [JSON-LD](http://json-ld.org/)                                 | ✗          | ✓             | ?        | Link data               | ?    |
+| [Siren](https://github.com/kevinswiber/siren)                  | ✗          | ✗             | ✗        | Entities and navigation | ✗    |
+| [Collection+JSON](http://amundsen.com/media-types/collection/) | ✗          | ✗             | ✗        | Collections and queries | ✗    |
+
+Interesting articles for comparisons of different hypermedia formats:
+* [Kevin Sookocheff’s On choosing a hypermedia type for your API](http://sookocheff.com/post/api/on-choosing-a-hypermedia-format/)
+* [Mike Stowe's API Best Practices: Hypermedia](http://blogs.mulesoft.com/dev/api-dev/api-best-practices-hypermedia-part-3/)
+
 ## {{ book.could }} Use URIs for Custom Link Relations
 
-Related or even embedded (sub-) resources should be linked via “meta link attributes” within the response payload; use here the following [HAL](http://stateless.co/hal_specification.html) compliant syntax:
+You should consider using a custom link relation if and only if standard [link relations](http://www.iana.org/assignments/link-relations/link-relations.xml)
+are not sufficient to express a relation.
+Related or even embedded (sub-) resources should be linked via “meta link attributes” within the response payload; use
+the following [HAL](http://stateless.co/hal_specification.html) compliant syntax:
 
     {
       ...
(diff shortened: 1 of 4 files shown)


hyper-media/Hypermedia.md
```diff
@@ -4,13 +4,47 @@
...
+## {{ book.must }} Use HAL
+
+Links to other resources must be defined exclusively using [HAL](http://stateless.co/hal_specification.html) and
+preferably using standard [link relations](http://www.iana.org/assignments/link-relations/link-relations.xml).
+
+We opted for HAL after conduction a comparison of different hypermedia formats based on properties like:
+
+* Simplicity: easy to understand and implement in a non-breaking way
+* Adoption: use in open-source libraries and tools as well as other companies
+* Docs: degree of good documentation
```
tkrop: I remember we had the argument of backward compatibility when we talked about comparing the below technologies. Can we add a column below (or express this in the list above)?

whiskeysierra: > implement in a non-breaking way

hyper-media/Hypermedia.md
```diff
@@ -4,13 +4,47 @@
...
+
+Interesting articles for comparisons of different hypermedia formats:
+* [Kevin Sookocheff’s On choosing a hypermedia type for your API](http://sookocheff.com/post/api/on-choosing-a-hypermedia-format/)
+* [Mike Stowe's API Best Practices: Hypermedia](http://blogs.mulesoft.com/dev/api-dev/api-best-practices-hypermedia-part-3/)
+
+## {{ book.must }} Do Not Use Link Headers
+
+We don't allow the usage of [link headers](http://tools.ietf.org/html/rfc5988#section-5) and favor [HAL](#must-use-hal) instead.
+The primary reason is to have a consistent place for links as well as the better support for links in JSON payloads compared
+to the uncommon link header syntax.
```
ePaul: Could this be made a SHOULD? There are cases where HAL-links don't work, like when you don't have JSON, but are delivering other content (like a PDF document or an image).

whiskeysierra: I interpret *must* as: *if you have links, then you must use HAL*

fmueller:
👍

hyper-media/Hypermedia.md
tkrop: @whiskeysierra I see, but this is hardly understandable for a reader and difficult to interpret together in one column. I suggest to use an own column for this aspect, and to give an better comprehensible explanation for the reader like:

Simplicity: resource link syntax and concepts are easy to understand and interpret for API clients.
Compatibility: introducing and adding links to resources is not breaking existing API clients.

fmueller: But this would break the table layout... ^^ (not a real reason, but we've fighted a lot to condense all the information in nice formatted table)

hyper-media/Hypermedia.md
ePaul: Yes, but if you want to send links with non-JSON data, maybe a link header is needed?

whiskeysierra: > Yes, but if you want to send links with non-JSON data, maybe a link header is needed?

Interesting aspect. I haven't thought about it. Which media types do you have in mind? For XML there should be better alternatives, e.g. atom links. I'd leave it like this until we see a use case for having links in media types like PDF etc.

hyper-media/Hypermedia.md
```diff
@@ -4,13 +4,47 @@
...
+| [JSON API](http://jsonapi.org/)                                | ✗          | ✓        | Response format         | ✓    | 2015-05-29  |
+| [JSON-LD](http://json-ld.org/)                                 | ✗          | ?        | Link data               | ?    | 2014-01-16  |
+| [Siren](https://github.com/kevinswiber/siren)                  | ✗          | ✗        | Entities and navigation | ✗    | 2015-10-21  |
+| [Collection+JSON](http://amundsen.com/media-types/collection/) | ✗          | ✗        | Collections and queries | ✗    | 2013-02-24  |
+
+Interesting articles for comparisons of different hypermedia formats:
+* [Kevin Sookocheff’s On choosing a hypermedia type for your API](http://sookocheff.com/post/api/on-choosing-a-hypermedia-format/)
+* [Mike Stowe's API Best Practices: Hypermedia](http://blogs.mulesoft.com/dev/api-dev/api-best-practices-hypermedia-part-3/)
+
+## {{ book.must }} Do Not Use Link Headers
```
dehora: `## {{ book.must }} Do Not Use Link Headers with JSON entities`

The intent of the guidance here makes sense, but I think it's worth qualifying.

hyper-media/Hypermedia.md
```diff
@@ -4,13 +4,47 @@
...
 
-Decoupled from HATEOAS, adding links and relationships should be considered as useful addition to APIs as it adds context and helps navigation in APIs to decent degree.
+Decoupled from HATEOAS, adding [links and relationships](#must-use-hal) should be considered as useful addition to APIs as it adds context and helps navigation in APIs to decent degree.
 
 Additional resources: [Wikipedia](http://en.wikipedia.org/wiki/HATEOAS), [The RESTful CookBook](http://restcookbook.com/Basics/hateoas/)
 
+## {{ book.must }} Use HAL
+
+Links to other resources must be defined exclusively using [HAL](http://stateless.co/hal_specification.html) and
+preferably using standard [link relations](http://www.iana.org/assignments/link-relations/link-relations.xml).
```
dehora: Having an idiomatic means to represent links makes sense.  It's possible just the way the guidance is worded, but taken literally, this means documents that have any property whose value is a URL are breaking guidance. Even HAL doesn't demand that.

I'm going to suggest a subset of HAL makes sense before committing to the full spec. For example, `_links` using `rel` values, no `_embedded`, no Curies, no URI Templates. This is compliant with HAL, but will greatly reduce the effort needed by clients to process hypertext data from servers while we gather more implementation experience.

In terms of client effort - HAL introduces three new specifications - HAL, which is not finalised, along with URI Templates and Curies. [RFC6570, URI Templates](https://tools.ietf.org/html/rfc6570) is not a small spec to implement. [Curies](https://www.w3.org/TR/2010/NOTE-curie-20101216/) are bringing in legacy baggage from another era  - and they're [not  clearly defined](https://groups.google.com/forum/#!msg/hal-discuss/p59KtwXnu78/tPl3L1r2oowJ) in the HAL spec. A lot of HAL is optional, but clients won't get to control what the optional parts of HAL they receive are, and will have to be more general than servers. 

Something worth bearing in mind is that HAL is a media type and not just a convention - it's expected by the RFC that clients and servers will negotiate with the `application/hal+json` type (JSON-LD also has a media type but it defines a workaround for other json types). It's not clear how this fits in with the guidelines elsewhere to use specific vendor media types.

hyper-media/Hypermedia.md
fmueller: So, the same what @ePaul mentioned. links header can be useful for non-JSON data.

hyper-media/Hypermedia.md
fmueller: An artificial use case could be that you traverse a directory of PDF files an wanna point the client to next PDF with links header. But I don't know if I would design an API like this...

We could add "for JSON" to this rule.

hyper-media/Hypermedia.md
fmueller: I think we should clarify what we mean by links. Because it mostly is about expressing relationships and not all urls - in my understanding. @whiskeysierra?

Good point with the subset idea @dehora I like that.

Is the media type a problem for us @whiskeysierra?

hyper-media/Hypermedia.md
```diff
@@ -4,13 +4,47 @@
...
+| [Siren](https://github.com/kevinswiber/siren)                  | ✗          | ✗        | Entities and navigation | ✗    | 2015-10-21  |
+| [Collection+JSON](http://amundsen.com/media-types/collection/) | ✗          | ✗        | Collections and queries | ✗    | 2013-02-24  |
+
+Interesting articles for comparisons of different hypermedia formats:
+* [Kevin Sookocheff’s On choosing a hypermedia type for your API](http://sookocheff.com/post/api/on-choosing-a-hypermedia-format/)
+* [Mike Stowe's API Best Practices: Hypermedia](http://blogs.mulesoft.com/dev/api-dev/api-best-practices-hypermedia-part-3/)
+
+## {{ book.must }} Do Not Use Link Headers
+
+We don't allow the usage of [link headers](http://tools.ietf.org/html/rfc5988#section-5) and favor [HAL](#must-use-hal) instead.
```
ePaul: As seen in the recent confusion in our internal chat, this might need to be written more clearly to make sure that only the `Link:` header defined in RFC 5988 is forbidden, not any other headers containing URIs.

fmueller: Good point.

hyper-media/Hypermedia.md
ePaul: What we actually want is something like this:

* Each JSON object (not just at root level, also for nested objects) can have a `_links` property, with links for each relation (conforming to HAL's spec). Links are (potentially relative) URI references, not templates to be filled in nor Curies. Same for link relation types.
* The API-relevant links are declared in the OpenAPI definition, though (as always per our compatibility rules) servers are allowed to extends this (and omit links which are not declared required).
* The media type part of HAL is ignored.

I'm not sure how to word this in guideline form.

hyper-media/Hypermedia.md
dehora: Maybe this

```
We don't allow the usage of the [Link Header](http://tools.ietf.org/html/rfc5988#section-5) defined in RFC 5988, and favor [HAL](#must-use-hal) instead.
```

hyper-media/Hypermedia.md
fmueller: I will prepare sth with Willi as both of us are in lead for this.

whiskeysierra:
I did some updates, please review again.

hyper-media/Hypermedia.md
```diff
@@ -4,13 +4,53 @@
...
+
+Links to other resources must be defined exclusively using [HAL](http://stateless.co/hal_specification.html) and
+preferably using standard [link relations](http://www.iana.org/assignments/link-relations/link-relations.xml).
+
+Clients and Servers are required to support `_links` with its `href` and `rel` attributes, not only at the root level
+but also in nested objects. To reduce the effort needed by clients to process hypertext data from servers it's not
+required to implement CURIEs, URI templates or embedded resources. Nor is it required to support the HAL media type
+`application/hal+json`.
+
+We opted for this subset of HAL after conduction a comparison of different hypermedia formats based on properties like:
```
fmueller: conducting a comparison?

hyper-media/Hypermedia.md
```diff
@@ -4,13 +4,53 @@
...
+| [Siren](https://github.com/kevinswiber/siren)                  | ✗          | ✗             | ✗        | Entities and navigation | ✗    |
+| [Collection+JSON](http://amundsen.com/media-types/collection/) | ✗          | ✗             | ✗        | Collections and queries | ✗    |
+
+Interesting articles for comparisons of different hypermedia formats:
+* [Kevin Sookocheff’s On choosing a hypermedia type for your API](http://sookocheff.com/post/api/on-choosing-a-hypermedia-format/)
+* [Mike Stowe's API Best Practices: Hypermedia](http://blogs.mulesoft.com/dev/api-dev/api-best-practices-hypermedia-part-3/)
+
+## {{ book.must }} Do Not Use Link Headers with JSON entities
+
+We don't allow the use of the [`Link` Header defined by RFC 5988 ](http://tools.ietf.org/html/rfc5988#section-5)
```
fmueller: remove extra space in closing bracket

fmueller:
👍

hyper-media/Hypermedia.md
```diff
@@ -4,13 +4,53 @@
...
 
+## {{ book.must }} Use a well-defined subset of HAL
+
+Links to other resources must be defined exclusively using [HAL](http://stateless.co/hal_specification.html) and
+preferably using standard [link relations](http://www.iana.org/assignments/link-relations/link-relations.xml).
+
+Clients and Servers are required to support `_links` with its `href` and `rel` attributes, not only at the root level
+but also in nested objects. To reduce the effort needed by clients to process hypertext data from servers it's not
+required to implement CURIEs, URI templates or embedded resources. Nor is it required to support the HAL media type
+`application/hal+json`.
```
dehora: `it's not required to implement` -> `it's not recommended to serve data with`

hyper-media/Hypermedia.md
```diff
@@ -4,13 +4,52 @@
...
 Additional resources: [Wikipedia](http://en.wikipedia.org/wiki/HATEOAS), [The RESTful CookBook](http://restcookbook.com/Basics/hateoas/)
 
+## {{ book.must }} Use a well-defined subset of HAL
+
+Links to other resources must be defined exclusively using [HAL](http://stateless.co/hal_specification.html) and
+preferably using standard [link relations](http://www.iana.org/assignments/link-relations/link-relations.xml).
+
+Clients and Servers are required to support `_links` with its `href` and `rel` attributes, not only at the root level
+but also in nested objects. To reduce the effort needed by clients to process hypertext data from servers it's not recommended to serve data with CURIEs, URI templates or embedded resources. Nor is it required to support the HAL media type
+`application/hal+json`.
```
ePaul: > Clients and Servers are required to support `_links` with its `href` and `rel` attributes, [...]

I'm not sure what "support" is meant to mean here.

Basically the API definition will define at which objects which links are to be expected, using the normal OpenAPI means of doing this (and have a description clarifying its meaning), and additional (unexpected) links in received data should always be ignored due to the [Robustness Principle rule](http://zalando.github.io/restful-api-guidelines/compatibility/Compatibility.html#must-prepare-clients-for-compatible-api-extensions-the-robustness-principle).
More generic support of links (i.e. with arbitrary, not in the OpenAPI definition mentioned link relation types) is only useful for some rare generic use cases.

dehora: > More generic support of links (i.e. with arbitrary, not in the OpenAPI definition mentioned link relation types) is only useful for some rare generic use cases.

I wonder if we have enough data or experience to say that existing link relations cover most of our use cases :)

The extension mechanism for rel links is problematic in HAL because like previous specs tries to make field names URIs which winds up with the CURIE model. CURIES and related variants are how this got dealt with in XML and XML-RDF and I don't think anyone can say it's been a slam dunk. CURIES mean servers will force extra engineering effort on clients  (plus it's not clear that curies are properly thought out in HAL, the impression I get from the HAL list is that the spec would move away them long term if there was a viable alternative). Now, the guidance in fairness does steer API designers away from imposing that cost on clients, but that leaves API designers in a bind - they can't use anything but HAL to express links, but are restricted to articulating only relations that already defined (or have to go through the IANA or other process to get their relations registered).  

The more I think about it, the more I think we should consider formally subsetting HAL or making HAL optional.

ePaul: I don't say "just use standardized link relations instead of custom ones", I say "the server will normally provide and the client will be normally using those link relations which are defined in the OpenAPI definition of that API". (For extensibility, also other links are allowed, but they will be ignored by unaware receivers.)

This allows using (and even implementing) the API from the OpenAPI definition without needing to know that it is based on HAL or anything else. The HAL-knowledge (i.e. "links from an object to related URIs should be in a `"_links": { ‹rel›: [{ "href": ‹URI› }, ...], ... }` wrapper) is just needed for the API designer.

(And yes, this would mean if curies are used, then they should only used in the exact format which is used in the API definition, so client and server can be Curie-agnostic. I guess using a full URI for custom link relations is clearer even then, so I'm not against forbidding their use.)

For the rare cases where at API definition time the relevant link relation types are not known, there is the `additionalProperties` property of OpenAPI' schema definition.

hyper-media/Hypermedia.md
```diff
@@ -4,13 +4,52 @@
...
+* Simplicity: resource link syntax and concepts are easy to understand and interpret for API clients.
+* Compatibility: introducing and adding links to resources is not breaking existing API clients.
+* Adoption: use in open-source libraries and tools as well as other companies
+* Docs: degree of good documentation
+
+<p></p>
+
+| Standard                                                       | Simplicity | Compatibility | Adoption | Primary Focus           | Docs |
+|----------------------------------------------------------------|------------|---------------|----------|-------------------------|------|
+| [HAL](http://stateless.co/hal_specification.html)              | ✓          | ✓             | ✓        | Links and relationships | ✓    |
```
dehora: HAL should be rescored to be an `✗` like the others - it's is not objectively simpler for clients when curies and uri templates are factored in.

hyper-media/Hypermedia.md
dehora: > I don't say "just use standardized link relations instead of custom ones", I say "the server will normally provide and the client will be normally using those link relations which are defined in the OpenAPI definition of that API". (For extensibility, also other links are allowed, but they will be ignored by unaware receivers.)

Got it. It'd be great to have an example Open API snippet in the text to make that clear.

> (And yes, this would mean if curies are used, then they should only used in the exact format which is used in the API definition, so client and server can be Curie-agnostic. I guess using a full URI for custom link relations is clearer even then, so I'm not against forbidding their use.)

A few things come to mind

- If we're saying that the API designer can declare their own relations in non-URI form via OpenAPI definitions (probably with some caveats to say don't redefine what's already in the IANA registry), then we should just outright say curies are not to be served to clients for the moment. 

- Given we're not going to recommend the full spec, I think HAL needs to be a should and not a must in the guidelines, again for the moment. After we have real world development experience with it, we can re-assess the recommendation level. 

- We should consider formally describing the subset of HAL and be in no doubt what parts we're using and not using. That doesn't need to be done through this pr - it can be in the guidelines as a follow on pr, or as an IETF draft the guidelines can point at. I'm happy to draft either option.

hyper-media/Hypermedia.md
ePaul: So maybe a "full HAL" line and another one for the subset we are using?

whiskeysierra: > another one for the subset we are using

*ZAL*? :stuck_out_tongue_winking_eye:

dehora: > ZAL?

@whiskeysierra don't tempt me :)

> So maybe a "full HAL" line and another one for the subset we are using?

@ePaul Sounds good - and if you don't want to overcomplicate this pr, happy to defer it to a later one.

tkrop:
Is the discussion finished? For me all changes are fine.

tkrop:
:+1:

fmueller:
@tkrop Can you pls give your upvote again? I merged changes from master branch.

fmueller:
👍

tfrauenstein:
👍

PR #71, opened 2016-05-24
update: limit resource number without hard upper limit

### LINKED ISSUE #28: Should we have a limit of resources?
Right now we have two [sections](http://zalando.github.io/restful-api-guidelines/naming/Naming.html#should-limit-of-resources) in the guidelines about limiting the number of resource/sub-resources to an arbitrary limit:

> Should: Limit of Resources
>
>To keep maintenance manageable, and to avoid violations of the “separation of concern” principle, an API should not expose more than 16 resources. If you exceed 16 resources, first check if you can split them into separate subdomains with distinct APIs.
>
>Should: Limit of Sub-Resource Levels
>
>There are main resources (with root url paths) and sub-resources (or “nested” resources with non-root urls paths). Use sub-resources if their life cycle is (loosely) coupled to the main resource, i.e. the main resource works as collection resource of the subresource entities. You should use <= 3 sub-resource (nesting) levels -- more levels increase API complexity and url path length. (Remember, some popular web browsers do not support URLs of more than 2000 characters)

I agree with limits, but I'm not sure whether we can have something as concrete as *16* in there. I'd consider 5 to be a lot of resources already. But there might be valid excuses for it.

tfrauenstein:
Fixes #28.

### DIFF
--- a/resources/Resources.md
+++ b/resources/Resources.md
@@ -50,7 +50,7 @@ However, if the resource can be accessed directly via its unique id, then the AP
 
 ## {{ book.should }} Limit of Resources
 
-To keep maintenance manageable, and to avoid violations of the “separation of concern” principle, an API should not expose more than 16 resources. If you exceed 16 resources, first check if you can split them into separate subdomains with distinct APIs.
+To keep maintenance and service evolution manageable, we should follow "functional segmentation" and "separation of concern" design principles and do not mix different business functionalities in same microservice API definition. In this sense the number of resources exposed via API should be limited - our experience is that a typical range for a functional focussed good API is some 4-8 different resources. There may be exceptions with more complex business domains that require more resources, but you should first check if you can split them into separate subdomains with distinct APIs.
 
 ## {{ book.should }} Limit of Sub-Resource Levels
 


ePaul:
Should we do a similar change for the "only three levels of nested resources" rule?

resources/Resources.md
```diff
@@ -50,7 +50,7 @@ However, if the resource can be accessed directly via its unique id, then the AP
 
 ## {{ book.should }} Limit of Resources
 
-To keep maintenance manageable, and to avoid violations of the “separation of concern” principle, an API should not expose more than 16 resources. If you exceed 16 resources, first check if you can split them into separate subdomains with distinct APIs.
+To keep maintenance and service evolution manageable, we should follow "functional segmentation" and "separation of concern" design principles and do not mix different business functionalities in same microservice API definition. In this sense the number of resources exposed via API should be limited - our experience is that a typical range for a functional focussed good API is some 4-8 different resources. There may be exceptions with more complex business domains that require more resources, but you should first check if you can split them into separate subdomains with distinct APIs.
```
whiskeysierra: Should we link terms like that to some specification/description, e.g. their [wikipedia article](https://de.wikipedia.org/wiki/Cross-Cutting_Concern#Separation_of_Concerns)?

whiskeysierra: > in same microservice API definition

I see *microservices* as an implementation detail of REST APIs. Is it important to talk about microservices here? Couldn't we just use e.g.

> in the same API definition

fmueller: I'm with @whiskeysierra in both points.

I would rephrase "our experience is that a typical range for a functional focussed good API is some 4-8 different resources" to something like "our experience is that a typical range of resources for a well-designed API is between 4 and 8".

tkrop: I would more like to see the term microservice as the foundation of this guideline. This rule here is exactly to give readers a hint how micro should be at least be interpreted. Anyhow, in my opinion a service should target to provide only a single resource. More than one resource should always trigger the question whether "separation of concerns" is violated because of laziness, etc and require good argumentation why the resources are only strongly related and not in a is-part-of relation that should be modeled by including it in the main resource.

Anyhow we should integrate this as soon as possible, because it points in the right direction.

ePaul: @tkrop I'm not sure if "get teams to make even smaller micro services" should be in the scope of the API guild (or the API guidelines).

(I think this might belong more to the architecture guild.)

whiskeysierra: I believe that the term *microservice* shouldn't not be mentioned anywhere in these guidelines, at all. We are talking explicitly about interfaces here, a microservice is an implementation.

fmueller:
Let's merge this and do further refinements in extra PRs.

fmueller:
👍

Issue #77, opened 2016-05-27
Rate limiting


bocytko:
I did not find any guideline for rate limiting yet. Would be nice to include this. Widely used headers include:

```
X-Rate-Limit-Limit
X-Rate-Limit-Remaining
X-Rate-Limit-Reset
Retry-After
```

PR #78, opened 2016-05-27
Guidance on 429 responses and Rate Limits.

dehora:
Describe options to use 'Retry-After' or 'X-RateLimit headers
in conjunction with a '429 Too Many Requests' response.

For #77

/ @bocytko @conorclifford @tlossen  @whiskeysierra

### DIFF
--- a/http/Http.md
+++ b/http/Http.md
@@ -150,6 +150,6 @@ The 'X-RateLimit' headers are:
 
 - `X-RateLimit-Limit`: The maximum number of requests that the client is allowed to make in this window.
 - `X-RateLimit-Remaining`: The number of requests allowed in the current window.
-- `X-RateLimit-Reset`: The time at which the current rate limit window will be reset in seconds as a [Unix timestamp](http://en.wikipedia.org/wiki/Unix_time).
+- `X-RateLimit-Reset`: The relative time in seconds when the rate limit window will be reset.
 
 The reason to allow both approaches is that APIs can have different needs. Retry-After is often sufficient for general load handling and request throttling scenarios and notably, does not strictly require the concept of a calling entity such as a tenant or named account. In turn this allows resource owners to minimise the amount of state they have to carry with respect to client requests. The 'X-RateLimit' headers are suitable for scenarios where clients are associated with pre-existing account or tenancy structures. 'X-RateLimit' headers are generally returned on every request and not just on a 429, which implies the service implementing the API is carrying sufficient state to track the number of requests made within a given window for each named entity.
--- a/http/Http.md
+++ b/http/Http.md
@@ -119,7 +119,7 @@ different HTTP methods on resources.
 | 422 | Unprocessable Entity - semantic error (as opposed to a syntax error which would usually trigger a 400) | POST, PUT, DELETE, PATCH |
 | 423 | Locked - Pessimistic locking, e.g. processing states | PUT, DELETE, PATCH |
 | 428 | Precondition Required - server requires the request to be conditional (e.g. to make sure that the “lost update problem” is avoided). | All |
-| 429 | Too many requests - the client does not consider rate limiting and sent too many requests | All |
+| 429 | Too many requests - the client does not consider rate limiting and sent too many requests. See ["Use 429 with Headers for Rate Limits"](#must-use-429-with-headers-for-rate-limits). | All |
 
 ###Server Side Error Codes:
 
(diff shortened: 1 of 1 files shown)


http/Http.md
```diff
@@ -137,3 +137,19 @@ APIs should define the functional, business view and abstract from implementatio
 
 The OpenAPI specification shall include definitions for error descriptions that will be returned; they are part of the interface definition and provide important information for service clients to handle exceptional situations and support troubleshooting. You should also think about a troubleshooting board — it is part of the associated online API documentation, provides information and handling guidance on application-specific errors and is referenced via links of the API definition. This can reduce service support tasks and contribute to service client and provider performance.
 
+
+## {{ book.must }} Use 429 and Headers to describe Rate Limits
+
+APIs that wish to manage the request rate of clients must use the ['429 Too Many Requests'](http://tools.ietf.org/html/rfc6585) response code. Such responses must also contain header information providing further details to the client. There are two approaches a service can take for header information -
```
ePaul: (Editorial.)

That might be a cultural difference, but I would use a colon (`:`) here, not a hyphen (`-`) (neither a dash).

ePaul: > APIs that wish to manage the request rate of clients must use the ['429 Too Many Requests'](http://tools.ietf.org/html/rfc6585) response code.

This needs a conditional subphrase like "if the client exceeded the request rate and therefore the request can't be fulfilled". Otherwise it could be read as "respond all requests with 429", which clearly is not meant.

http/Http.md
```diff
@@ -137,3 +137,19 @@ APIs should define the functional, business view and abstract from implementatio
...
+
+The 'X-RateLimit' headers are -
+
+- `X-RateLimit-Limit`: The maximum number of requests that the client is allowed to make in this window.
+- `X-RateLimit-Remaining`: The number of requests allowed in the current window.
+- `X-RateLimit-Reset`: The time at which the current rate limit window will be reset in seconds as a [Unix timestamp](http://en.wikipedia.org/wiki/Unix_time).
+
+The reason to allow both approaches is that APIs can have different needs. Retry-After is often sufficient for general load handling and request throttling scenarios and notably, does not strictly require the concept of a calling entity such as a tenant or named account. In turn this allows resource owners to minimise the amount of state they have to carry with respect to client requests. The 'X-RateLimit' headers are suitable for scenarios where clients are associated with pre-existing account or tenancy structures. 'X-RateLimit' headers are generally returned on every request and not just on a 429, which implies the service implementing the API is carrying sufficient state to track the number of requests made within a given window for each named entity.
```
ePaul: Just for understanding: What requests does this count? All requests to the API (i.e. all operations described in one OpenAPI definition), just the one operation (i.e. path + method combination), or something else? Or can each API define this itself? Or would this depend on the individual contract with the client, and doesn't need to be in the  API description?

bocytko: I would leave the services freedom to define how exactly rate limiting is implemented and which operations are limited and how (also incl. client contracts).

ePaul:
Another note: The HTTP response code list (in the same file) should have a link to this section where 429 is mentioned.

http/Http.md
dehora: @ePaul typically the `X-RateLimit` headers  count requests and the scope of the count is bound to some kind of account/tenant entity that is presenting itself by name/token. Some applications will rate count limit parts of their system and not others, or apply multiple sets of limits - how that works case by case is probably outside general guidance on how to present limits.  [Twitter](https://dev.twitter.com/rest/public/rate-limiting), [Github](https://developer.github.com/v3/rate_limit/), or [Intercom](https://developers.intercom.io/reference/#rate-limiting) are working examples. 

Retry-After can be a generic backoff that works across all resources and clients (eg when the system is under load and wants to throttle access), or it can be scoped to user agents/ip addresses/etc, HTTP doesn't say what its scope is, but there's less implication of tracking per entity/account/tenant. Looking at it a bit more broadly, Retry-After isn't strictly tied to prior requests (eg you could return it based on a physical load or network bandwidth).

dehora:
Feedback addressed in 7633735. I'll rebase later as a follow on.

http/Http.md
```diff
@@ -137,3 +137,19 @@ APIs should define the functional, business view and abstract from implementatio
...
+
+ - Return a ['Retry-After'](https://tools.ietf.org/html/rfc7231#section-7.1.3) header indicating how long the client ought to wait before making a follow-up request. The Retry-After header can contain a HTTP date value to retry after or the number of seconds to delay. Either is acceptable but APIs should prefer to use a delay in seconds.
+ 
+ - Return a trio of 'X-RateLimit' headers. These headers (described below) allow a server to express a service level in the form of a number of allowing requests within a given window of time and when the window is reset. 
+
+The 'X-RateLimit' headers are:
+
+- `X-RateLimit-Limit`: The maximum number of requests that the client is allowed to make in this window.
+- `X-RateLimit-Remaining`: The number of requests allowed in the current window.
+- `X-RateLimit-Reset`: The time at which the current rate limit window will be reset in seconds as a [Unix timestamp](http://en.wikipedia.org/wiki/Unix_time).
```
conorclifford: not sure it makes sense to use an absolute time for this - its highly unlikely that all servers involved in client/service relationships would be time-synchronous to the second (far from it on occasion).

A relative time (number of seconds until limit is reset) would be more practical.

dehora: @conorclifford  :+1: updated in 9586fea

http/Http.md
```diff
@@ -137,3 +137,19 @@ APIs should define the functional, business view and abstract from implementatio
...
 
+
+## {{ book.must }} Use 429 with Headers for Rate Limits
+
+APIs that wish to manage the request rate of clients must use the ['429 Too Many Requests'](http://tools.ietf.org/html/rfc6585) response code if the client exceeded the request rate and therefore the request can't be fulfilled. Such responses must also contain header information providing further details to the client. There are two approaches a service can take for header information:
+
+ - Return a ['Retry-After'](https://tools.ietf.org/html/rfc7231#section-7.1.3) header indicating how long the client ought to wait before making a follow-up request. The Retry-After header can contain a HTTP date value to retry after or the number of seconds to delay. Either is acceptable but APIs should prefer to use a delay in seconds.
```
fmueller: If api returns 'HTTP date value' should this also be compliant with our guidelines for date fields in JSON? I don't know what you mean with HTTP dates.

ePaul: "HTTP date" is a (quite ugly) date format which is common for HTTP headers. See [Use standard date and time formats](http://zalando.github.io/restful-api-guidelines/data-formats/DataFormats.html#must-use-standard-date-and-time-formats).

whiskeysierra: > is a (quite ugly) date format

That's actually an understatement.

dehora: > If api returns 'HTTP date value' should this also be compliant with our guidelines for date fields in JSON? I don't know what you mean by HTTP dates.

@fmueller @ePaul @whiskeysierra This header allows sending the time as a date structure (which is often considered clunky) or a time in seconds. 

```
  Retry-After: Weds, 1st June 2016 22:26:00 GMT
  Retry-After: 60
```

As a matter of style the format in seconds is arguably simpler while conveying the same information, hence the bias in the recommendation (you could argue there's room for errors around timezoning with date strings maybe), but HTTP as it stands allows the date representation as well so I'm not inclined to say it simply can't be used, plus http libraries can generally handle it.

ePaul: @fmueller `Retry-After: 60` is not a "date in seconds", but a time difference (in seconds). It effectively says "try again 60 seconds after you received this response". That works without any need to synchronize clocks. (The other format works also without synchronized clocks, but then you need to consider both the `Date` and the `Retry-After` header.)

fmueller:
👍

fmueller:
@ePaul @whiskeysierra any things to add? if not, please thumbvote :)

whiskeysierra:
:+1:

PR #80, opened 2016-05-30
Fix link pointing to wrong chapter

henkmeulekamp:
Support Filtering of Resource Fields link repaired.
  
Naming section **Could: Use Conventional Query Strings** has a link for filtering pointing the wrong location,


ePaul:
Thanks for the contribution. Could you please change the link to use the `.md` file instead `.html`? That way it will work also in the Github file view, and the HTML generator will change the link anyways.

henkmeulekamp:
Ah sorry, missed that. Should be fixed now.

ePaul:
:+1:

PR #87, opened 2016-06-08
Events: Use the hash strategy for data change events

dehora:
This guidance tells people to use the hash partition strategy for data change events, unless they have a very good reason. Nakadi defaults to `random`, but `hash` is almost always a good default for data change and this text should guide people towards the right option.

Making it a must is an option, but I suspect we'd need an escape valve of some kind to allow exceptional cases where producer assigned/random makes sense, so left it as a should with a warning :)

### DIFF
--- a/events/events.md
+++ b/events/events.md
@@ -50,7 +50,13 @@ Nakadi defines an event for signalling data changes, called a [DataChangeEvent](
 
 - Change events must be published reliably by the service.
 
-- When the `hash` partition strategy is used, change events should identify a partition key that allows all related events for the entity to be consistently to assigned to a partition.
+## {{ book.should }} Use the hash partition strategy for Data Change Events
+
+The `hash` partition strategy allows a producer to define which fields in an event are used as input to compute the partition the event should be added to. 
+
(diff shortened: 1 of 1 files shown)


whiskeysierra:
:+1:

tkrop:
Hi Bill, I'm a bit concerned that this paragraph will be interpreted as a proposal to use partitioning as default. I would strongly recommend users to stay away from this idea as long as partitioning is avoidable. The problems and complexities of partitions are manifold, while the general throughput of Kafka is so hug, that your rarely should need this option for usual backend data change events.

Can you include a preceding section or paragraph that guide users to avoid partitions until really needed for throughput purposes?

tfrauenstein:
👍   for the guideline; additional comment...

a) Today "random" is the default partitioning strategy of Nakadi. IMO it should be changed to "hash" -- pls. discuss with Nakadi team. I also would give hint about default in the guideline for info. 

b) A different thing is how partitions are defined in Nakadi -- I think it is worth a separate guidance.
Today number of partition cannot specified directly with event type. It is inferred from load related event definition params (msg size, write rate, write parallelism, read parallelism). These are optional params with low load and partitions=1 as default. Since (often) re-partitioning should be avoided we should give guidance, that load params should be dimensioned such that peak loads with forecast of 1-2y are covered. We also should give more hints and examples of how the params should be interpretated and estimated. I see this as part of the Nakadi User Manual that you currently are writing and not as part of the Event API guidelines. In the rule above, this manual may be referenced.

fmueller:
👍

PR #89, opened 2016-06-09
Use the Prefer header to indicate processing preferences.

dehora:
The Prefer header is defined in [RFC7240](https://tools.ietf.org/html/rfc7240) and allows a client to
indicate request processing behaviors from servers. This recommends
using the Prefer header rather than defining a proprietary "X-"
header for scenarios where processing can be qualified (eg such
as strict/lax error handling, or sync/async responses). It also
recommends an extension preference called  `respond-sync` to ask
for synchronous processing resulting in responses other than `202`.

### DIFF
--- a/headers/CommonHeaders.md
+++ b/headers/CommonHeaders.md
@@ -1,6 +1,6 @@
 # Common Headers
 
-This section describes a handful of headers, which we found raised the most questions in our daily usage.
+This section describes a handful of headers, which we found raised the most questions in our daily usage, or which are useful in particular circumstances but not widely known.
 
 ## Content-Location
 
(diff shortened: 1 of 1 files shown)


whiskeysierra:
+1

fmueller:
I really like the concept. On the other hand I'm afraid we make our guidelines to detailed and I doubt that this will be implemented by any API of ours.

conorclifford:
I also really like this, and will definitely be using it several APIs.

dehora:
> I really like the concept. On the other hand I'm afraid we make our guidelines to detailed and I doubt that this will be implemented by any API of ours.

@fmueller :+1: it was a bit verbose, I've trimmed it down. It's also marked as discussion, so plenty of time  to think it through. The reason to add is that it async vs sync processing directives came up in an api review; also the header is not that well known, probably because it's not part of the main HTTP RFC (I only found it after doing some digging).

fmueller:
You got me. ;-)

fmueller:
👍

whiskeysierra:
:+1:

dehora:
😍 awesome  :)

dehora:
related: https://www.ietf.org/id/draft-dehora-respond-sync-00.txt

PR #91, opened 2016-06-10
added first version of a paragraph about API deprecation

codekultur:
...please heavily comment on this one... Is this the right place in the guidelines?


kakawait:
What do you think about adding a `Warning` header with [299 Miscellaneous Persistent Warning](https://tools.ietf.org/html/rfc7234#section-5.5.7) warn code that specify that API `/a` is deprecated and will removed on a given date and you can switch to the new API `/b`.

ePaul:
@kakawait In the last meeting (from which this PR resulted) we talked about adding a header with a deprecation note – it looks like the `Warning` header is like the correct thing to use here. Thanks for the note.

I guess the Warning should just say something like "The path/operation/parameter/... ‹name› is deprecated, please see ... for details" with a link to a web site explaining the deprecation – usually the replacement is not to be used in the exact same way (because then you wouldn't need to deprecate the old API), so having some place to read about the differences is a good thing.

codekultur:
Thanks @kakawait and @ePaul for pointing this out. Yes, we talked about this in the meeting and I totally forgot to add a sentence about the Warning header... I'll fix this and add it to the proposal...

kakawait:
@ePaul Yes you're completely right about description content that should point to *documentation* that explain everything

---

If needed I write a tool (for *Java* - `Spring`) to manage such `Warning` https://github.com/kakawait/spring-response-warning (quite young)

codekultur:
I added a proposal for adding the Warning header. I'm unsure if it should be a "should" or "may" - especially because it's not sufficient adding the header for gaining client consent. What's your opinion?

kakawait:
In our case (specific to our app, because we are providing an infrastructure that other developers build apps on it) we are using `Warning` headers but during *integration phase* a client code is spawning a web modale or `console.warn` (depending on configuration). But I don't think that way is possible in any team/app/situation.

kakawait:
@codekultur and maybe add some reference about [RFC 7234 - Warning header](https://tools.ietf.org/html/rfc7234#section-5.5) because regarding the *RFC* `Warning` header has a specific format

```
Warning       = 1#warning-value

warning-value = warn-code SP warn-agent SP warn-text
                                   [ SP warn-date ]

warn-code  = 3DIGIT
warn-agent = ( uri-host [ ":" port ] ) / pseudonym
             ; the name or pseudonym of the server adding
             ; the Warning header field, for use in debugging
             ; a single "-" is recommended when agent unknown
warn-text  = quoted-string
warn-date  = DQUOTE HTTP-date DQUOTE
```

codekultur:
Thanks @kakawait for you great contribution. I added a link to the RFC and stated, that the code must be 299. As far as I understand, this should be the only code that fits this warning. Do you agree?

kakawait:
Except if you wanna deprecate an API just for a moment (I don't have use case) you should use `199` lol

compatibility/Compatibility.md
```diff
@@ -143,3 +143,15 @@ are still under review; for example:
 Hint: This versioning scheme differs in the less strict DRAFT aspect from
 [semantic version information](http://semver.org) used for released APIs and
 service applications.
+
+## Deprecating an API
+
+Sometimes it is necessary to phase out an API endpoint (or version). I.E. this may be necessary if a field is no longer supported in the result or a whole business functionality behind an endpoint has to be shut down. There are multiple other reasons as well.
+
+First of all the provider must make sure, that there is no downwards compatible way to maintain the API any longer.
```
mostlymagic: instead of "must make sure" I'd prefer "should be certain"

compatibility/Compatibility.md
```diff
@@ -143,3 +143,15 @@ are still under review; for example:
 Hint: This versioning scheme differs in the less strict DRAFT aspect from
 [semantic version information](http://semver.org) used for released APIs and
 service applications.
+
+## Deprecating an API
+
+Sometimes it is necessary to phase out an API endpoint (or version). I.E. this may be necessary if a field is no longer supported in the result or a whole business functionality behind an endpoint has to be shut down. There are multiple other reasons as well.
```
mostlymagic: I.e.

mostlymagic: a whole business functionality -> an entire business functionality

mostlymagic: "There are multiple other reasons as well." -> this sentence is pretty worthless unless we supply valid scenarios

fmueller:
@codekultur Can you please prepare wording fixes as suggested? Then we can merge this.

codekultur:
@fmueller we should wait for the next guild meeting. I just added a paragraph about how to document api deprecation in swagger files, which has to be discussed...

compatibility/Compatibility.md
```diff
@@ -143,3 +143,18 @@ are still under review; for example:
...
+
+Before shutting down an API (or version of an API) the producer must make sure, that all clients have given their consent to shut down the endpoint. Producers should help consumers to migrate to a potential new endpoint (i.e. by providing a migration manual). After all clients are migrated, the producer may shut down the deprecated API.
+
+API deprecation must reflect in the producers OpenAPI definition. If a method on a path, a whole path or even a whole API endpoint (multiple paths) should be deprecated, the producers must set `deprecated=true` on each method / path element that will be deprecated (OpenAPI 2.0 only allows you to define deprecation on this level). If deprecation should happen on a finer level (i.e. query parameter, payload etc.), the producer should set `deprecated=true` on the affected method / path element and add further explanation to the `description` section.
+If `deprecated` is set to `true`, the producer must describe what clients should use instead and when the API will be shut down in the `description` section of the API definition.
```
tkrop: I think we should insert a short example here. Even the OpenAPI specification has no example to guide readers.

tkrop:
@codekultur @fmueller good work. I like the current wording. Short and precise.

whiskeysierra:
:+1:

codekultur:
After the last API guild meeting, I'm unsure, if we should keep the paragraph like this or change it to the MUST/SHOULD headline style. What do you think?

tkrop:
:+1:

tkrop:
merge it as is. than start new discussion.

PR #98, opened 2016-06-28
Add environment specific host config to api discovery topic

fmueller:
To support multiple environments (test system, production) the crawlable API definitions need to support this. If an API definition from staging system is crawled, the endpoint should be the staging system and not production.

### DIFF
--- a/api-discovery/ApiDiscovery.md
+++ b/api-discovery/ApiDiscovery.md
@@ -21,3 +21,12 @@ API Viewer (Swagger-UI) for central access to all discovered API definitions.
 For the time being, this document is an appropriate place to mention this rule, even though it is
 not a RESTful API definition rule or related to our STUPS infrastructure for application service
 management.
+
+## {{ book.should }} Enable environment specific host configuration
+
+With [Twintip](http://docs.stups.io/en/latest/components/twintip.html) we also provide a central
+Swagger UI installation from which every API can be used. Every service which implements an API
+should therefore support environment specific host configuration in its OpenAPI definition:
+the endpoint where the service is running should express the actual environment, e.g. for a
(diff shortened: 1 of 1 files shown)


ePaul:
This is about the `host` property? My solution for this would be to omit the `host` part entirely from the API definition (it defaults to the host from which you are retrieving the definition).

fmueller:
Also a good idea. @whiskeysierra already invented some templating for this. Let us catch up with him and his team.

whiskeysierra:
> This is about the host property? My solution for this would be to omit the host part entirely from the API definition (it defaults to the host from which you are retrieving the definition).

Interesting, I didn't know that. What we are doing in twintip-spring-web is to configure our base url and then change the `host`, `schemes` and `basePath` based on this: https://github.com/zalando-stups/twintip-spring-web/blob/master/src/main/java/org/zalando/twintip/spring/SchemaResource.java#L111

ePaul:
The omitting of the host might be problematic for auto-discovery, though ... I guess Twintip has no idea that the service serves at https://warehouse-allocation.wholesale.zalan.do/ when crawling the APIs from a list of deployed micro services for team Lumberjack. (How exactly does Twintip know where too look? Does it get a list of IP addresses from somewhere?)

fmueller:
Twintip crawls the endpoint information from the application registry KIO. The endpoint which is persisted there is used.

I think, we can reject this change request for the guidelines. It looks like something Twintip can handle by itself (and should as not every API definition will be done in a host-specific way either).

hjacobs:
@fmueller I don't really get what this PR is about and the proposed text does not tell me much --- therefore I would advice to close this PR.

PR #99, opened 2016-07-01
Restore changes from #25

ePaul:
Pull request #25 was merged, but then the change was soon later lost [in a broken local merge](https://github.com/zalando/restful-api-guidelines/commit/d969093cfbbf8f7697184549cd45395059677a6e), it seems.

This has the same changes as #25 cherry-picked into a new branch.

### DIFF
--- a/http/Http.md
+++ b/http/Http.md
@@ -24,12 +24,15 @@ Be compliant with the standardized HTTP method semantics summarized as follows:
 ###PATCH:
 
 - partial upload, i.e. only a specific subset of resource fields are replaced
-- partial resource representation passed as payload has either resource content type with optional fields (to be
-  updated) or a custom content type that also may include instructions of how to change the resource
-- usually accepted only by single resources, because the semantics for PATCH on a collection resource
-  are very hard to define
 - usually not robust against non existence of the entity
+- since implementing PATCH correctly is a bit tricky we strongly suggest go for the following:
 
+1. use PUT with a custom media type and only if that isn't sufficient anymore
+2. use PATCH with [JSON Merge Patch](https://tools.ietf.org/html/rfc7386), a specialized media type 
+   `application/merge-patch+json` that is a partial resource representation. If updates are too complex to be expressed
+   with a merge patch, you should
+3. use PATCH with [JSON Patch](http://tools.ietf.org/html/rfc6902), a specialized media type 
+   `application/json-patch+json` that includes instructions on how to change the resource
 
 ###DELETE:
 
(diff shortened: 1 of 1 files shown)


fmueller:
👍

http/Http.md
```diff
@@ -21,24 +21,28 @@ Be compliant with the standardized HTTP method semantics summarized as follows:
...
-- partial resource representation passed as payload has either resource content type with optional fields (to be
-  updated) or a custom content type that also may include instructions of how to change the resource
-- usually accepted only by single resources, because the semantics for PATCH on a collection resource
-  are very hard to define
 - usually not robust against non existence of the entity
-
-
-###DELETE:
+- since implementing PATCH correctly is a bit tricky we strongly suggest to choose one and only one of the following patterns per endpoint, unless forced by [a backwards compatible change](../compatibility/Compatibility#compatibility):
+    1. use PUT with a custom media type and only if that isn't sufficient anymore
```
tfrauenstein: Unclear message of this sentence: When to use "PUT with customer media type" or not?
We should also make preference statement on the variants (in case they are applicable).
I would prefer "PATCH with JSON Merge" before "PUT with customer media type".

ePaul:
:+1:

http/Http.md
ePaul: "custom", not "customer" ;-)

The preference order was meant to be given by the order and the phrases on the end of each paragraph referring to the next one.

The first point basically means to prefer to send a full object (of whatever media type you are usually using for your GET/and creation PUT) to replace the full object, not doing a delta-update. This might need to be worded differently so it is clearer, though.

tfrauenstein: ok, this definitely should be reworded -- both, 1st point (as you described) and preference order (mention it before itemization and terminate phrases on the end of each paragraph referring to the next one by "..."). Thx.

tkrop: How about:

1. use PUT with complete objects to update a resource as long as feasible.
2. use PUT with partial objects identified by custom media types to only update parts of a resource.
3. ... as before

ePaul: @tkrop In point 2. I would use PATCH, not PUT – otherwise you are changing the meaning of PUT. Otherwise, sounds fine.

I just noted that I did do this in my fork, so I guess I'll have to do any rewordings.

whiskeysierra:
@ePaul Are you working on this?

ePaul:
@whiskeysierra I should be, but I totally forgot this.

ePaul:
@tfrauenstein @tkrop I reworded this section, I hope it now confirms to your wishes?

tfrauenstein:
:+1   @ePaul thx.

fmueller:
@tfrauenstein Your upvote was not complete. Please only a message with 👍  in it.

fmueller:
👍

tfrauenstein:
👍

PR #100, opened 2016-07-05
Added paragraph about technical status codes

### LINKED ISSUE #42: Don't specify all possible HTTP status codes

whiskeysierra:
This fixes #42

### DIFF
--- a/compatibility/Compatibility.md
+++ b/compatibility/Compatibility.md
@@ -46,6 +46,7 @@ How to do this:
 * Be prepared for new enum values declared with x-extensible-enum (see above);
   provide default behavior for unknown values, if applicable
 * Follow the redirect when the server returns an “HTTP 301 Moved Permanently” response code
+* Be prepared for unexpected status codes, and in case of doubt handle them like you would handle the corresponding x00 code
 
 ## {{ book.should }} Avoid Versioning
 
--- a/compatibility/Compatibility.md
+++ b/compatibility/Compatibility.md
@@ -46,7 +46,7 @@ How to do this:
 * Be prepared for new enum values declared with x-extensible-enum (see above);
   provide default behavior for unknown values, if applicable
 * Follow the redirect when the server returns an “HTTP 301 Moved Permanently” response code
-* Be prepared for unexpected status codes, and in case of doubt handle them like you would handle the corresponding x00 code
+* Be prepared to handle HTTP status codes not explicitly specified in endpoint definitions! Note also, that status codes are extensible -- default handling is how you would treat the corresponding x00 code (see [RFC7231  Section 6](https://tools.ietf.org/html/rfc7231#section-6))
 
 ## {{ book.should }} Avoid Versioning
 
--- a/http/Http.md
+++ b/http/Http.md
@@ -161,7 +161,7 @@ Service providers should differentiate between technical and functional errors. 
 
 Even though they might not be documented - they may very much occur in production, so clients should be prepared for unexpected response codes, and in case of doubt handle them like they would handle the corresponding x00 code. Adding new response codes (specially error responses) should be considered a compatible API evolution.
 
-Functional errors on the other hand, that convey domain-specific semantics, must be documented and are are strongly encouraged to be expressed with [*Problem types*](../common-data-objects/CommonDataObjects.md#must-use-common-error-return-objects).
+Functional errors on the other hand, that convey domain-specific semantics, must be documented and are strongly encouraged to be expressed with [*Problem types*](../common-data-objects/CommonDataObjects.md#must-use-common-error-return-objects).
 
 ## {{ book.must }} Use 429 with Headers for Rate Limits
 
--- a/http/Http.md
+++ b/http/Http.md
@@ -136,14 +136,32 @@ different HTTP methods on resources.
 | 501 | Not Implemented -  server cannot fulfill the request (usually implies future availability, e.g. new feature). | All |
 | 503 | Service Unavailable - server is (temporarily) not available (e.g. due to overload) -- client retry may be senseful. | All |
 
-All error codes you can find in [RFC7231](https://tools.ietf.org/html/rfc7231#section-6) and [Wikipedia](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) or via https://httpstatuses.com/<error_code>.
+All error codes can be found in [RFC7231](https://tools.ietf.org/html/rfc7231#section-6) and [Wikipedia](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) or via https://httpstatuses.com/<error_code>.
 
 ## {{ book.must }} Providing Error Documentation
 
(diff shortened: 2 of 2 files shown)


whiskeysierra:
Pull request #100

http/Http.md
```diff
@@ -121,22 +120,40 @@ different HTTP methods on resources.
...
 
 The OpenAPI specification shall include definitions for error descriptions that will be returned; they are part of the interface definition and provide important information for service clients to handle exceptional situations and support troubleshooting. You should also think about a troubleshooting board — it is part of the associated online API documentation, provides information and handling guidance on application-specific errors and is referenced via links of the API definition. This can reduce service support tasks and contribute to service client and provider performance.
 
+Service providers should differentiate between technical and functional errors. In most cases it's not useful to document technical errors that are not in control of the service provider unless the status code convey application-specific semantics. The list of status code that can be omitted from API specifications includes but is not limited to:
+- `401 Unauthorized`
+- `403 Forbidden`
+- `404 Not Found` unless it
```
ePaul: It looks like you wanted to write something here?

mostlymagic:
lgtm

http/Http.md
```diff
@@ -121,22 +120,40 @@ different HTTP methods on resources.
...
 
 ## {{ book.must }} Providing Error Documentation
 
-APIs should define the functional, business view and abstract from implementation aspects. Errors become a key element providing context and visibility into how to use an API. The error object should be extended by an application-specific error identifier since the HTTP status code often is not specific enough to articulate the domain error situation. For this reason, we use a standardized error return object definition — see [*Use Common Error Return Objects*](../common-data-objects/CommonDataObjects.md#must-use-common-error-return-objects).
+APIs should define the functional, business view and abstract from implementation aspects. Errors become a key element providing context and visibility into how to use an API. The error object should be extended by an application-specific error identifier if and only if the HTTP status code is not specific enough to convey the domain-specific error semantic. For this reason, we use a standardized error return object definition — see [*Use Common Error Return Objects*](../common-data-objects/CommonDataObjects.md#must-use-common-error-return-objects).
```
tkrop: @whiskeysierra  "convey" is a very unusual word ... transmit, transport, communicate are in my opinion more suitable alternatives.

http/Http.md
```diff
@@ -121,22 +120,40 @@ different HTTP methods on resources.
...
+- `414 URI Too Long`
+- `415 Unsupported Media Type`
+- `500 Internal Server Error`
+- `502 Bad Gateway`
+- `503 Service Unavailable`
+- `504 Gateway Timeout`
+
+Even though they might not be documented - they may very much occur in production, so clients should be prepared for unexpected response codes, and in case of doubt handle them like they would handle the corresponding x00 code. Adding new response codes (specially error responses) should be considered a compatible API evolution.
+
+Functional errors on the other hand, that convey domain-specific semantics, must be documented and are are strongly encouraged to be expressed with [*Problem types*](../common-data-objects/CommonDataObjects.md#must-use-common-error-return-objects).
```
tkrop: are are -> are

compatibility/Compatibility.md
```diff
@@ -46,6 +46,7 @@ How to do this:
 * Be prepared for new enum values declared with x-extensible-enum (see above);
   provide default behavior for unknown values, if applicable
 * Follow the redirect when the server returns an “HTTP 301 Moved Permanently” response code
+* Be prepared for unexpected status codes, and in case of doubt handle them like you would handle the corresponding x00 code
```
tkrop: I think we should extend this guideline:

* Be prepared to handle all possible status codes, known as well as unknown.

> In case of doubt handle them like you would handle the corresponding x00 code.

I'm not sure that I like to encourage clients to be super smart - they tend to fail in unexpected harmful ways -, so I suggest to drop the second part of the sentence. Better we add an additional sentence to discourage super smart clients.

My rule of thumb: fail and assume constant failure unless the failure indicates that a client may retry later successful. The only viable reason retrying are connection problems or when the server indicates that a retry will make sense later. Being smart and trying to compensate for failures will make clients as well as business logic unnecessary complex and mostly results in unexpected failures that are hard to track to their source.

whiskeysierra: @ePaul What's your opinion? Since I took that phrase almost word-by-word from you :stuck_out_tongue_winking_eye:

tfrauenstein: basically fine, just 2 comments:

1) editorial:
- "Be prepared for unexpected status codes" --> "Be prepared to handle status codes not explicitly specified in endpoint definition"
- add link to "Providing Error Documentation" rule below

2) @tkrop I do not understand your comment on super smart client, above: 
"In case of doubt handle them like you would handle the corresponding x00 code" does not sound super smart and like a simple valid hint. 
You mentioned retry behaviour of error handling -- this is a specific aspect that is not really addressed by the rule. I propose to have here a separate discussion on whether we want to give guidance here.

ePaul: @whiskeysierra @tkrop I think I paraphrased that sentence [from the HTTP standard](https://tools.ietf.org/html/rfc7231#section-6):

>   HTTP status codes are extensible.  HTTP clients are not required to
   understand the meaning of all registered status codes, though such
   understanding is obviously desirable.  However, a client MUST
   understand the class of any status code, as indicated by the first
   digit, and treat an unrecognized status code as being equivalent to
   the x00 status code of that class, with the exception that a
   recipient MUST NOT cache a response with an unrecognized status code.

tfrauenstein: @ePaul  thx, I like this reference -- propose to add it...

"Be prepared for unexpected status codes, ..." --> 
"Be prepared to handle HTTP status codes not explicitly specified in endpoint definitions! Note also, that status codes are extensible -- default handling is how you would treat the corresponding x00 code (see https://tools.ietf.org/html/rfc7231#section-6)."

Incl. minor editorial update of corresponding phrase in "Providing Error Documentation" rule below.

tkrop: @tfrauenstein sorry for taking this up again. I would go one step beyond the HTTP standard that we should warn developers explicitly not to handle `4xx` or `5xx` different from a failure, unless they can argue for this with very good reasons.

Failure documentation and sensible response codes are important for clients to learn what they did wrong - or not -, but service consumers should not overestimate the value of this information.

Handling `4xx` and `5xx` as something different than a failure, is what I meant with super smart clients/service consumers.

tkrop:
:+1:

compatibility/Compatibility.md
fmueller: @whiskeysierra @ePaul Would some of you would like to incorporate the suggested editorial changes of @tfrauenstein? I think then we can merge this long-open PR finally. :-)

tkrop:
:+1:

fmueller:
👍

whiskeysierra:
@tkrop @fmueller I had to fix a merge conflict. Please re-approve.

fmueller:
👍

tkrop:
:+1:

PR #104, opened 2016-07-07
Improved description of uid pseudo scope

tfrauenstein:


security/Security.md
```diff
@@ -13,25 +13,28 @@ securityDefinitions:
...
 ```
 
-In order to see how to define and assign scopes, please refer to the next section.
+The example defines OAuth2 with implicit flow as security standard used for authentication when accessing endpoints; additionally, there are two API access rights defined via the scopes section for later endpoint authorization usage - please see next section.
 
-## {{ book.must }} Define and assign scopes
+## {{ book.must }} Define and Assign Access Rights (Scopes)
 
-Every API needs to define scopes and every endpoint needs to have at least one scope assigned. Scopes are defined by a name and a description once per API specification, as shown in the previous section. Please refer to the following rules when creating scope names:
+Every API needs to define access rights, called scopes here, and every endpoint needs to have at least one scope assigned. Scopes are defined by a name and a description once per API specification, as shown in the previous section. Please refer to the following rules when creating scope names:
```
fmueller: editorial: by name and description per API specifiction

security/Security.md
```diff
@@ -57,4 +60,20 @@ paths:
...
+  /public-information:
+    get:
+      summary: Provides public information about ... 
+               Accessible by any user; no access rights needed. 
+      security:
+        - oauth2:
+          - uid
+```
+
+Hint: you need not explicitly define the "Authorization" header; it is a standard header so to say implicitly defined via the security section.
```
fmueller: Declaring the header explicitly enables easy access_token support in Swagger UI later. This would help integration in central Swagger UI installation a lot. For that reason, I would reverse the hint.

tfrauenstein: Almost no API has defined it explicitly, since it is standard and implicitly defined via security section. 
Even if we require it, it will not always be provided and we should not restrict API discovery experience, if feasible. 
Let us discuss this tooling problem offline.

whiskeysierra:
:+1:

security/Security.md
tfrauenstein: done

security/Security.md
fmueller: Ok.

fmueller:
👍

whiskeysierra:
:+1:

PR #105, opened 2016-07-08
Add link to Spring Boot twintip crawler lib

raphaelbauer:
Just a tiny update after Felix's talk at BNB. The Spring Boot lib makes
it very easy to expose the relevant files for the twintip crawler. We just
accidentally discovered that lib - so we guess it makes more sense
to mention it explicitly.


fmueller:
Thanks for your contribution. This library is listed at http://zalando.github.io/restful-api-guidelines/tooling/Tooling.html#support-libraries

It could be better to link this supporting libraries also directly in Api Discovery section.

tfrauenstein:
👍

fmueller:
👍

PR #107, opened 2016-07-18
Added common pattern must to deprecation.

tkrop:
> After the last API guild meeting, I'm unsure, if we should keep the paragraph like this or change it to the MUST/SHOULD headline style. What do you think? (@codekultur  #91)

How is the opinion of the others?

### DIFF
--- a/deprecation/Deprecation.md
+++ b/deprecation/Deprecation.md
@@ -3,7 +3,7 @@
 Sometimes it is necessary to phase out an API endpoint (or version). I.e. this
 may be necessary if a field is no longer supported in the result or a whole
 business functionality behind an endpoint has to be shut down. There are
-multiple other reasons as well.
+many other reasons as well.
 
 
 ## {{ book.must }} Obtain Approval of Clients
@@ -14,7 +14,7 @@ endpoint. Producers should help consumers to migrate to a potential new
 endpoint (i.e. by providing a migration manual). After all clients are
 migrated, the producer may shut down the deprecated API.
 
-## {{ book.must }} External Partners Must Agree on Lifecycle Timespans
+## {{ book.must }} External Partners Must Agree on Deprecation Timespan
 
 If the API is consumed by any external partner, the producer must define a
 reasonable timespan that the API will be maintained after the producer has
@@ -23,12 +23,12 @@ minimum after-deprecation-lifespan before he starts using the API.
 
 ## {{ book.must }} Reflect Deprecation in API Definition
 
-API deprecation must reflect in the producers OpenAPI definition. If a method
+API deprecation must be part of the OpenAPI definition. If a method
 on a path, a whole path or even a whole API endpoint (multiple paths) should
 be deprecated, the producers must set `deprecated=true` on each method / path
 element that will be deprecated (OpenAPI 2.0 only allows you to define
-deprecation on this level). If deprecation should happen on a finer level
-(i.e. query parameter, payload etc.), the producer should set
+deprecation on this level). If deprecation should happen on a more fine grained 
+level (i.e. query parameter, payload etc.), the producer should set
 `deprecated=true` on the affected method / path element and add further
 explanation to the `description` section. 
 
@@ -53,7 +53,7 @@ header is not sufficient to gain client consent to shut down an API.
 
 ## {{ book.should }} Add Monitoring for Warning Header
 
-Clients should monitor the `Warning` header in HTTP responses.
+Clients should monitor the `Warning` header in HTTP responses to see if an API will be deprecated in future.
 
 ## {{ book.must }} Not Start Using Deprecated APIs
 
--- a/deprecation/Deprecation.md
+++ b/deprecation/Deprecation.md
@@ -38,7 +38,7 @@ section of the API definition.
 
 ## {{ book.should }} Monitor Usage of Deprecated APIs
 
-Producers should monitor usage of deprecated APIs.
+Producers should monitor usage of deprecated APIs until it the API can be shut down.
 
 ## {{ book.should }} Add a Warning Header to Responses
 
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ serve: | $(gitbook)
 	$(gitbook) serve
 
 build: | $(gitbook)
-	$(gitbook) build
+	$(gitbook) build -v 2.6.7
 
 publish: build | $(gh_pages)
 	$(gh_pages) -d _book
--- a/compatibility/Compatibility.md
+++ b/compatibility/Compatibility.md
@@ -143,22 +143,3 @@ are still under review; for example:
 Hint: This versioning scheme differs in the less strict DRAFT aspect from
 [semantic version information](http://semver.org) used for released APIs and
 service applications.
-
-## {{ book.must }} If Neccessary Declare Deprecation
-
-Sometimes it is necessary to phase out an API endpoint (or version). I.e. this may be necessary if a field is no longer supported in the result or a whole business functionality behind an endpoint has to be shut down. There are multiple other reasons as well.
-
-Before shutting down an API (or version of an API) the producer must make sure, that all clients have given their consent to shut down the endpoint. Producers should help consumers to migrate to a potential new endpoint (i.e. by providing a migration manual). After all clients are migrated, the producer may shut down the deprecated API.
-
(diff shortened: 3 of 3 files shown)


deprecation/Deprecation.md
```diff
@@ -0,0 +1,60 @@
+#Deprecation
+
+Sometimes it is necessary to phase out an API endpoint (or version). I.e. this
+may be necessary if a field is no longer supported in the result or a whole
+business functionality behind an endpoint has to be shut down. There are
+multiple other reasons as well.
```
fmueller: I'd rephrase: There are many other reasons as well.

deprecation/Deprecation.md
```diff
@@ -0,0 +1,60 @@
...
+endpoint. Producers should help consumers to migrate to a potential new
+endpoint (i.e. by providing a migration manual). After all clients are
+migrated, the producer may shut down the deprecated API.
+
+## {{ book.must }} External Partners Must Agree on Lifecycle Timespans
+
+If the API is consumed by any external partner, the producer must define a
+reasonable timespan that the API will be maintained after the producer has
+announced deprecation. The external partner (client) must agree to this
+minimum after-deprecation-lifespan before he starts using the API.
```
fmueller: What if external client could not agree with producer on lifecycle span? Announce it afterwards and try to reach consensus - then the last sentence of this rule is somehow confusing.

deprecation/Deprecation.md
```diff
@@ -0,0 +1,60 @@
...
+
+## {{ book.must }} Obtain Approval of Clients
+
+Before shutting down an API (or version of an API) the producer must make
+sure, that all clients have given their consent to shut down the
+endpoint. Producers should help consumers to migrate to a potential new
+endpoint (i.e. by providing a migration manual). After all clients are
+migrated, the producer may shut down the deprecated API.
+
+## {{ book.must }} External Partners Must Agree on Lifecycle Timespans
```
fmueller: ...agree on deprecation timespan

deprecation/Deprecation.md
```diff
@@ -0,0 +1,60 @@
...
+
+If the API is consumed by any external partner, the producer must define a
+reasonable timespan that the API will be maintained after the producer has
+announced deprecation. The external partner (client) must agree to this
+minimum after-deprecation-lifespan before he starts using the API. 
+
+## {{ book.must }} Reflect Deprecation in API Definition
+
+API deprecation must reflect in the producers OpenAPI definition. If a method
```
fmueller: ... must be part of OpenAPI definition

deprecation/Deprecation.md
```diff
@@ -0,0 +1,60 @@
...
+on a path, a whole path or even a whole API endpoint (multiple paths) should
+be deprecated, the producers must set `deprecated=true` on each method / path
+element that will be deprecated (OpenAPI 2.0 only allows you to define
+deprecation on this level). If deprecation should happen on a finer level
```
fmueller: ... should happen on a more fine-grained level

SUMMARY.md
```diff
@@ -16,6 +16,7 @@
 * [Common Data Objects](common-data-objects/CommonDataObjects.md)
 * [Common Headers](headers/CommonHeaders.md)
 * [Proprietary Headers](headers/ProprietaryHeaders.md)
+* [Deprecation](deprecation/Deprecation.md)
```
fmueller: I'd put it under Compatibility as it describes what happens when compatibility can not be guaranteed anymore.

deprecation/Deprecation.md
```diff
@@ -0,0 +1,60 @@
...
+(i.e. query parameter, payload etc.), the producer should set
+`deprecated=true` on the affected method / path element and add further
+explanation to the `description` section. 
+
+If `deprecated` is set to `true`, the producer must describe what clients
+should use instead and when the API will be shut down in the `description`
```
fmueller: What if the deprecated business functionality is not provided anymore in any sense?

deprecation/Deprecation.md
```diff
@@ -0,0 +1,60 @@
...
+section of the API definition. 
+
+## {{ book.should }} Monitor Usage of Deprecated APIs
+
+Producers should monitor usage of deprecated APIs.
```
fmueller: ... and when they fade out.

deprecation/Deprecation.md
codekultur: I'll change it..

deprecation/Deprecation.md
```diff
@@ -0,0 +1,60 @@
...
+field. When adding the `Warning` header, the `warn-code` must be `299` and the
+`warn-text` should be in form of *"The path/operation/parameter/... {name} is
+deprecated and will be removed by {date}. Please see {link} for details."*
+with a link to a documentation describing why the API is no longer supported
+in the current form and what clients should do about it. Adding the `Warning`
+header is not sufficient to gain client consent to shut down an API.
+
+## {{ book.should }} Add Monitoring for Warning Header
+
+Clients should monitor the `Warning` header in HTTP responses.
```
fmueller: ... to see if an API will be deprecated in future.

deprecation/Deprecation.md
codekultur: The trick is that the external partner has to agree with the producer before he is allowed to use the API (i.e. by terms & conditions), just like you do when you use the Salesforce, eBay, AWS API...

deprecation/Deprecation.md
codekultur: sounds good. I'll change it

deprecation/Deprecation.md
codekultur: I'll change it

deprecation/Deprecation.md
fmueller: Right but when we add this chapter to the guidelines there are already APIs for external partners available without negotiate this in Terms&Conditions. For new APIs this makes totally sense.

SUMMARY.md
codekultur: We discussed this option as well.. We put it in the end, as it reflects the API lifespan, but I agree it would also fit under Compatibility...

fmueller: Ok, then we leave it there.

deprecation/Deprecation.md
codekultur: I could add "if there is an alternative", but this would make the sentence more complex and I think it's clear to everyone, isn't it?

deprecation/Deprecation.md
codekultur: I'll add this

deprecation/Deprecation.md
codekultur: I'll change it

deprecation/Deprecation.md
codekultur: Sorry... Don't understand what you mean... 

"Producers should monitor usage of deprecated APIs and when they fade out." ?!?

deprecation/Deprecation.md
codekultur: Hmm... Maybe this is a general discussion: Our guidelines evolve over time. Should we really always consider how the guidelines once looked like and provide migration scenarios, or should we keep the guidelines consistent at one point in time?
I would vote for the second option, as it keeps the guidelines short and easy to understand. But of course it won't give guidance for what to do with existing APIs. I would say: Try to get them compliant to the guidelines, as soon as possible. In this case, there weren't any rules for deprecation before, so the external clients have no guarantee at all so far...

SUMMARY.md
tkrop: I originally voted for compatibility, too. But my main idea for setting it below, was making it less prominent.

deprecation/Deprecation.md
tkrop: I think, than it is a legal problem too. If external consumers have not agreed upfront, we may be in trouble, because we may be forced to produce infinitely on this contract, except there is a possibility for canceling it.

deprecation/Deprecation.md
tkrop: May be we should add 'until it can be shut down.'

fmueller: I mean monitoring the timespan til the endpoint/api is shut down.

deprecation/Deprecation.md
fmueller: Ok

SUMMARY.md
fmueller: Right, making it less prominent is also a good idea as this should be an edge case.

deprecation/Deprecation.md
fmueller: Both of you are right. Let's leave it as it is.

deprecation/Deprecation.md
codekultur: OK. Changed the sentence...

fmueller:
👍

tkrop:
:+1:

PR #110, opened 2016-07-21
fix example to use plural in URL path

hjacobs:
The example was using singular resource names, make it plural to be consistent with our own guidelines :smile:


whiskeysierra:
+1

hjacobs:
:+1:

PR #121, opened 2016-08-18
Dropped support for URI-style link relations

whiskeysierra:
We used to promote custom URI-style link relations before, but I think it makes sense to drop this. The result was rather chaotic since we don't have a canonical domain for this. I've seen custom wikis/pages etc. being used for this. A clean, readable name, similar to the standard link relations is preferable in my opinion.


tfrauenstein:
👍

fmueller:
👍

ePaul:
:+1: (even if it is already merged).

PR #122, opened 2016-08-18
Dropped mentions of 422 Unprocessable Entity

### LINKED ISSUE #82: Which status code to use for semantic error on GET parameters
On [Client Side Errors](http://zalando.github.io/restful-api-guidelines/http/Http.html#client-side-error-codes) you define that `422 Unprocessable Entity` should be use for *semantic errors*. Nevertheless you are limiting to `POST`, `PUT`, `DELETE`, `PATCH` *HTTP verbs*.

I understand the limitation because *request entity* is about *Headers* + *body* thus there is no sense for `GET` *HTTP verb*.

However `GET` can be requested with parameters (here *query string parameter*) that we could validate like:

```
/feed?url=XXX
```

Where `url=XXX` should be respect following validation:

1. Present, if not `400 Bad Request` because is not semantic error
2. To be a valid URL (so not empty) (`@URL`)
3. To be a valid Feed URL

For *2.* I really think is semantic error if user does not pass correct URL. Moreover there is also a edge case for *post-validation* (it may out-scope, *2.* is the main goal of this issue), because it is not possible to determine validity without fetching it.

To conclude which status code should I return for *2.* (and if you have advice for *3.*)?

- `422 Unprocessable Entity` seems to not appropriate because of *Unprocessable Entity*. There is no *entity* on `GET` request
- `400 Bad Request` seems to be the least bad?
- `404 Not Found` can be a great alternative and most *RFC respectful* but can deceive user.

--- 

Upstream discussion: https://github.com/zalando/problem-spring-web/issues/49

whiskeysierra:
This fixes #82

### DIFF
--- a/http/Http.md
+++ b/http/Http.md
@@ -123,7 +123,6 @@ different HTTP methods on resources.
 | 409 | Conflict - returned if, e.g. when two clients try to create the same resource or if there are concurrent, conflicting updates | PUT, DELETE, PATCH |
 | 412 | Precondition Failed - returned for conditional requests, e.g. If-Match if the condition failed. Used for optimistic locking. | PUT, DELETE, PATCH |
 | 415 | Unsupported Media Type - e.g. clients sends request body without content type | PUT, DELETE, PATCH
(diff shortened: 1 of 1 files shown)


ePaul:
:+1:

fmueller:
👍

Issue #117, opened 2016-08-08
Prefer objects for results/lists


dehora:
With Open API, there's two basic ways to return a list of items

- arrays
- objects

Both are common use for APIs defined against these guidelines in Zalando.

I think it's worth recommending against array based responses and preferring object results. The main reason is that arrays design out extensibility, which is counter to the general principles. There's no way to extend an array with extra information or provide a home for common structures like pagination. This seems to lead to the definition of X-Headers to carry information (such as counts, but any kind of metadata really). Since Link headers have been dropped, APIs based on arrays have no real options for paging. Finally, there are odd bugs that can crop up when empty ends up meaning no more data in array responses, but also no data in the the current range, that are less likely to happen with explicit next/previous page declarations supportable by objects.

A knock-on consequence could be a need to firm up pagination models, and possibly defining an idiomatic structure for collections.

fmueller:
We already mention this implicitly in our pagination chapter.

+1 for an explicit recommendation in our guidelines. It's also kinda mandatory when we think about our favoring of backward compatible changes in contrast to versioning: only with surrounding object one can add new fields.

codekultur:
Created a PR for this: https://github.com/zalando/restful-api-guidelines/pull/126

PR #126, opened 2016-08-23
added rule to always return objects, no arrays

codekultur:
https://github.com/zalando/restful-api-guidelines/issues/117

I was not sure where to put it. Feedback welcome..

### DIFF
--- a/compatibility/Compatibility.md
+++ b/compatibility/Compatibility.md
@@ -49,7 +49,7 @@ How to do this:
 * Be prepared to handle HTTP status codes not explicitly specified in endpoint definitions! Note also, that status codes are extensible -- default handling is how you would treat the corresponding x00 code (see [RFC7231  Section 6](https://tools.ietf.org/html/rfc7231#section-6))
 
 ## {{book.must}} Always Return JSON Objects As Top-Level Data Structures To Support Extensibility
-In a response body, you must always return a JSON objects (and not i.e. arrays) as a top level data structure to support future extensibility. JSON objects support compatible extension by additional attributes. This allows you to easily extend your response and i.e. add pagination later, without breaking the API.
+In a response body, you must always return a JSON objects (and not i.e. an array) as a top level data structure to support future extensibility. JSON objects support compatible extension by additional attributes. This allows you to easily extend your response and i.e. add pagination later, without breaking the API.
 
 ## {{ book.should }} Avoid Versioning
 
--- a/compatibility/Compatibility.md
+++ b/compatibility/Compatibility.md
@@ -48,6 +48,9 @@ How to do this:
 * Follow the redirect when the server returns an “HTTP 301 Moved Permanently” response code
 * Be prepared to handle HTTP status codes not explicitly specified in endpoint definitions! Note also, that status codes are extensible -- default handling is how you would treat the corresponding x00 code (see [RFC7231  Section 6](https://tools.ietf.org/html/rfc7231#section-6))
 
+## {{book.must}} Always Return JSON Objects As Top-Level Data Structures To Support Extensibility
+In a response body, you must always return a JSON objects (and not i.e. arrays) as a top level data structure to support future extensibility. JSON objects support compatible extension by additional attributes. This allows you to easily extend your response and i.e. add pagination later, without breaking the API.
+
 ## {{ book.should }} Avoid Versioning
 
 When changing your RESTful APIs,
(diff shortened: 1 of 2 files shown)


dehora:
:+1: 

We can follow up with guidance/details on collection structures another time.

hjacobs:
Do we really want to have a MUST rule for this? There might IMHO be good reasons to return simple arrays without paging. The pagination page (http://zalando.github.io/restful-api-guidelines/pagination/Pagination.html) is currently to vague to enforce objects (header links like GitHub uses seem not to be forbidden).

hjacobs:
I would consider this simple API to be OKish and I would hate to see rules enforced which make simple things more complicated than necessary: https://registry.opensource.zalan.do/ui/#!/Pier_One_API

codekultur:
@hjacobs you are right: The guidelines still allow to add pagination via header fields. I think in the last Guild meeting we said it should be a MUST, but I see your point. Let's decide in the next guild meeting and collect arguments here..
I can live with a SHOULD and an explanation that pagination within the body cannot be added otherwise without breaking the API. So you would be stuck with header-based pagination.

tfrauenstein:
1) Editorial: pls. change headline "...To Support Future Pagination" --> "...To Support Extensibility"
And make clear: objects support compatible extension by additional attributes; arrays have to be avoided as top level return structure because it makes API not extensible on top level; for instance, later introduction of pagination as standardised by rule below will not be possible as compatible extension. 

2) We should update Guidelines to make clear that there is one pagination standard different from pagination via header fields which is discarded.

3) I vote for MUST Guideline -- IMO it does not really add "complexity" and is easy feasible also for simple APIs. But we can discuss this in next Guild meeting.

dehora:
> Do we really want to have a MUST rule for this? 

@hjacobs I think we do because there's no value in having multiple ways to do this. It also means people don't get caught out because they underestimated the result size, or have to hack in headers after the fact because they need to return more than the array results.

pagination/Pagination.md
```diff
@@ -1,5 +1,8 @@
 # Pagination
 
+## {{book.must}} Always Return JSON Objects To Support Extensibility
+When returning a collection, you must always return a JSON object (and not i.e. an array). This holds true for all collections. JSON objects support compatible extension by additional attributes. This way you can easily add pagination later, without breaking the API.
```
ePaul: With some bad will one can understand this as "arrays are forbidden, I have to format my collection as an object (by introducing some key for each item)" (i.e. use a map instead of a list). We should make it clear that lists on non-top-level are still okay.

> Even when returning a collection, the top-level JSON value in the result must be a JSON object, not an array. The actual array would usually be a property of this JSON object. JSON objects support compatible extension by additional attributes. This way you can easily add pagination later, without breaking the API, as well as other metadata properties.

ePaul:
Possibly a better location would be the compatibility section (because this is not just for pagination, but also for other future extensions), or the JSON section (because maybe then people don't read it just when they need to introduce pagination or incompatible changes).

ePaul:
I'm not sure that having an object as the root item is enough to introduce pagination later on without breaking clients ... or rather, you can only introduce optional pagination, not pagination enforced by the server. (If you introduce a limit parameter with a default value of less than the actual number of items available, old clients will simply receive just the first page of the results and don't note that they didn't get all the data. If you introduce a required limit parameter, old clients will be broken visibly. Both is bad.)

Not sure if this needs to be mentioned somewhere.

codekultur:
@ePaul thanks! I moved it to the Compatibility section and rephrased it. Pagination now is only an example. And true, it depends on how you implement pagination, when no request parameters are given, if the API stays backwards compatible.

compatibility/Compatibility.md
```diff
@@ -48,6 +48,9 @@ How to do this:
 * Follow the redirect when the server returns an “HTTP 301 Moved Permanently” response code
 * Be prepared to handle HTTP status codes not explicitly specified in endpoint definitions! Note also, that status codes are extensible -- default handling is how you would treat the corresponding x00 code (see [RFC7231  Section 6](https://tools.ietf.org/html/rfc7231#section-6))
 
+## {{book.must}} Always Return JSON Objects As Top-Level Data Structures To Support Extensibility
+In a response body, you must always return a JSON objects (and not i.e. an array) as a top level data structure to support future extensibility. JSON objects support compatible extension by additional attributes. This allows you to easily extend your response and i.e. add pagination later, without breaking backwards compatibility.
```
ePaul: `i.e.` → `e.g.`?

whiskeysierra: Obligatory link to [theoatmeal.com](http://theoatmeal.com/comics/ie)

compatibility/Compatibility.md
```diff
@@ -48,6 +48,9 @@ How to do this:
 * Follow the redirect when the server returns an “HTTP 301 Moved Permanently” response code
 * Be prepared to handle HTTP status codes not explicitly specified in endpoint definitions! Note also, that status codes are extensible -- default handling is how you would treat the corresponding x00 code (see [RFC7231  Section 6](https://tools.ietf.org/html/rfc7231#section-6))
 
+## {{book.must}} Always Return JSON Objects As Top-Level Data Structures To Support Extensibility
```
whiskeysierra: There is a second one.

ePaul:
:+1:

whiskeysierra:
:+1:

PR #131, opened 2016-08-30
Reformulated http methods to better expose differences between put and post

### LINKED ISSUE #88: Explain difference between POST and PUT for resource creation
Several times during review I noted that API designers often think "Create = POST, Update = PUT".

The actual meaning is more like this:

* PUT means »*please put the enclosed representation at the resource mentioned by the URL*«, which will either create that resource (if it didn't exist before) or replace the resource currently there (if there was one). This should be used for creation when the client decides the URI where the resource will live.
   It has the additional benefit that it is idempotent, i.e. you can send the request twice without having two resources created (because the second one will simply update), and thus are resilient against connection failures.

* POST means (beyond other meanings like "fallback if no other method fits") »*please add the enclosed representation as a new element of the collection resource identified by the URL*«. The server will "invent" a new URI and put the new resource there. This should be used for creation when the server is deciding the resource identifier.

We should adapt the wording in the [HTTP](http://zalando.github.io/restful-api-guidelines/http/Http.html) section to make this clearer.

tkrop:
Fixes #88.

In this first approach, I have extended the description of the different aspects of PUT&POST, but although the other methods without directly referring to other methods. Please check whether this approach is sufficient. Please have a look on my Easter eggs, too ;).


whiskeysierra:
:+1:

http/Http.md
```diff
@@ -6,65 +6,115 @@ Be compliant with the standardized HTTP method semantics summarized as follows:
...
+- PUT requests are usually robust against non-existence of resources by implicitly creating before
+  updating
+- on successful PUT requests, the server will replace the entire resource addressed by the URL with
+  the representation passed in the payload
+- successful PUT requests will usually generate 200, if the resources is updated, 201, if the
+  resource is created, or 204, if no content is returned
+
+**Note:** Resource IDs for PUT are maintained by the client and passed as a URL path segment.
+Putting the same resource twice is required to be idempotent and to result in the same single
+resource instance. If PUT is applied for creating a resource, only URIs should be allowed as
```
jhorstmann: The previous sentence says, resource ids are passed as path segments, but here resource ids themselves are URIs, isn't this a contradiction?

Is the term "Resource ID" defined somewhere? The only mention I found in the guidelines was about security scopes.

http/Http.md
```diff
@@ -6,65 +6,115 @@ Be compliant with the standardized HTTP method semantics summarized as follows:
...
+  would imply deleting the entire collection
+- successful DELETE request will usually generate 200, if the deleted resource is returned, or 204,
+  if no content is returned
+- failed DELETE request will usually generate 404, if the resource cannot be found, or 410, if the
+  resource does not exist
 
 ### HEAD
 
-- has exactly the same semantics as GET, but returns headers only, no body
+HEAD request are used to header information of single resources and resource collections.
```
jhorstmann: Missed a word, how about

> HEAD request are used to *retrieve* header information

http/Http.md
```diff
@@ -6,65 +6,115 @@ Be compliant with the standardized HTTP method semantics summarized as follows:
...
 ### GET
 
-- reads a resource or set of resource instances, respectively
-- individual resources will usually generate a 404 if the resource does not exist;
-  collection resources may return either 200 or 404 if the listing is empty
-- must NOT have request body payload
+GET requests are used to read a single resource or query set of resources.
+
+- GET requests for individual resources will usually generate a 404 if the resource does not exist
+- GET requests for collection resources may return either 200 or 404 if the listing is empty
```
ePaul: Maybe better:

> GET requests for collection resources may return either 200 (with an empty list in the result object) or 404 (with an error object) if the listing is empty.

http/Http.md
```diff
@@ -6,65 +6,115 @@ Be compliant with the standardized HTTP method semantics summarized as follows:
...
+- GET requests must NOT have request body payload
+
+**Note:** GET requests on collection resources should provide a sufficient filter mechanism as well
+as pagination.
```
ePaul: Add a link to the pagination section here?

http/Http.md
```diff
@@ -6,65 +6,115 @@ Be compliant with the standardized HTTP method semantics summarized as follows:
...
+the URL*«.
+
+- PUT requests are usually applied to single resources, and not to collection resources, as this
+  would imply replacing the entire collection
+- PUT requests are usually robust against non-existence of resources by implicitly creating before
+  updating
+- on successful PUT requests, the server will replace the entire resource addressed by the URL with
+  the representation passed in the payload
+- successful PUT requests will usually generate 200, if the resources is updated, 201, if the
+  resource is created, or 204, if no content is returned
```
ePaul: The creation case is 201, regardless of whether content is returned or not. This is not clear from the current wording.

> - successful PUT requests will usually generate 200 (if some content is returned) or 204 (if no content is returned) if the resources is updated, or 201, if the resource is created.

http/Http.md
ePaul: The sentence "only URIs should be allowed as resource IDs" sounds strange, yes. What is this supposed to mean?

http/Http.md
```diff
@@ -6,65 +6,115 @@ Be compliant with the standardized HTTP method semantics summarized as follows:
...
+
+### POST
+
+POST requests are idiomatically used to create single resources, but other semantics on
+collection resources are equally possible. The semantic is best described as »*please add the
+enclosed representation or execute a well specified request to/on the collection resource
+identified by the URL*«.
+
+- POST request should only be applied to collection resources, and normally not on single resource,
+  as this has an undefined semantic
```
ePaul: Hmm, no.

* For collection resources, the (idiomatic) semantic is »*please add the enclosed representation to the collection resource identified by the URI*«.
* For non-collection resources, there is no predefined semantic, but one can given by the API definition, and then we have the »*execute a well specified request on the resource identified by the URI*« meaning. (This is the "more generally" case listed down there.)

http/Http.md
```diff
@@ -6,65 +6,115 @@ Be compliant with the standardized HTTP method semantics summarized as follows:
...
+- successful POST requests will usually generate 200, if a resources have been updated, 201 if the
+  resources have been created, and 202, if the request was accepted but has not been finished
+
+**More generally:** POST should be used for scenarios that cannot be covered by the other methods
+sufficiently. For instance, GET with complex (e.g. SQL like structured) query that needs to be
+passed as request body payload because of the URL-length constraint. In such cases, make sure to
+document the fact that POST is used as a workaround.
+
+**Note:** Resource IDs for POST are created and maintained by server and returned with response
+payload. Posting the same resource twice is by itself **not** required to be idempotent and may
```
ePaul: "Resource IDs for resources created by POST are ..."

http/Http.md
```diff
@@ -6,65 +6,115 @@ Be compliant with the standardized HTTP method semantics summarized as follows:
...
+result in multiple resource instances. Anyhow, in the presence of external URIs it is best practice
+to implement POST in an idempotent way.
```
ePaul: > Anyhow, in the presence of external URIs it is best practice to implement POST in an idempotent way.

I don't understand what is meant by this.

http/Http.md
tkrop: This is the first hint on the generalized ZRN (=> URIs) debate.

http/Http.md
```diff
@@ -88,6 +88,11 @@ only one of the following patterns per endpoint, unless forced by a [backwards c
 4. use POST (with a proper description of what is happening) instead of PATCH if the request does
    not modify the resource in a way defined by the semantics of the media type.
 
+In practice [JSON Merge Patch](https://tools.ietf.org/html/rfc7396) quickly turns out to to
+limited, especially when trying to update single objects in large collections. In this cases
+[JSON Patch](http://tools.ietf.org/html/rfc6902) can shown its full power while still showing
+readable patch requests ([see also](http://erosb.github.io/post/json-patch-vs-merge-patch)).
```
ePaul: Please don't use "see also" as a link title. Either just attach the link to some part of the sentence, or use the document's title, like »(see also [JSON Patch and JSON Merge Patch](http://erosb.github.io/post/json-patch-vs-merge-patch)).«

http/Http.md
```diff
@@ -6,65 +6,120 @@ Be compliant with the standardized HTTP method semantics summarized as follows:
...
-       (This is basically  [JSON Merge Patch](https://tools.ietf.org/html/rfc7396), a specialized media type
-       `application/merge-patch+json` that is a partial resource representation.)
-     3. use PATCH with [JSON Patch](http://tools.ietf.org/html/rfc6902), a specialized media type
-       `application/json-patch+json` that includes instructions on how to change the resource.
-     4. use POST (with a proper description of what is happening) instead of PATCH if the request does
-        not modify the resource in a way defined by the semantics of the media type.
+PATCH request are only used for partial update of single resources, i.e. where only a specific
+subset of resource fields should be replaced. The semantic is best described as »*please change
+the resource identified by the URL according to my change request*«. The semantic of the change
+request is undefined and must be described in the API specification.
```
ePaul: "... request is not defined by the HTTP standard and must be described in the API specification, using suitable media types".

http/Http.md
tkrop: Personally, I would like to drop the 204 on PUT, because I think it is bad practice to not indicate whether the resource was created (201) or updated (200) - independent whether we provide content. What do others think about that?

http/Http.md
```diff
@@ -6,65 +6,120 @@ Be compliant with the standardized HTTP method semantics summarized as follows:
...
+   at all).
+2. use PATCH with partial objects to only update parts of a resource, when ever possible. (This is
+   basically  [JSON Merge Patch](https://tools.ietf.org/html/rfc7396), a specialized media type
+   `application/merge-patch+json` that is a partial resource representation.)
+3. use PATCH with [JSON Patch](http://tools.ietf.org/html/rfc6902), a specialized media type
+   `application/json-patch+json` that includes instructions on how to change the resource.
+4. use POST (with a proper description of what is happening) instead of PATCH if the request does
+   not modify the resource in a way defined by the semantics of the media type.
+
+In practice [JSON Merge Patch](https://tools.ietf.org/html/rfc7396) quickly turns out to to
```
ePaul: "... turns out to be too ..."

http/Http.md
whiskeysierra: > What do others think about that?

I agree.

http/Http.md
tkrop: I would like to let the error object out of the description here.

http/Http.md
```diff
@@ -6,65 +6,120 @@ Be compliant with the standardized HTTP method semantics summarized as follows:
...
+limited, especially when trying to update single objects in large collections. In this cases
```
ePaul: > In practice [JSON Merge Patch](https://tools.ietf.org/html/rfc7396) quickly turns out to **be too** limited, especially when trying to update single objects in large collections **(as part of a single resource)**.

(Highlights for marking changes, not to be included.)

http/Http.md
tkrop: Is this a suggestion for change?

http/Http.md
```diff
@@ -6,65 +6,120 @@ Be compliant with the standardized HTTP method semantics summarized as follows:
...
-- more generally, POST should be used for scenarios that cannot be covered by the other methods.
-  For instance, GET with complex (e.g. SQL like structured) query that needs to be passed as
-  request body payload. In such cases, make sure to document the fact that POST is used as a
-  workaround
+- DELETE requests are usually applied to single resources, not on collection resources, as this
+  would imply deleting the entire collection
+- successful DELETE request will usually generate 200, if the deleted resource is returned, or 204,
+  if no content is returned
+- failed DELETE request will usually generate 404, if the resource cannot be found, or 410, if the
+  resource does not exist
```
ePaul: Hmm, where is the difference between "does not exist" and "cannot be found"?

I would say "410 if the server knows that the resource was already deleted before".

http/Http.md
tkrop: Isn't that redundant?

http/Http.md
tkrop: How about: Anyhow, if external URIs are present that can be used to identity duplicate requests, it is best practice to implement POST in an idempotent way.

http/Http.md
ePaul: Hmm, maybe. The point is that a 200 response would have the usual API result for a successful retrieval, the 404 would not.

(I would return 404 just when the collection resource does not exist, not when it is empty. Though in some APIs it might be difficult to distinguish both cases.)

http/Http.md
ePaul: I think 200 without content is not permitted by the HTTP standard, that needs to be 204. (But 201 without content is okay.) Therefore my reordering of the sentence.

http/Http.md
ePaul: Not yet worded as such, I'll have to think about how to do this (because it affects more than just this sentence). But feel free to implement it yourself.

http/Http.md
tkrop: Yes, excellent distinction. I already got this from your first comment :)

http/Http.md
ePaul: Hmm, for "Resource IDs for POST" I can think of the target URI. Of course, this makes no sense when you read the rest of the sentence. The point is that POST doesn't always create resources (just when adding to collections), and in the case where it doesn't there are no resource IDs.

http/Http.md
tkrop: The standard says:

> Aside from responses to CONNECT, a 200 response always has a payload,
> though an origin server MAY generate a payload body of zero length.
> If no payload is desired, an origin server ought to send 204 (No
> Content) instead.

I think, it is still an open discussion to send 200 with empty body, which is explicitly allowed.

http/Http.md
tkrop: Okay, I give it a try to tore this two cases apart.

http/Http.md
tkrop: Okay, will get longer, but you are right.

http/Http.md
tkrop: Yes sorry, I assumed this implicitly.

tkrop:
@ePaul @jhorstmann thanks for the helpful reviews. I have tried to incorporate all of your suggestions.

http/Http.md
```diff
@@ -27,52 +28,55 @@ the URL*«.
   updating
 - on successful PUT requests, the server will replace the entire resource addressed by the URL with
   the representation passed in the payload
-- successful PUT requests will usually generate 200, if the resources is updated, 201, if the
-  resource is created, or 204, if no content is returned
+- successful PUT requests will usually generate 200 or 204 (if the resources was updated - with or
```
ePaul: resources → resource

http/Http.md
```diff
@@ -27,52 +28,55 @@ the URL*«.
...
+  without actual content returned), and 201 (if the resource was created)
 
-**Note:** Resource IDs for PUT are maintained by the client and passed as a URL path segment.
-Putting the same resource twice is required to be idempotent and to result in the same single
-resource instance. If PUT is applied for creating a resource, only URIs should be allowed as
-resource IDs. If URIs are not available POST should be preferred.
+**Note:** Resource IDs with respect PUT requests are maintained by the client and passed as a URL
```
ePaul: "... with respect to PUT ..."

http/Http.md
```diff
@@ -27,52 +28,55 @@ the URL*«.
...
 
-**Note:** Resource IDs for POST are created and maintained by server and returned with response
-payload. Posting the same resource twice is by itself **not** required to be idempotent and may
-result in multiple resource instances. Anyhow, in the presence of external URIs it is best practice
-to implement POST in an idempotent way.
+**Note:** Resource IDs with respect to POST requests are created and maintained by server and
+returned with response payload. Posting the same resource twice is by itself **not** required to
+be idempotent and may result in multiple resource instances. Anyhow, if external URIs are present
+that can be used to identity duplicate requests, it is best practice to implement POST in an
+idempotent way.
```
ePaul: identity → identify

ePaul:
:+1:

tkrop:
Anybody else that wants to lift his thumb?

whiskeysierra:
:+1:

PR #133, opened 2016-08-30
Added recommandation against enums

### LINKED ISSUE #119: Recommend against enums
We cover Open API enums in a few places, notably here: http://zalando.github.io/restful-api-guidelines/compatibility/Compatibility.html

> - Enum ranges cannot not be extended when used for output parameters — clients may not be prepared to handle it. However, enum ranges can be extended when used for input parameters.

> - Enum ranges can be reduced when used as input parameters, only if the server is ready to accept and handle old range values too. Enum values can be reduced when used as output parameters.

> - Use x-extensible-enum, if range is used for output parameters and likely to be extended with growing functionality. It defines an open list of explicit values and clients must be agnostic to new values (e.g. via casuistic with default behavior).

This is great and is accurate, but I see enums consistently in APIs based on the guidelines that are not closed sets. I think we need to recommend against using `enum` as top level guidance and not just describe the pitfalls in a general section. The `x-extensible-enum` is useful as a workaround, but the concept of an enum is closed by design, also it can't be used in JSON Schema which is necessary for events. Maybe we need a more direct way to describe an open-ended list of values.

tkrop:
Fixes #119

### DIFF
--- a/compatibility/Compatibility.md
+++ b/compatibility/Compatibility.md
@@ -29,7 +29,8 @@ Apply the following rules to evolve RESTful APIs in a backward-compatible way:
 * Enum ranges can be reduced when used as input parameters, only if the server
   is ready to accept and handle old range values too. Enum values can be reduced
   when used as output parameters.
-* Use x-extensible-enum, if range is used for output parameters and likely to
+* Use [`x-extensible-enum`](#should-used-openended-list-of-values-xextensibleenum-instead-of-enumerations),
+  if range is used for output parameters and likely to
   be extended with growing functionality. It defines an open list of explicit
   values and clients must be agnostic to new values (e.g. via casuistic with
   default behavior).
@@ -43,7 +44,8 @@ How to do this:
 
 * Ignore new and unknown fields in the payload (see also Fowler’s
   “[TolerantReader](http://martinfowler.com/bliki/TolerantReader.html)” post)
-* Be prepared for new enum values declared with x-extensible-enum (see above);
+* Be prepared for new enum values declared with
+  [`x-extensible-enum`](#should-used-openended-list-of-values-xextensibleenum-instead-of-enumerations);
   provide default behavior for unknown values, if applicable
 * Follow the redirect when the server returns an “HTTP 301 Moved Permanently” response code
 * Be prepared to handle HTTP status codes not explicitly specified in endpoint definitions! Note also, that status codes are extensible -- default handling is how you would treat the corresponding x00 code (see [RFC7231  Section 6](https://tools.ietf.org/html/rfc7231#section-6))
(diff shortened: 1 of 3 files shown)


ePaul:
Maybe adapt the compatibility section which mentions this, too.

tkrop:
Added links and improved wording.

ePaul:
During the meeting we discussed whether the name `x-extensible-enum` for this property is really the right thing (no decision was yet done about this).

I originally proposed this name for the extension property in the guidelines, and took the syntax from Swagger's `enum` property (which came from JSON schema). In JSON schema (and thus also in Swagger/OpenAPI) `enum` restricts the values to just the given values – which is useful for validation, and in code-generating tools is also used for generating enums in programming languages.

Some arguments for/against changing:
* "Extensible enum" is an oxymoron, and thus a bad name.
* What we have here is basically a list of examples (for which possibly some meaning is already defined – though not documented), not a list of *all* possible examples (which would be an enumeration).
* We already established this name, and it is used in several API definitions already. Changing it now would confuse people.
* There is no (known) software support for this property name, changing it would not break any API (nor implementation).

@all What would be better names? Just `x-examples`? (Other proposals welcome.)

If we are already changing the name of the property, we could consider to add a description for each item. [The next version of OpenAPI likely will have be some way of documenting each enum item](https://github.com/OAI/OpenAPI-Specification/issues/348), maybe we want to do this similarly? (Though that is not yet finished.)

whiskeysierra:
> @all What would be better names? Just x-examples? (Other proposals welcome.)

`x-samples`, pun intended

whiskeysierra:
+1 on using an object for this to add descriptions per value.

tkrop:
> +1 on using an object for this to add descriptions per value.

@whiskeysierra hmm, can you give an example, what you have in mind?

ePaul:
@tkrop 
One possible way (which would work not just for strings, also for whole objects or similar):
```yaml
  Event:
    description:
      An event which will be delivered to Nakadi (or other event sinks),
      together with some metadata used by Tarbela.
    type: object
    properties:
      [...]
      delivery_status:
        type: string
        description: |
          The delivery status of an event.
       x-samples:
        - value: NEW
          description:
             the event was created in the producer, and not yet tried to publish.
        - value: SENT
          description: the event was successfully published to the event sink.
        - value: ERROR
          description: something did go wrong with publishing the event.
```

For enum-like string properties something like this would also work:
```yaml
properties:
  discount_type:
    type: string
    description: the code of the discount type for this discount condition.
       x-samples:
         NEW: the event was created in the producer, and not yet tried to publish.
         SENT: the event was successfully published to the event sink.
         ERROR: something did go wrong with publishing the event.
```

(This could go into [Tarbela's producer API](https://github.com/zalando-incubator/tarbela/blob/master/src/main/resources/api/event-producer-api.yaml#L161)).

tkrop:
@ePaul @whiskeysierra great suggestion. Which one should I pick for example? both?

whiskeysierra:
The first one is more powerful, I'd say. But I don't like to have Java-idomatic enum values (upper underscore). I think snake case or lower hyphen would be a better fit.

tkrop:
@whiskeysierra I think it is not only Java, but many other programming languages too ... event if http://blog.josephwilk.net/rhetorical-programming/why-are-you-shouting-programmer.html isn't a good reference ;)

@all I'm willing to change it to lower snake case if some more reviewers support this request.

ePaul:
The example was from an existing API, which used upper-case values. For an example in the Guidelines of how this should look of course you can write them differently. (I guess we still need some explanation text around this.)

whiskeysierra:
:+1:

fmueller:
👍

Issue #116, opened 2016-08-03
Clarify use of X-Total-Count


dehora:
In ["Could: Use Pagination Links Where Applicable"](http://zalando.github.io/restful-api-guidelines/pagination/Pagination.html#could-use-pagination-links-where-applicable), `X-Total-Count` is mentioned as an option for describing result count.

In JSON (which is the required media type), it's not necessary to ask clients to look in a header when a total count field can be added to the result entity, simplifying client access. The guidelines could be interpreted as saying when you serve pagination links, you are to use `X-Total-Count` in conjunction with paging links. It would be good to edit the example to include a total count in the data, and making `X-Total-Count` entirely optional. A final thought I guess is that we should consider avoiding custom headers to work with APIs unless they're strictly needed. 

----

`X-Total-Count` seems to make most sense in these cases:

- The design style is to use the the Link header
- The response media type isn't able to carry the information
- The JSON response is an array instead of an object

The first two are not directly applicable (Link was removed a while back, we require JSON). The last one is (imho) something we shouldn't encourage because arrays don't afford extensibility the way object based collections/results can (eg you can't easily include pagination links with an array result in the example).

ePaul:
I think you are right here. When we deprecated links in headers, we should have also deprecated this total count in headers.

Of course, when deciding whether to include such a total count at all in your API design, you should consider the performance implications (i.e. your database might need to have a look at all candidate items in order to count them, depending on filter criteria) and whether this information is actually needed.

dehora:
> Of course, when deciding whether to include such a total count at all in your API design,

Absolutely agree; I've seen this a number of times :) Perhaps it could be added as an aside to the guidelines.

ePaul:
During the API Guild meeting we agreed this is a good idea.
@dehora, could you submit a pull request?

PR #134, opened 2016-09-02
Qualify the use of X-Total-Count header.

dehora:
This makes the X-Total-Count header optional and not preferred
over count data in a JSON result, along with guidance on when
the header can be useful.

For #116

### DIFF
--- a/pagination/Pagination.md
+++ b/pagination/Pagination.md
@@ -58,7 +58,6 @@ Further reading:
 
 ## {{ book.could }} Use Pagination Links Where Applicable
 
-* Set `X-Total-Count` to send back the total count of entities.
 * Set links to provide information to the client about subsequent paging options.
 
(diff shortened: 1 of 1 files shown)


whiskeysierra:
:+1:

ePaul:
:+1:

tfrauenstein:
Content is fine. However, ...
a) example still uses X-Total-Count in header and not as JSON attribute
b) attribute would be named, "total_count", I assume

And fine that considering possible performance implication has been mentioned.
Total count often does not provide that much value. I would state it a bit more conservative, and recommend to not support it per default unless it really provides value and can be supported efficiently.

dehora:
updated in bc1aa0c

ePaul:
:+1:

fmueller:
👍

Issue #93, opened 2016-06-20
Keep event guidelines in sync with updates in other sections


fmueller:
At least the enumeration in http://zalando.github.io/restful-api-guidelines/events/events.html#must-event-payloads-must-be-defined-in-accordance-with-the-overall-guidelines is out of sync with our current HATEOAS rules. We need to check this with other rules as well.

fmueller:
@dehora As 'owner' of event section, can you please take a look at this? Maybe simply linking the referenced sections would be enough. Thank you :)

PR #136, opened 2016-09-02
Keep event guidelines in sync with updates in other sections.

dehora:
This references the top level sections instead of subsections to avoid event guidance drifting from other content as it evolves.

Also, add explanation on what entity data to use for a data change event. This clarifies the guidelines for using hash with a data change event by pointing out that the partition key should be composed from the identity of the underlying entity.

For #93


tfrauenstein:
👍

ePaul:
:+1:

PR #140, opened 2016-09-27
Events: event data must conform to data protection and access controls.

dehora:
Adds a guideline to ensure that sensitive data is not published
without first checking it is ok to do so.

### DIFF
--- a/events/events.md
+++ b/events/events.md
@@ -22,6 +22,14 @@ An EventType is registered with Nakadi via its _Schema Registry API_. Once the E
 
 The service specific data defined for an event is called the _payload_. Only this non-Nakadi defined part of the event is expected to be submitted with the EventType schema.  When defining an EventType's payload schema as part of your API and for review purposes, it's ok to declare the payload as an object definition using Open API, but please note that Nakadi currently expects the EventType's schema to be submitted as JSON Schema and not as Open API JSON or YAML. Further details on how to register EventTypes are available in the Nakadi project's documentation.
 
+## {{ book.must }} Events must not provide sensitive customer personal data.
+
++Similar to API permission scopes, there will be Event Type permissions passed via an OAuth token supported in near future by the Nakadi API. Hence, Nakadi will restrict event data access to clients with sufficient authorization. However, teams are asked to note the following:
+
+ - Sensitive data, such as (e-mail addresses, phone numbers, etc) are subject to strict access and data protection controls. 
+ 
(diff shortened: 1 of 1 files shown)


events/events.md
```diff
@@ -22,6 +22,12 @@ An EventType is registered with Nakadi via its _Schema Registry API_. Once the E
 
 The service specific data defined for an event is called the _payload_. Only this non-Nakadi defined part of the event is expected to be submitted with the EventType schema.  When defining an EventType's payload schema as part of your API and for review purposes, it's ok to declare the payload as an object definition using Open API, but please note that Nakadi currently expects the EventType's schema to be submitted as JSON Schema and not as Open API JSON or YAML. Further details on how to register EventTypes are available in the Nakadi project's documentation.
 
+## {{ book.must }} Event data must conform to data protection and access controls.
+
+Sensitive data, such as (e-mail addresses, phone numbers, etc) are subject to strict access and data protection controls. As a result, event type owners **must** ensure data is being published is in accordance with any such criteria. 
+
+Also, please note that the Nakadi API currently has no support in its OAuth scopes for restricting access to event types. It is intended in the future to allow per topic scope controls, but in the meantime be aware that event data in Nakadi is broadly accessible to any consumers with a read scope.
+
```
tfrauenstein: Title: Events must not provide sensible customer personal data.

We should more clearly say: Analog to API access permissions, there will be Event Type access permissions passed via OAuth token supported in near future by Nakadi access authorisation.

This is a temporary rule which will be obsolete as soon as Nakadi supports event type authorization: 

Sensitive personal data, such as (e-mail addresses, phone numbers, etc) are subject to strict access and data protection controls. Event type owners **must** ensure not to publish this sensible customer data.

valgog: +1

fmueller:
This topics looks quite Zalando specific and Nakadai specific. I'm not sure if APi guidelines is the right place for this.

Furthermore, it's problematic to forbid sensible data when Nakadi is used for service-to-service communication. Also I think at the moment it's a tooling problem: we miss access scopes in events, right?

dehora:
>  I'm not sure if APi guidelines is the right place for this.

It is Zalando specific; that's true, but that's not unique in the guidelines. 

In any case, I've updated to match @tfrauenstein's feedback in d4d6481; just moved some sentences around, so imo it's  ready to merge. And fwiw I'm +0 on whether it goes here or somewhere else; my only observation is that this is one place teams are likely to see it, as well as API reviewers who can query event data as needed. 

/ @fmueller @valgog @whiskeysierra

dehora:
@fmueller @valgog @whiskeysierra @tfrauenstein small edits in 8ecbc69

fmueller:
👍

whiskeysierra:
:+1:

Issue #153, opened 2016-11-08
UUIDs versus sequential (or speaking?) IDs


ePaul:
* when to use what?
* configuration data (warehouses, sales channels, ...) vs. mass data (like articles, orders, ...)
* client vs. server side ID generation

whiskeysierra:
**Pros**
- distributed id management becomes easier (but **not necessarily** [simpler](https://www.infoq.com/presentations/Simple-Made-Easy))

**Cons**
- poor readability for developers (log files, etc.) and  readability/usability for end users (in case they are exposed as e.g. invoice numbers)
- length of 36 characters (by default) can at most be reduced to 23 characeters (`uuid -F BIN | base64 | sed 's/=//g'` produces `6FVxOKZhEeaFzKOd3UAmeg`)

ePaul:
The main reason to use an UUID as an identifier for entities is to make sure no one else generates the same ID for a different entity (whether of same or different type).
This is reached by having a vast space of possible values, and using either good (pseudo)randomness or some namespace mechanism (time + MAC address + counter) to select values from that space.

This vast space makes the UUIDs long (128 bits → 16 bytes, 32 in hexadecimal (with standard delimiters 36), 22 chars in base64)). This is not really a problem when storing or transmitting them (storage and bandwith is cheap and becoming even cheaper), but it is when reading them, in logs, for phone calls etc.

For configuration data like business partners, warehouses, sales channels, etc. creations are not that often that we need to ensure concurrent creations – a simple counter would be enough, or possibly a speaking identifier ("code") selected by a human user (like as we are using for warehouses in Wholesale context: `WH-DE-EF` for our Erfurt warehouse, `WH-IT-ST` for the one in Stradella).
When different business organizations are creating the same type of entity, possibly namespace prefixes as discussed in #62 are a solution – while here the combined identifiers might get even longer than UUIDs, they still are much more easily to read, type and compare for humans.

tlossen:
i think that UUID's should only be used if ID's have to be generated in a distributed (= uncoordinated) manner by multiple actors. basically, what they were invented for. in all other cases, costs (lack of readability, index-friendlyness etc.) outweigh their benefits.

or, as [richard clayton](https://rclayton.silvrback.com/do-you-really-need-a-uuid-guid) puts it:
> UUID's should be a last resort tactic,employed only after you have exhausted other strategies for determining uniqueness in an architecture.

fmueller:
[...]

PR #155, opened 2016-11-22
added UUID guideline

tlossen:


fmueller:
references #153

fmueller:
I'd like to add some concerns for incorporating business logic inside IDs like having color codes in product ids. It's good to have readable IDs where API clients quite easily understand what to do with it. It becomes a problem when domain entities or parts of it are hard coded into IDs.

Do you think it make sense to add a paragraph about such reasoning to this rule?

tlossen:
i've reworked the proposed guideline based on our discussion in the guild meeting today.

everything fine?

resources/Resources.md
```diff
@@ -37,6 +37,16 @@ Examples:
     /carts/1681e6b88ec1/items
     /carts/1681e6b88ec1/items/1
 
+
+## {{ book.should }} Only Use UUIDs If Necessary
+
+The main advantage of UUIDs is the fact that they can be generated in a distributed way without risk of conflict, and that they are unique across different resources. Nevertheless, they should only be used if those properties are really needed.
+
+In most situations, using either an integer ID or a (manually assigned) code string is better than using a UUID. They are very long (36 characters), which makes them hard to read, memorize and reason about &mdash; both for internal users, and even more so for end users (in case the ID is ever exposed to them).
```
dehora: > using either an integer ID 

Can we restrict identifiers to strings here? If a service wants to base ids on internal numerics, I'd suggest they do that as an opaque string rather than exporting a number type.

resources/Resources.md
```diff
@@ -37,6 +37,16 @@ Examples:
...
+
+Use of UUIDs is especially discouraged for any kind of reference data with low cardinality.
```
dehora: Can we add the following caveat - `"If used as identifiers UUIDs should not be qualified using a format property, as this limits the server's ability to evolve its identifiers"`.

resources/Resources.md
tlossen: hmmm .... not sure about this one as a general recommendation.

fmueller: Thanks @dehora Good point. Any objections @tlossen?

resources/Resources.md
fmueller: Not sure I understand your point. Can you give an example?

hjacobs:
BTW: did you know [Universally Unique Lexicographically Sortable Identifier (ULID)](https://github.com/alizain/ulid)? This sounds like a nice alternative for use cases where sorting is relevant --- maybe it sounds exotic right now, but I like the idea of having something like UUID but with "some more" information than just randomness. We might at least mention/link it...

dehora:
:+1:

tfrauenstein:
@hjacobs thx. for the hint -- added it to the PR

hjacobs:
@tfrauenstein thanks!

tfrauenstein:
👍

tlossen:
👍

PR #156, opened 2016-12-02
fixed issue 152 - fixed security definitions

codekultur:
fixed example to password flow (as currently used by STUPS services) and added a sentence, that API endpoints should not care about this information.


tfrauenstein:
👍

fmueller:
👍

PR #172, opened 2017-01-10
Clarify and discourage content location header

### LINKED ISSUE #118: Clarify use of Location and Content-Location for PUT/POST methods
HTTP does not make understanding the use Location and Content-Location easy. You have to bounce around RFCs, reason about combinations of response codes and methods, and generally tie a bunch of stuff together. We touch on this in a few places, but it's not clear the guidance is making things easy to understand as it could. For example, http://zalando.github.io/restful-api-guidelines/http/Http.html#success-codes focuses on 201 and says always set the Location header, but doesn't say why or for what reason (201 itself does not require a Location header, and a PUT request that would require a Location header would be odd, as PUT is sent to the target resource). As another example http://zalando.github.io/restful-api-guidelines/headers/CommonHeaders.html says the combination of Location and Content-Location is such that if they are equal this makes "a subsequent GET operation from the client side not necessary": on the other hand, [PUT is pretty clear in saying responses cannot be cached](https://tools.ietf.org/html/rfc7231#section-4.3.4), 

```
Responses to the PUT method are not cacheable.
```

But in the [definition of Content-Location elsewhere](https://tools.ietf.org/html/rfc7231#section-3.1.4.2) it says:

```
For a state-changing request like PUT (Section 4.3.4) or
   POST (Section 4.3.3), it implies that the server's response contains
   the new representation of that resource, thereby distinguishing it
   from representations that might only report about the action (e.g.,
   "It worked!").  This allows authoring applications to update their
   local copies without the need for a subsequent GET request." 
```

It's not clear how to resolve these statements without hairsplitting around client updating state versus clients updated caches.

This is less a criticism of the guidelines, and more that HTTP isn't organised in a way that makes explaining these things obvious to developers because it defines methods, headers and response codes independently with occasional cross-references. I suspect it requires discussing how the headers are to be used for each combination of PUT/POST/201/200. In turn that would suggest we should have a section dedicated to lifecycles (creating, updating and maybe deleting resources) that is optimised for API design, rather than organising the guidelines around HTTP elements.

tkrop:
Fixes #118

### DIFF
--- a/headers/CommonHeaders.md
+++ b/headers/CommonHeaders.md
@@ -11,11 +11,11 @@ they can be used in both, HTTP requests and responses. Commonly used content hea
  - [`Content-Encoding`](https://tools.ietf.org/html/rfc7231#section-3.1.2.2) indicates compression or encryption algorithms applied to the content.
  - [`Content-Length`](https://tools.ietf.org/html/rfc7230#section-3.3.2) indicates the length of the content (in bytes).
  - [`Content-Language`](https://tools.ietf.org/html/rfc7231#section-3.1.3.2) indicates that the body is meant for people literate in some human language(s).
- - [`Content-Location`](https://tools.ietf.org/html/rfc7231#section-3.1.4.2) indicates where the body can be found otherwise ([see below for more details](#must-use-contentlocation-header-correctly)).
+ - [`Content-Location`](https://tools.ietf.org/html/rfc7231#section-3.1.4.2) indicates where the body can be found otherwise ([see below for more details](#could-use-contentlocation-header)).
  - [`Content-Range`](https://tools.ietf.org/html/rfc7233#section-4.2) is used in responses to range requests to indicate which part of the requested resource representation is delivered with the body.
  - [`Content-Type`](https://tools.ietf.org/html/rfc7231#section-3.1.1.5) indicates the media type of the body content.
 
-## {{ book.must }} Use Content-Location Header Correctly
+## {{ book.could }} Use Content-Location Header
 
 The Content-Location header is *optional* and can be used in successful write operations (PUT, POST or PATCH) or read operations (GET, HEAD) to guide caching and signal a receiver the actual location of the resource transmitted in the response body. This allows clients to identify the resource and to update their local copy when receiving a response with this header.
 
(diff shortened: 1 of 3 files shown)


headers/CommonHeaders.md
```diff
@@ -11,17 +11,23 @@ they can be used in both, HTTP requests and responses. Commonly used content hea
...
-In the case of the GET requests it points to a location where an alternate representation of the
-entity in the response body can be found. In this case one
-has to set the Content-Type header as well. For example:
+The Content-Location header can be used to support the following use cases:
+
+- For reading operations GET and HEAD, a different location than the requested URI can be used to indicate that the returned resource is subject to content negotiations, and that the value provides a more specific identifier of the resource.
+- For writing operations PUT and PATCH, an identical location to the requested URI, can be used to explicitly indicate that the returned resource is the current representation of the newly created or updated resource.
+- For writing operations POST and DELETE, a content location can be used to indicate that the body contains a status report resource in response to the requested action, which is available at provided location.
+
+The standard application of the last use case is the creation of resource via POST, where the Content-Location is used to provide the location and identity of the created resource.
```
whiskeysierra: That's only true if the representation is part of the response. Otherwise `Location` is the correct header here.

tkrop: @whiskeysierra This is just very fuzzy an example where it could be used ...

fmueller: @tkrop I would remove the last example as it may adds more confusion than it clarifies this topic for us.

headers/CommonHeaders.md
```diff
@@ -11,17 +11,23 @@ they can be used in both, HTTP requests and responses. Commonly used content hea
...
  - [`Content-Length`](https://tools.ietf.org/html/rfc7230#section-3.3.2) indicates the length of the content (in bytes).
  - [`Content-Language`](https://tools.ietf.org/html/rfc7231#section-3.1.3.2) indicates that the body is meant for people literate in some human language(s).
- - [`Content-Location`](https://tools.ietf.org/html/rfc7231#section-3.1.4.2) indicates where the body can be found otherwise ([see below for more details](#must-use-contentlocation-correctly)).
+ - [`Content-Location`](https://tools.ietf.org/html/rfc7231#section-3.1.4.2) indicates where the body can be found otherwise ([see below for more details](#must-use-contentlocation-header-correctly)).
  - [`Content-Range`](https://tools.ietf.org/html/rfc7233#section-4.2) is used in responses to range requests to indicate which part of the requested resource representation is delivered with the body.
  - [`Content-Type`](https://tools.ietf.org/html/rfc7231#section-3.1.1.5) indicates the media type of the body content.
 
-## {{ book.must }} Use Content-Location Correctly
+## {{ book.must }} Use Content-Location Header Correctly
```
fmueller: Can we change this to a COULD rule? And rephrase title to "Use Content-Location header"

tkrop: Okay. sounds reasonable. I noticed, that I also should change the next rule to should.

fmueller:
👍

fmueller:
@whiskeysierra Is it also ok for you now?

fmueller:
@ePaul Could you have a look and give your ok if you are fine with it? I'd like to finish this PR finally. :)

ePaul:
:+1:

PR #173, opened 2017-01-11
clarification on which common field name type is referenced

hbelen:
This wording makes it slightly more clear that a field called `type` should have a type `string`. Initially one might think that every type of every property should be a `string`.


hjacobs:
:+1:

tfrauenstein:
👍

PR #178, opened 2017-01-17
#170 Events should be design for commutativity

### LINKED ISSUE #170: Recommend use of commutative events
Designing a resilient and fail safe event based architecture is hard when you are relying on implicit event ordering assumptions. 

As an event sender, you will be forced to send events in the order they occurred. If you are unable to send one event (for example due to an illegal null field in your payload), your only option is to stop sending events completely, let an engineer fix that one event, and then continue. Skipping an event and sending it later will violate the ordering guarantees that are part of your APIs contract.

The same problem exists on the consumers side: If you are unable to process an event for any reason, you have two options:
- Retry again and again to consume the event, until it works (for example, because an engineer built a workaround). If the error is just a temporary database hick up, that is fine. If it is a software bug, event consumption will effectively freeze until an engineer fixes the bug. Even if all the other following events would be perfectly consumable. Freezing in a live environment seems no acceptable solution.
- Log the event (or put it into a dead letter queue), skip it and continue. This approach solves the "freezing" problem, but comes with it's own downside: You will not be able to reconstruct the application state reliably and (semi-) automatically, because replaying the failed events later will violate the ordering guarantees and therefore give undefined results and probably lead to data corruption/data loss.

As a conclusion, relying on message ordering in your event based architecture is a pretty bad idea. 

Sending commutative events, that can be sent and received out of order will solve these problems. As a consumer, even if i could not process one of the events for some time, i will eventually converge into a consistent state again. It will probably take some more thoughts during system design to get those right, but i have not seen cases in the last four years where this was an unsolvable problem.

So, here is my Pull Request: https://github.com/zalando/restful-api-guidelines/pull/178

I would even say, let's make it a *must* instead of a *should*, but since i do not know if i've missed some essential requirements, i sticked with the should.

So: feedback welcome.

BGehrels:
fixes #170

### DIFF
--- a/events/events.md
+++ b/events/events.md
@@ -144,9 +144,11 @@ The `eid` property is part of the standard metadata for an event and gives the e
 
 Note that uniqueness checking of the `eid` is not enforced by Nakadi at the event type or global levels and it is the responsibility of the producer to ensure event identifiers do in fact distinctly identify events. A straightforward way to create a unique identifier for an event is to generate a UUID value.
 
-## {{ book.should }} Design your set of events to be commutative
+## {{ book.should }} Design for commutativity
 
-Events are commutative, if the can be received in an arbitrary order without changing the processing result. Together with the [idempotency requirement](#must-use-unique-event-identifiers), commutativity allows you to build very resilient systems. If your consumer fails to process a received event, it can continue with the rest of the events and reprocess the failed event later. The same holds true for the producer: If the producer fails to send an event, it can reschedule it for later out-of-order delivery without breaking the API contract.
+Event consumption is commutative, if events can be received in an arbitrary order without changing the processing result. Together with the [idempotency requirement](#must-use-unique-event-identifiers), commutativity allows you to build very resilient systems. If your consumer fails to process a received event, it can continue with the subsequent events and retry the failed event later. The same holds true for the producer: If the producer fails to send an event, it can reschedule it for later out-of-order delivery without corrupting the processing result.
+
+To enable commutative event consumption, the events sent by the producer must be able to be consumed out of order. They must either contain enough information to reorder them if needed during consumption, or your domain must be designed in a way that order becomes irrelevant.
 
 For data change events, this can be achieved by sending
 - the entity Id
--- a/events/events.md
+++ b/events/events.md
@@ -148,7 +148,7 @@ Note that uniqueness checking of the `eid` is not enforced by Nakadi at the even
 
 Events that are designed for [idempotent](#must-use-unique-event-identifiers) out-of-order processing allow for extremely resilient systems: If processing an event fails, consumers and producers can skip/delay/retry it without stopping the world or corrupting the processing result. 
 
-To enable this freedom of processing, events must explicitly designed for idempotent out-of-order processing. They must either contain enough information to infer the order of them during consumption or your domain must be designed in a way that order becomes irrelevant. 
+To enable this freedom of processing, you must explicitly design for idempotent out-of-order processing: Either your events must contain enough information to infer their original order during consumption or your domain must be designed in a way that order becomes irrelevant.
 
 As common example similar to data change events, idempotent out-of-order processing can be supported by sending the following information:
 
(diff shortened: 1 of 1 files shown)


events/events.md
```diff
@@ -144,6 +144,18 @@ The `eid` property is part of the standard metadata for an event and gives the e
 
 Note that uniqueness checking of the `eid` is not enforced by Nakadi at the event type or global levels and it is the responsibility of the producer to ensure event identifiers do in fact distinctly identify events. A straightforward way to create a unique identifier for an event is to generate a UUID value.
 
+## {{ book.should }} Design for commutativity
```
tkrop: We should better better use less mathematical terms here: e.g. "{{ book.should }} design for out-of-order idempotent processing"

The subsequent paragraph could be reformulated as follows:

Events that are designed for idempotent out-of-order processing allow for extremely resilient systems, as consumers and producers can skip/delay events, e.g. if they fail, without stopping the world or corrupting the result. To enable this freedom of processing, events must explicitly designed for idempotent out-of-order processing. There are many ways to achieve this goal depending on the underlying processes to implement.

As common example similar to data change events, idempotent out-of-order processing can be supported by sending the following information:

- the process/entity identifier,
- a monotonically increasing ordering key with respect to the entity (e.g. a version identifier), and
- the process/resource state after the change.

A receiver that is interested in the current state can then ignore events that are older than the last processed event of each resource. A receiver interested in the history of a resource can merge all events of partially ordered sequence.

BGehrels: Reformulated your proposal a bit, what do you think of the current version?

events/events.md
```diff
@@ -144,6 +144,20 @@ The `eid` property is part of the standard metadata for an event and gives the e
...
+
+Events that are designed for [idempotent](#must-use-unique-event-identifiers) out-of-order processing allow for extremely resilient systems: If processing an event fails, consumers and producers can skip/delay/retry it without stopping the world or corrupting the processing result. 
+
+To enable this freedom of processing, events must explicitly designed for idempotent out-of-order processing. They must either contain enough information to infer the order of them during consumption or your domain must be designed in a way that order becomes irrelevant. 
+
+As common example similar to data change events, idempotent out-of-order processing can be supported by sending the following information:
+
+- the process/resource/entity identifier,
+- a [monotonically increasing ordering key](#should-provide-a-means-of-event-ordering) and
+- the process/resource state after the change.
```
maxim-tschumak: maybe an additional keyword `self-contained messages`

events/events.md
```diff
@@ -144,17 +144,19 @@ The `eid` property is part of the standard metadata for an event and gives the e
...
+Events that are designed for [idempotent](#must-use-unique-event-identifiers) out-of-order processing allow for extremely resilient systems: If processing an event fails, consumers and producers can skip/delay/retry it without stopping the world or corrupting the processing result. 
 
-To enable commutative event consumption, the events sent by the producer must be able to be consumed out of order. They must either contain enough information to reorder them if needed during consumption, or your domain must be designed in a way that order becomes irrelevant.
+To enable this freedom of processing, events must explicitly designed for idempotent out-of-order processing. They must either contain enough information to infer the order of them during consumption or your domain must be designed in a way that order becomes irrelevant.
```
tkrop: missing word: .. events must **be** explicitly designed ...

... but the wording is still not optimal. The first sentence does not work good together with the second clause of the second sentence. Can you try another formulation?

events/events.md
BGehrels: While i like the term in verbal communication (since it catches a lot of semantics in one term), i do not like it that much for specs because it lacks precision: You could argue a lot about what a "really" and "truly" self contained event is. One could, for example, say, that a really self contained data change event should contain the old _and_ new version of the resource. and so on and so on and so on.

events/events.md
BGehrels: You are right, thanks. rephrased that one.

events/events.md
```diff
@@ -148,15 +148,15 @@ Note that uniqueness checking of the `eid` is not enforced by Nakadi at the even
 
 Events that are designed for [idempotent](#must-use-unique-event-identifiers) out-of-order processing allow for extremely resilient systems: If processing an event fails, consumers and producers can skip/delay/retry it without stopping the world or corrupting the processing result. 
 
-To enable this freedom of processing, events must explicitly designed for idempotent out-of-order processing. They must either contain enough information to infer the order of them during consumption or your domain must be designed in a way that order becomes irrelevant. 
+To enable this freedom of processing, you must explicitly design for idempotent out-of-order processing: Either your events must contain enough information to infer their original order during consumption or your domain must be designed in a way that order becomes irrelevant.
```
tkrop: I'm not a fan of introducing 'you' into a single sentence of a paragraph that is else written without. But I'm also German and dislike this personal wording any way ;)

events/events.md
maxim-tschumak: @BGehrels We can add some reasoning about tolerating duplicates (`at-least-once` delivery guarantee), as described in #185. This is just another argument to design for idempotent processing.

ePaul:
@tfrauenstein you wanted to do a wording sweep here?

ePaul:
:+1:

ePaul:
Wording changes can be done later.

fmueller:
👍

PR #180, opened 2017-01-17
Stop suggesting timestamps for event ordering

### LINKED ISSUE #179: Stop suggesting timestamps for event ordering
In the event chapter, we suggest that

> Change events should contain a means of ordering events for a given entity (such as created or updated timestamps [...])

Clocks may jump forwards and backwards, two events may happen in the same microsecond, and when sending from a distributed environment, they may diverge between the systems. So relying on time as a ordering key is not a good idea.

BGehrels:
fixes #179

### DIFF
--- a/events/events.md
+++ b/events/events.md
@@ -54,7 +54,7 @@ Nakadi defines an event for signalling data changes, called a [DataChangeEvent](
 
 - Change events must identify the changed entity to allow aggregation of all related events for the entity.
 
-- Change events should contain a means of ordering events for a given entity. This can - among others - be a strictly monotonically increasing entity version or a message counter. Timestamps are not a good choice, since your system clock may jump and exact synchronization of clocks in distributed systems is hard (or even impossible). Note that basing events on data structures that can be converged upon (such as [CRDTs](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type) or [logical clocks](https://en.wikipedia.org/wiki/Logical_clock)) in a distributed setting are outside the scope of this guidance.
+- Change events should contain a means of ordering events for a given entity. This can - among others - be a strictly monotonically increasing entity version or a message counter. Timestamps are not necessarily a good choice, since your system clock may jump backwards, two events may occur in the same microsecond and exact synchronization of clocks in distributed systems is difficult. If you use timestamps to indicate event ordering, you must carefully ensure they move forward for each event. Note that basing events on data structures that can be converged upon (such as [CRDTs](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type) or [logical clocks](https://en.wikipedia.org/wiki/Logical_clock)) in a distributed setting are outside the scope of this guidance.
 
 - Change events must be published reliably by the service.
 
--- a/events/events.md
+++ b/events/events.md
@@ -57,7 +57,7 @@ Nakadi defines an event for signalling data changes, called a [DataChangeEvent](
 - Change events must be published reliably by the service.
 
 ## {{ book.should }} Provide a means of event ordering
-Events should contain a means of reordering events for a given entity. This can be done in many ways, two common ways are:
+Events should contain a way to restore the partial ordering of events. This can be done in many ways, two common ways are:
 
 - A strictly monotonically increasing entity version or a message counter.
 
--- a/events/events.md
+++ b/events/events.md
@@ -54,7 +54,7 @@ Nakadi defines an event for signalling data changes, called a [DataChangeEvent](
 
 - Change events must identify the changed entity to allow aggregation of all related events for the entity.
 
-- Change events should contain a means of ordering events for a given entity (such as created or updated timestamps that are assured to move forwards with respect to the entity, or version identifiers provided by some databases). Note that basing events on data structures that can be converged upon (such as [CRDTs](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type) or [logical clocks](https://en.wikipedia.org/wiki/Logical_clock)) in a distributed setting are outside the scope of this guidance.
+- Change events should contain a means of ordering events for a given entity. This can - among others - be a strictly monotonically increasing entity version or a message counter. Timestamps are not a good choice, since your system clock may jump and exact synchronization of clocks in distributed systems is hard (or even impossible). Note that basing events on data structures that can be converged upon (such as [CRDTs](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type) or [logical clocks](https://en.wikipedia.org/wiki/Logical_clock)) in a distributed setting are outside the scope of this guidance.
 
 - Change events must be published reliably by the service.
 
--- a/events/events.md
+++ b/events/events.md
@@ -53,10 +53,19 @@ At the moment we cannot state whether it's best practice to publish all the even
 Nakadi defines an event for signalling data changes, called a [DataChangeEvent](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml#/definitions/DataChangeEvent). When publishing events that represents created, updated, or deleted data, change event types must be based on the DataChangeEvent category.
 
 - Change events must identify the changed entity to allow aggregation of all related events for the entity.
-
-- Change events should contain a means of ordering events for a given entity. This can - among others - be a strictly monotonically increasing entity version or a message counter. Timestamps are not necessarily a good choice, since your system clock may jump backwards, two events may occur in the same microsecond and exact synchronization of clocks in distributed systems is difficult. If you use timestamps to indicate event ordering, you must carefully ensure they move forward for each event. Note that basing events on data structures that can be converged upon (such as [CRDTs](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type) or [logical clocks](https://en.wikipedia.org/wiki/Logical_clock)) in a distributed setting are outside the scope of this guidance.
-
+- Change events should [contain a means of ordering](#provide-a-means-of-event-ordering) events for a given entity.
 - Change events must be published reliably by the service.
+
+## {{ book.should }} Provide a means of event ordering
+Events should contain a means of reordering events for a given entity. This can be done in many ways, two common ways are:
+
+- A strictly monotonically increasing entity version or a message counter.
+
+- Nakadi provides a concept (similar to vector clocks) for expressing this order by adding parent event identifiers to the event meta data. This allows clients to check whether they have processed all parent events before they continue with the current event.
+
+System timestamps are not necessarily a good choice, since your system clock may jump backwards, two events may occur in the same microsecond and exact synchronization of clocks in distributed systems is difficult. If you use timestamps to indicate event ordering, you must carefully ensure they move forward for each event. 
+
+Note that basing events on data structures that can be converged upon (such as [CRDTs](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type) or [logical clocks](https://en.wikipedia.org/wiki/Logical_clock)) in a distributed setting are outside the scope of this guidance.
 
 ## {{ book.should }} Use the hash partition strategy for Data Change Events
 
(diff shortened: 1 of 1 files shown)


fmueller:
👍

events/events.md
```diff
@@ -54,7 +54,7 @@ Nakadi defines an event for signalling data changes, called a [DataChangeEvent](
...
 
-- Change events should contain a means of ordering events for a given entity (such as created or updated timestamps that are assured to move forwards with respect to the entity, or version identifiers provided by some databases). Note that basing events on data structures that can be converged upon (such as [CRDTs](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type) or [logical clocks](https://en.wikipedia.org/wiki/Logical_clock)) in a distributed setting are outside the scope of this guidance.
+- Change events should contain a means of ordering events for a given entity. This can - among others - be a strictly monotonically increasing entity version or a message counter. Timestamps are not a good choice, since your system clock may jump and exact synchronization of clocks in distributed systems is hard (or even impossible). Note that basing events on data structures that can be converged upon (such as [CRDTs](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type) or [logical clocks](https://en.wikipedia.org/wiki/Logical_clock)) in a distributed setting are outside the scope of this guidance.
```
dehora: > Clocks may jump forwards and backwards, two events may happen in the same microsecond, and when sending from a distributed environment, they may diverge between the systems

Sure. The guidance says use timestamps when you can keep them moving forward, not use them blindly:

> ...such as created or updated timestamps that are assured to move forwards with respect to the entity

I agree the existing text above could be better, it's a bit of a run-on sentence. "such as created or updated timestamps when they are assured to move forwards with respect to the entity" might have made the implication clear.

I like the basic clarification, but I'd suggest altering the pr from this

"""
Timestamps are not a good choice, since your system clock may jump and exact synchronization of clocks in distributed systems is hard (or even impossible).
"""

to this:

"""
Timestamps are not necessarily a good choice, since your system clock may jump backwards and exact synchronization of clocks in distributed systems is difficult. If you use timestamps to indicate event ordering, you must carefully ensure they move forward. 
"""

Fwiw:

- I don't see shifting from timestamps to strictly monotonic counters as an slam dunk better option in a distributed setting (I need a centralised something to coordinate increments, or investigate options beyond the general state of the art like hybrid logical clocks or true-time for timestamps, or the crdt/vector clock examples already mentioned). But I think it's ok to leave that option mentioned. 

- I think we may want to look at something more fundamental in Nakadi events to allow happens-before to be clearly expressed. The data change event is pretty loose at the moment. A topic for another pr probably.

- In general, I'd like us to be careful making recommendations that impose engineering requirements on teams versus highlighting the issues that let them make appropriate decisions based on their actual needs.

BGehrels:
@dehora Great feedback, thank you! merged it into the text and extended it a bit:

They have to move forward *for each event*. Two events, that occured one after another but in the same second may not have the same ordering key.

tfrauenstein:
discussed in JF:  give hint that parent id may also be used as ordering criteria

events/events.md
```diff
@@ -54,7 +54,7 @@ Nakadi defines an event for signalling data changes, called a [DataChangeEvent](
...
+- Change events should contain a means of ordering events for a given entity. This can - among others - be a strictly monotonically increasing entity version or a message counter. Timestamps are not necessarily a good choice, since your system clock may jump backwards, two events may occur in the same microsecond and exact synchronization of clocks in distributed systems is difficult. If you use timestamps to indicate event ordering, you must carefully ensure they move forward for each event. Note that basing events on data structures that can be converged upon (such as [CRDTs](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type) or [logical clocks](https://en.wikipedia.org/wiki/Logical_clock)) in a distributed setting are outside the scope of this guidance.
```
tkrop: I'm not happy to elaborate that much on timestamps in this context. Firstly, the original formulation is in my opinion precise enough as it proposes an abstract concept of timestamps (including logical clocks etc.) only in a very restricted scope that ensures ordering and that can easily be achieved by most business processes implementations. Secondly, it opens up a huge about clock synchronization issues, that majorly depend on a distributed timestamp creation concept suggested in Zalando, not implied or suggested with the original formulation.

In my opinion the main perception problem of the original sentence is that the combination of `created` or `updated` with `timestamp` for most readers this suggest that it is created by a system clock. To address this aspect I would suggest to reformulate the original sentence to:

- Change events should contain a means of ordering events for a given entity, such as a timestamps assured to move forwards with respect to the entity, e.g. a version identifiers provided by some databases. Note that basing events on data structures that can be converged upon, such as [CRDTs](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type), [logical clocks](https://en.wikipedia.org/wiki/Logical_clock), and [vector clocks](
https://en.wikipedia.org/wiki/Vector_clock) in a distributed setting are outside the scope of this guidance.

tkrop:
@tfrauenstein @BGehrels I think, that the `parent_id` concept should not mentioned in the concerned section, because they are way to complex and unnecessary for data integration needs. Besides, we excluded them from the discussion in this section intentionally with the note sentence.

If we want to bring this in, it should be an new section may be as follows:

{{ book.could }} Consider using Parent Event Identifiers

While the partition key in Nakadi provides grate means of ordering events for certain requirements, this solution may be insufficient for other use cases. If correct event processing needs to be based on the (partial) order of events independent of partitions, Nakadi provides a simple concept (similar to [logical clocks](https://en.wikipedia.org/wiki/Logical_clock), and [vector clocks](
https://en.wikipedia.org/wiki/Vector_clock)) for expressing this order by adding parent identifiers to the event meta data. This allows clients to check whether they have processed all parent events before they continue with the current event.

BGehrels:
I guess _if_ we use the term "timestamp", most people that are not into distributed computing will think of system clocks aka unix timestamps. So we should then make their tradeoffs very very clear.

I like your idea of not inflating this section much more. I would try to extract the whole ordering foo from the data change event into a section of it's own if no one objects. It can then contain the parent_id suggestion as one alternative. This way, we will have everything related to ordering in one place (section), making it much easier for the reader.

BGehrels:
Extracted a section of it's own, linked it in the data change event section, added parent_eid as one possible implementation concept, gave the timestamp tradeoffs a seperate side note paragraph. Ready for review again.

events/events.md
```diff
@@ -53,10 +53,19 @@ At the moment we cannot state whether it's best practice to publish all the even
...
 
 - Change events must identify the changed entity to allow aggregation of all related events for the entity.
+- Change events should [contain a means of ordering](#should-provide-a-means-of-event-ordering) events for a given entity.
+- Change events must be published reliably by the service.
 
-- Change events should contain a means of ordering events for a given entity (such as created or updated timestamps that are assured to move forwards with respect to the entity, or version identifiers provided by some databases). Note that basing events on data structures that can be converged upon (such as [CRDTs](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type) or [logical clocks](https://en.wikipedia.org/wiki/Logical_clock)) in a distributed setting are outside the scope of this guidance.
+## {{ book.should }} Provide a means of event ordering
+Events should contain a means of reordering events for a given entity. This can be done in many ways, two common ways are:
```
maxim-tschumak: IMO we could explicitly mention what we want to achieve here: `partial ordering` of the events.

BGehrels: Done, like the wording much more now.

events/events.md
```diff
@@ -53,10 +53,19 @@ At the moment we cannot state whether it's best practice to publish all the even
 Nakadi defines an event for signalling data changes, called a [DataChangeEvent](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml#/definitions/DataChangeEvent). When publishing events that represents created, updated, or deleted data, change event types must be based on the DataChangeEvent category.
 
 - Change events must identify the changed entity to allow aggregation of all related events for the entity.
+- Change events should [contain a means of ordering](#should-provide-a-means-of-event-ordering) events for a given entity.
+- Change events must be published reliably by the service.
 
-- Change events should contain a means of ordering events for a given entity (such as created or updated timestamps that are assured to move forwards with respect to the entity, or version identifiers provided by some databases). Note that basing events on data structures that can be converged upon (such as [CRDTs](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type) or [logical clocks](https://en.wikipedia.org/wiki/Logical_clock)) in a distributed setting are outside the scope of this guidance.
+## {{ book.should }} Provide a means of event ordering
```
tkrop: I'm not sure whether this is to strong Providing means for (partial) ordering is *only necessary*, if their are *direct* or *indirect* **execution dependencies**. On the other side, from a pragmatic approach, when designing an event, people do not think about all possible use-cases, and some might occur later, that have not anticipated ordering needs.

Anyhow, the formulation is still not consistent and is missing my nice introduction that argues why at all one should be required to provide additional ordering besides the natural order in partitions. So I suggest to change the paragraph as follows:

```
{{ book.could }} Provide means for explicit event ordering

While the partition key in Nakadi provides great means of ordering events for certain requirements, this solution may be insufficient for other use cases.  If correct event processing needs to be based on the (partial) order of events independent of partitions, this could be done in the following ways we want to outline here:

- Using a strictly monotonically increasing entity version, e.g. as created by a database, to allow for partial ordering of events with respect to entities as required by data change events.
- Using a strictly monotonically increasing message counter, e.g. per message source, to allow for partial ordering of events with respect to event producers as required for monitoring.
- Adding parent event identifiers to the Nakadi event meta data (a concept similar to [vector clocks](
https://en.wikipedia.org/wiki/Vector_clock)), to allow for partial ordering of related events in a business flow.

While timestamps are widely used for monotonically increasing sequence and may be sufficient in some situations, they are not a good choice in distributed systems, since synchronization of clocks is difficult, system clock may jump backward or forward to compensate drifts or leap-seconds, while multiple events may occur in the same microsecond. If you want to use timestamps to indicate event ordering, you must carefully ensure that your designated event order is not effected by the before mentioned effects.

**Note** that basing events on data structures that can be converged upon, such as [CRDTs](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type), [logical clocks](https://en.wikipedia.org/wiki/Logical_clock), and [vector clocks](https://en.wikipedia.org/wiki/Vector_clock) in a distributed setting are outside the scope of this guidance.
```

BGehrels:
Implemented the next round of (verbal and written) Feedback:
- Added a nice introductory sentence to the paragraph
- Removed the reference to vector clocks on the parent_eid bullet point.  It is part of the last sentence and a bit too mathematical here. (And, to be honest, those are not *really* comparable to vector clocks)
- Used a lot of Tronjes proposed wording

events/events.md
```diff
@@ -53,10 +53,22 @@ At the moment we cannot state whether it's best practice to publish all the even
...
 
 - Change events must identify the changed entity to allow aggregation of all related events for the entity.
+- Change events should [contain a means of ordering](#should-provide-a-means-of-event-ordering) events for a given entity.
+- Change events must be published reliably by the service.
 
-- Change events should contain a means of ordering events for a given entity (such as created or updated timestamps that are assured to move forwards with respect to the entity, or version identifiers provided by some databases). Note that basing events on data structures that can be converged upon (such as [CRDTs](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type) or [logical clocks](https://en.wikipedia.org/wiki/Logical_clock)) in a distributed setting are outside the scope of this guidance.
+## {{ book.should }} Provide a means for explicit event ordering
 
-- Change events must be published reliably by the service.
+While Nakadi guarantees ordering per partition, some common error cases may require your consumers to reconstruct message ordering and
```
tkrop: I think the argumentation should be more general here. It sounds like their is only in 'error cases' the need for reconstructing the partial ordering of events.

ePaul:
:+1:

tfrauenstein:
Yes, in reasonably shape now. 
I would like to give some more concrete guidance of the order handling variants and their trade-offs -- but this should not block the merge of this PR. 

One aspect I would like to change now: as Tronje proposed, let us take parentId out of this guideline -- may be worth a separate guideline, but not for the time being.

ePaul:
:+1:

ePaul:
@tfrauenstein and @BGehrels will afterwards meet for editorial improvements (in a separate PR).

tkrop:
👍

PR #181, opened 2017-01-18
Remove Avoid Javascript Keywords Rule

### LINKED ISSUE #167: Does the rule Avoid JavaScript Keywords make sense?
What was the reasoning behind the rule [Reserved JavaScript keywords should be avoided](http://zalando.github.io/restful-api-guidelines/json-guidelines/JsonGuidelines.html#should-reserved-javascript-keywords-should-be-avoided) in our API guidelines?

@dehora Can you follow up with more explanation regarding *security and sanity reasons*?

fmueller:
Fixes #167

### DIFF
--- a/changelog/Changelog.md
+++ b/changelog/Changelog.md
@@ -8,6 +8,7 @@ To see a list of all changes, please have a look at the [commit list in Github](
 
 ## Rule Changes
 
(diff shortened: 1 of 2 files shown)


tfrauenstein:
👍

hjacobs:
:+1:

PR #182, opened 2017-02-02
Add guideline for batch and bulk requests using 207

### LINKED ISSUE #127: Provide guidelines on batch/bulk requests and 207
WebDAV's 207 response code comes up from time to time and there's no specific guidance on using it, or not using it. More generally the guidelines don't cover idioms for batch requests. It would be worth looking at this area as bulk/batch processing is a common case. Even clearing up API terminology would be useful - "batch" in the industry is used to describe multiple http requests send inside a single request, as well as sending multiple records in a single object - "bulk" is also used for the latter case. There is further detail around whether the bulk/batch request results in an async "job" that can be checked for processing status.

An alternate approach taken to marking partial successes with 207 is to document 200 as successfully _processing_ a batch, but informing clients they need to check the details of the response entity to see what the processing results were for each item. Let's call this "client-must-check-response". For example this is how some AWS APIs work (Kinesis, SQS, DynamoDB). 

207 comes from WebDAV and it's context specific that they used 2xx instead of some other range. There's probably no ideal response class for a bulk response, but a downside is that 207 maps to the 2xx class which means generic http clients need to treat it as a special case when 207 is being used to describe partial failures/success. See [here](https://gist.github.com/dehora/409a14c52eece610c501) for more details on the history/intent.

tkrop:
Fixes #127

### DIFF
--- a/changelog/Changelog.md
+++ b/changelog/Changelog.md
@@ -8,6 +8,7 @@ To see a list of all changes, please have a look at the [commit list in Github](
 
 ## Rule Changes
 
+* `2017-02-02:` Add guideline for batch and bulk request
 * `2017-01-18:` Removed "Avoid Javascript Keywords" rule
 * `2017-01-05:` Clarification on the usage of the term "REST/RESTful"
 * `2016-12-07:` Introduced "API as a Product" principle
--- a/changelog/Changelog.md
+++ b/changelog/Changelog.md
@@ -8,7 +8,7 @@ To see a list of all changes, please have a look at the [commit list in Github](
 
 ## Rule Changes
 
-* `2017-02-02:` Add guideline for batch and bulk request
+* `2017-02-02:` Added guideline for batch and bulk request
 * `2017-01-18:` Removed "Avoid Javascript Keywords" rule
 * `2017-01-05:` Clarification on the usage of the term "REST/RESTful"
 * `2016-12-07:` Introduced "API as a Product" principle
--- a/http/Http.md
+++ b/http/Http.md
@@ -160,6 +160,7 @@ different HTTP methods on resources.
 | 201  | Created - Returned on successful entity creation. You are free to return either an empty response or the created resource in conjunction with the Content-Location header. (More details found in the [Common Headers section](../headers/CommonHeaders.md).) *Always* set the Location header. | POST, PUT |
 | 202  | Accepted - The request was successful and will be processed asynchronously. | POST, PUT, DELETE, PATCH |
 | 204  | No content - There is no response body | PUT, DELETE |
+| 207  | Multi-Status - The response body contains multiple status informations for different parts of a batch/bulk request. See ["Use 207 for Batch or Bulk Requests"](#use-207-for-batch-or-bulk-requests). | POST |
 
 ### Redirection Codes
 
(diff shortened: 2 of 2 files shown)


changelog/Changelog.md
```diff
@@ -8,6 +8,7 @@ To see a list of all changes, please have a look at the [commit list in Github](
 
 ## Rule Changes
 
+* `2017-02-02:` Add guideline for batch and bulk request
```
fmueller: Minor thing: we favor to use past tense in changelog. "Added..."

http/Http.md
```diff
@@ -223,6 +224,18 @@ Even though they might not be documented - they may very much occur in productio
 
 Functional errors on the other hand, that convey domain-specific semantics, must be documented and are strongly encouraged to be expressed with [*Problem types*](../common-data-objects/CommonDataObjects.md#must-use-common-error-return-objects).
 
+## {{ book.must }} Use 207 for Batch or Bulk Requests
+
+Some APIs are required to provide either *batch* or *bulk* requests using POST for performance reasons, i.e. for communication and processing efficiency. In this case services may be in need to signal multiple response codes for each part of an batch or bulk request. As HTTP does not provides proper guidance for handling batch an bulk requests and responses, we herewith define the following approach:
```
fmueller: In my understanding, batch and bulk both mean same thing. What the reasoning behind mentioning both terms? It kinda confuses me.

fmueller: Typo: "As HTTP does not provide" (without s at the end)

http/Http.md
```diff
@@ -223,6 +224,18 @@ Even though they might not be documented - they may very much occur in productio
...
+
+- A batch or bulk request **always** has to respond with HTTP status code **207**, unless it encounters a generic or unexpected failure before looking at individual parts.
+- A batch or bulk response with status code 207 **always** returns a multi-status object containing sufficient status and/or monitoring information for each part of the batch or bulk request.
+- A batch or bulk request may result in a status code 400/500, only iff the service encounters a failure before looking at individual parts or iff an unanticipated failure occurs.
```
fmueller: Typo: off -> of

http/Http.md
```diff
@@ -223,6 +224,18 @@ Even though they might not be documented - they may very much occur in productio
...
+
+The before rules apply *even in the case* that processing of all individual part *fail* or each part is executed *asynchronously*! They are intended to allow clients to act on batch and bulk responses by inspecting the individual results in a consistent way.
```
fmueller: Wording: This rule applies

http/Http.md
```diff
@@ -223,6 +224,18 @@ Even though they might not be documented - they may very much occur in productio
...
+
+**Note**: while a *batch* defines a collection of requests triggering independent processes, a *bulk* defines a collection of independent resources created or updated together in one request. With respect to response processing this distinction normally does not matter.
```
fmueller: Ah, here you are giving the term definitions. Can you move this to the beginning? :)

http/Http.md
tkrop: replaced "an" against "/" in addition.

http/Http.md
tkrop: @fmueller iff? This is an abbreviation for "if and only if". But you are still correct, I have to remove the beginning "only" 😸

http/Http.md
tkrop: I didn't found a proper place, that would not break the explanation flow. So I put it here to the end for readers, that do not understand this distinction. I'm not sure whether this explanation is really necessary, but I decided to stick as close to the original Idea of @dehora as possible.

http/Http.md
tkrop: See below.

http/Http.md
tkrop: I'm not sure. For my understanding it should be plural, because I have three rules - especially as this remark addresses the first two rules. Using singular could be understood as it only addresses the last one.

http/Http.md
```diff
@@ -226,11 +226,11 @@ Functional errors on the other hand, that convey domain-specific semantics, must
...
+Some APIs are required to provide either *batch* or *bulk* requests using POST for performance reasons, i.e. for communication and processing efficiency. In this case services may be in need to signal multiple response codes for each part of an batch or bulk request. As HTTP does not provide proper guidance for handling batch/bulk requests and responses, we herewith define the following approach:
 
 - A batch or bulk request **always** has to respond with HTTP status code **207**, unless it encounters a generic or unexpected failure before looking at individual parts.
 - A batch or bulk response with status code 207 **always** returns a multi-status object containing sufficient status and/or monitoring information for each part of the batch or bulk request.
-- A batch or bulk request may result in a status code 400/500, only iff the service encounters a failure before looking at individual parts or iff an unanticipated failure occurs.
+- A batch or bulk request may result in a status code 400/500, iff the service encounters a failure before looking at individual parts or iff an unanticipated failure occurs.
```
fmueller: Not everyone is aware of abbreviations like iff. What about writing it in the long version? :)

http/Http.md
fmueller: Ok, in my understanding this is one rule (one rule per headline). That's why I was confused.

http/Http.md
tkrop: I know, but I like the aspect, that in mathematical sense it is a very strict rule, while people without deep engl. math-knowledge would interpret it as a spelling mistake and apply this rule a it more relaxed, which would be okayish in most circumstances. Writing it the long way would eliminate this small freedom ;). But, iff you insists ...

fmueller: I got your sense of humor :) but our guidelines should be as simple to read as possible - also without such a knowledge. I would really change it.

fmueller:
👍

fmueller:
👍

fmueller:
@codekultur @ePaul What are your thoughts about this? Can we merge it?

codekultur:
:+1

codekultur:
👍

PR #187, opened 2017-02-22
Document conditional requests If-Match/If-None-Match

### LINKED ISSUE #154: Document If-Match + If-None-Match for PUT
We should mention the usage of conditional requests for updates/creates, to avoid the lost update problem.

* `If-None-Match: *` when creating a resource with PUT makes sure you don't overwrite a resource created by someone else without noting.
* `If-Match: ...` when updating a resource with PUT (or PATCH) makes sure you don't overwrite changes by someone else without noting.

tkrop:
Fixes #154


ePaul:
:+1:

duergner:
👍

PR #191, opened 2017-02-27
Add hint that validation logic is part of API contract

### LINKED ISSUE #177: Mention validation logic as part of API and therefore considerable for possible breaking changes
What happened? One team changed validation logic because it was too strict. This caused issues in writing api client user of their service. Discussion in the API guild today made clear that it was mostly missing communication between the teams.

It's an open discussion topic if we should mention in our guidelines that validations are also part of the API and should not break in a non-backward compatible way.

fmueller:
Fixes #177 – not yet ready for merge.

### DIFF
--- a/compatibility/Compatibility.md
+++ b/compatibility/Compatibility.md
@@ -7,6 +7,10 @@ running. Consumers usually have independent release lifecycles, focus on
 stability, and avoid changes that do not provide additional value. APIs are
 service contracts that cannot be broken via unilateral decisions.
 
+An API contract consists of endpoints and allowed methods on these, payloads
+and also validation logic, e.g. changing validation logic would be a breaking
+API change.
+
 There are two techniques to change APIs without breaking them:
 
 - follow rules for compatible extensions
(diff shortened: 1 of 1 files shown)


compatibility/Compatibility.md
```diff
@@ -7,14 +7,18 @@ running. Consumers usually have independent release lifecycles, focus on
 stability, and avoid changes that do not provide additional value. APIs are
 service contracts that cannot be broken via unilateral decisions.
 
+An API contract consists of endpoints and allowed methods on these, payloads
+and also validation logic, e.g. changing validation logic would be a breaking
+API change.
```
ePaul: > ... changing validation logic *to be more restrictive* would be a breaking API change.

Changing to be more relaxed (i.e. allow stuff which was not allowed before) should not hurt any client.

tkrop: I think adding this sentence here would it to invisible. We are currently lacking a guideline what a good API schema definition should look like. It is distributed over HTTP, Resources, Compatibility and many more locations. I would like to have an own topic under principles for this, but than we would have no clear addressable rule. So I also do not know how to handle this.

dehora: A reasonable step would be to give this its own heading. That will make it linkable, etc. / @ePaul @tkrop @fmueller

ePaul:
From the meeting: Will be incorporated in #248.

PR #202, opened 2017-03-10
Update Pagination.md

chilicoder:
it looks like HAL spec complaint, with an exception of "products" properties. It should be wraped in "_embedded". 
see for instance
http://phlyrestfully.readthedocs.io/en/latest/halprimer.html#collections
https://softinstigate.atlassian.net/wiki/display/RH/API+tutorial#APItutorial-GetallDocumentsfromtheCollection
http://docs.clarify.io/api/

I wrapped it up in `_embedded` and also divided `count_total` into `count` and `total`

### DIFF
--- a/pagination/Pagination.md
+++ b/pagination/Pagination.md
@@ -78,14 +78,17 @@ Content-Type: application/json
       "href": "http://catalog-service.zalando.net/products?offset=5&limit=5"
     }
   },
-  "total_count": 42,
-  "products": [
-    {"id": "7e4ab218-1772-11e6-892c-836df3feeaee"},
-    {"id": "9469725a-1772-11e6-83c2-ab22ac368913"},
-    {"id": "a3625d80-1772-11e6-9213-d3a20f9e6bf4"},
-    {"id": "a802f070-1772-11e6-a772-c3020d55eb5f"},
-    {"id": "adb407ac-1772-11e6-b255-078fb28ff55b"}
-  ]
+  "count": 5,
+  "total": 42,
+  "_embedded":{
+    "products": [
+      {"id": "7e4ab218-1772-11e6-892c-836df3feeaee"},
+      {"id": "9469725a-1772-11e6-83c2-ab22ac368913"},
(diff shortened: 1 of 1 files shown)


pagination/Pagination.md
```diff
@@ -78,14 +78,17 @@ Content-Type: application/json
...
   },
-  "total_count": 42,
-  "products": [
-    {"id": "7e4ab218-1772-11e6-892c-836df3feeaee"},
-    {"id": "9469725a-1772-11e6-83c2-ab22ac368913"},
-    {"id": "a3625d80-1772-11e6-9213-d3a20f9e6bf4"},
-    {"id": "a802f070-1772-11e6-a772-c3020d55eb5f"},
-    {"id": "adb407ac-1772-11e6-b255-078fb28ff55b"}
-  ]
+  "count": 5,
```
whiskeysierra: This is exactly always the size of the array, so pretty redundant, I'd say.

We chose `total_count` because we started with a header with a similar name: `X-Total-Count`.

ePaul:
I guess the question is what exactly is part of the resource itself, and what are related resources which are embedded. I would say the list is not a related resource here, but the original resource.
Having a resource consisting of just `count` and `total` seems a bit boring.

chilicoder:
@ePaul but its a collection. So, it has only properties, that collection has. I checked several HAL complaint interfaces (see links in PR desc) they all has this structure

tkrop:
Hi @chilicoder,

thank you for your pull request. You are certainly right, that many API designers, implementers are using HAL in the described way. Anyhow, we have decided in the API guild, that we do not want to recommend writing APIs in a way that moves the content of a collection request into the meta data of HAL. We regard the collection as a primary resource itself and therefor expect it to be modeled accordingly. In addition, we are currently reconsidering our recommendation for HAL, as its rules and its way of separating meta information from resources seems not to match our API first approach. We already started to relax our rules to enable simpler control flow based APIs.

Therefore, we have decided to kindly rejected your pull request, as we do not want to guide developers in this direction.

Best regards
Tronje

PR #212, opened 2017-03-30
Model complete business processes

### LINKED ISSUE #183: Recommend rules for restful API to model (complete) business processes
Recently, I observed that we get very small API definitions for review which cover only parts of business processes - or worse provide only a thin wrapper around the data base resource.  As effect the business processes and resources are distributed over many APIs creating invisible implicit dependencies, which make it hard to understand the logic of the business processes and the relation between resources.

In my opinion this is an API design (and service design) we should discourage in the guidelines - as we currently try do in our API workshop.

Therefore, I suggest add rules to the guideline that should to encourage the following design practices:

1. Resources should be designed to represent business processes, e.g. like orders, to prevent thin wrappers around databases that normally tend to shift business logic from service to clients.
2. An API should contain the complete business processes with respect to its representing resources, to enable clients to understand the business process, to foster a consistent design of the business process, to allow for synergies from description and implementation perspective, and to prevent implicit invisible dependencies between APIs.

Beside, this new rules, we should rephrase our rules about resource limits to make clear, that limitation rules are targeted to encourage the detection of complete and sometimes generic sub-processes, that should be cut out to reduce the complexity of an API and service.

codekultur:
Fixes #183

### DIFF
--- a/changelog/Changelog.md
+++ b/changelog/Changelog.md
@@ -9,7 +9,8 @@ To see a list of all changes, please have a look at the [commit list in Github](
 ## Rule Changes
 
 * `2017-03-30:` Added rule that APIs should be modelled around business processes
-* `2017-02-23:` Added rule that permissions on resources in events must correspond to permissions on API resources
+* `2017-02-28:` Extended information about how to reference sub-resources and the usage of composite identifiers in the [Resources](../resources/Resources.md#-bookmust--identify-resources-and-sub-resources-via-path-segments) part.
+* `2017-02-22:` Added guidance for conditional requests with If-Match/If-None-Match
 * `2017-02-02:` Added guideline for batch and bulk request
 * `2017-02-01:` Discouraged use of [Content-Location](../headers/CommonHeaders.md#could-use-contentlocation-header) header by requiring to use [Location header](../headers/CommonHeaders.md#should-use-location-header-instead-of-contentlocation-header).
 * `2017-01-18:` Removed "Avoid Javascript Keywords" rule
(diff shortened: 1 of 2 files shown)


changelog/Changelog.md
```diff
@@ -8,8 +8,8 @@ To see a list of all changes, please have a look at the [commit list in Github](
 
 ## Rule Changes
 
-* `2017-02-28:` Extended information about how to reference sub-resources and the usage of composite identifiers in the [Resources](../resources/Resources.md#-bookmust--identify-resources-and-sub-resources-via-path-segments) part.
-* `2017-02-22:` Added guidance for conditional requests with If-Match/If-None-Match
```
ePaul: Did you intentionally remove those changelog entries?

codekultur: No, sorry, that was deleted by accident. Thanks for pointing that out!

tkrop:
@codekultur short and precise! I like it 😸

tkrop:
👍

resources/Resources.md
```diff
@@ -20,7 +25,7 @@ clients to specify their needs for more/less information by supporting filtering
 ## {{ book.must }} Keep URLs Verb-Free
 
 The API describes resources, so the only place where actions should appear is in the HTTP methods.
-In URLs, use only nouns.
+In URLs, use only nouns. Instead of thinking of actions (verbs), it's often helpful to think about putting a message in a letter box: E.g. indead of having the verb *cancel* in the url, think of sending a message to cancel an order to the *cancallations* letter box on the server side.
```
dehora: `E.g indead` -> `e.g., instead`

`cancallations` -> `cancellations`

codekultur: merci :) fixed the typos

tkrop:
👍

ePaul:
:+1:

PR #217, opened 2017-04-05
Guidance on GET with complex query objects (GET with body).

### LINKED ISSUE #207: Give better guidance on when to use GET with headers and query parameters or POST with body regarding size limits.
API designers sometimes face the problem, that they have to provide extensive structured request information with GETs, that may even conflicts with the size limits of clients, load-balancers, and servers. In this situations API designers and developers have to chose between:

1. GET with URL encoded query parameters,
2. GET with header parameters,
3. POST with body content.

Currently, we do not give proper guidance to handle this case, and what are the advantages and disadvantages.

We discussed this in the last guild meeting and came to the following conclusions:

* It feels unnatural for users to provide request parameters via headers.
* Header parameters are itself object to size limitations, thus no *real* solution to the size problem.
* Size limits on query parameters and headers are not reliable at all and depend more on clients, routers, and server as well as settings as on the difference between header and query parameters.
* Query parameters are the preferable choice to express request information, even if the URL encoding limits the readability to a small degree.
* If the structuring of the query information it too extensive, a POST with the request in the content body should be used to transport the query information. The end-point should be marked to behave like a GET (need to work out syntax).

In addition we would like to add an advise that headers are reserved for contextual information that are not part of the data model of the API.

Please comment further suggestions and opinions to this issue.

tkrop:
Fixes #207.

### DIFF
--- a/http/Http.md
+++ b/http/Http.md
@@ -20,8 +20,8 @@ as [pagination](../pagination/Pagination.md).
 
 APIs sometimes face the problem, that they have to provide extensive structured request information
 with GET, that may even conflicts with the size limits of clients, load-balancers, and servers. As
-we require APIs to be standard conform (body in GET must be ignored on server side), developers have
-to check the following two options:
+we require APIs to be standard conform (body in GET must be ignored on server side), API designers
+have to check the following two options:
 
 1. GET with URL encoded query parameters: when it is possible to encode the request information in
    query parameters, respecting the usual size limits of clients, gateways, and servers, this should
--- a/http/Http.md
+++ b/http/Http.md
@@ -27,9 +27,9 @@ to check the following two options:
    query parameters, respecting the usual size limits of clients, gateways, and servers, this should
    be the first choice. The request information can either be provided distributed to multiple query
    parameters or a single structured URL encoded string.
-2. POST with body content: when a GET with URL encoded query parameters ins impossible, a POST
-   with body content must be used. In this case the endpoint must be documented with  the hint
-   `GET with body` to transport the GET semantic of this call.
+2. POST with body content: when a GET with URL encoded query parameters is not possible, a POST with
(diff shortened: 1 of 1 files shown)


http/Http.md
```diff
@@ -16,6 +16,28 @@ GET requests are used to read a single resource or query set of resources.
...
+APIs sometimes face the problem, that they have to provide extensive structured request information
+with GET, that may even conflicts with the size limits of clients, load-balancers, and servers. As
+we require APIs to be standard conform (body in GET must be ignored on server side), developers have
+to check the following two options:
+
+1. GET with URL encoded query parameters: when it is possible to encode the request information in
+   query parameters, respecting the usual size limits of clients, gateways, and servers, this should
+   be the first choice. The request information can either be provided distributed to multiple query
+   parameters or a single structured URL encoded string.
+2. POST with body content: when a GET with URL encoded query parameters ins impossible, a POST
```
hjacobs: typo "is not"

http/Http.md
```diff
@@ -16,6 +16,28 @@ GET requests are used to read a single resource or query set of resources.
 **Note:** GET requests on collection resources should provide a sufficient filter mechanism as well
 as [pagination](../pagination/Pagination.md).
 
+### GET with Body
+
+APIs sometimes face the problem, that they have to provide extensive structured request information
+with GET, that may even conflicts with the size limits of clients, load-balancers, and servers. As
+we require APIs to be standard conform (body in GET must be ignored on server side), developers have
```
ePaul: developers → API designers?

ePaul:
In general, looks fine.

http/Http.md
tkrop: Yes better.

ePaul:
:+1:

tfrauenstein:
👍

PR #222, opened 2017-04-25
Move "required" block of common address data structure

netme:
That should simplify reading and maintain the common address data structure.

📗 

Inspired by @tlossen during internal API review


fmueller:
👍

hjacobs:
:+1:

PR #230, opened 2017-05-09
Elides references to Nakadi in event guidelines.

dehora:
Part of series of edits to abstract out the dependency the guidelines have
on the Nakadi project (guidelines should not have hard dependency on
technology implementations). This reduces the number of Nakadi references
without changing the meaning or intent of the guidelines leaving Nakadi
referenced as an example of a system that supports events. The guidelines
still link to the Open API definition hosted by the Nakadi project - how the
formats evolve and where they live is separate work.

The single structural change is to call out the need for reviews of event  
designs as a separate heading ("{{ book.must }} Make Events available 
for review"). This is making what's already stated in the  guidelines 
clearer.

For #237.

### DIFF
--- a/events/events.md
+++ b/events/events.md
@@ -1,30 +1,40 @@
 # Events
 
-Zalando’s architecture centers around decoupled microservices and in that context we favour asynchronous event driven approaches. The [Nakadi](https://github.com/zalando/nakadi) framework provides an event publish and subscribe system, abstracting exchanges through a RESTful API. 
-
-Nakadi is a key element for data integration in the architecture, and the standard system for inter-service event messaging. The guidelines in this section are for use with the Nakadi framework and focus on how to design and publish events intended to be shared for others to consume. Critically, once events pass service boundaries they are considered part of the service API and are subject to the API guidelines. 
+Zalando’s architecture centers around decoupled microservices and in that context we favour asynchronous event driven approaches. The guidelines in this section focus on how to design and publish events intended to be shared for others to consume. 
 
 ## {{ book.must }} Treat Events as part of the service interface
 
-Events are part of a service’s interface to the outside world equivalent in standing to a service’s REST API. Services publishing data for integration must treat their events as a first class design concern, just as they would an API. For example this means approaching events with the "API first" principle in mind [as described in the Introduction](../Introduction.md) and making them available for review.
+Events are part of a service’s interface to the outside world equivalent in standing to a service’s REST API. Services publishing data for integration must treat their events as a first class design concern, just as they would an API. For example this means approaching events with the "API first" principle in mind [as described in the Introduction](../Introduction.md).
+
+## {{ book.must }} Make Events available for review
+
+Services publishing event data for use by others must make the event schema available for review. 
 
 ## {{ book.must }} Ensure that Events define useful business resources
  
  Events are intended to be used by other services including business process/data analytics and monitoring. They should be based around the resources and business processes you have defined for your service domain and adhere to its natural lifecycle (see also "Should: Define useful resources" in the [General Guidelines](../general-guidelines/GeneralGuidelines.md)).
  
- As there is a cost in creating an explosion of event types and topics, prefer to define event types that are abstract/generic enough to be valuable for multiple use cases, and avoid publishing event types without a clear need.
+As there is a cost in creating an explosion of event types and topics, prefer to define event types that are abstract/generic enough to be valuable for multiple use cases, and avoid publishing event types without a clear need.
 
-## {{ book.must }} Ensure that Event Types conform to Nakadi's API
+## {{ book.must }} Ensure that events conform to Zalando's event types
 
-Nakadi defines a structure called an _EventType_, which describes details for a particular kind of event. The EventType declares standard information, such as a name, an owning application (and by implication, an owning team), a well known event category (business process or data change), and a schema defining the event payload. It also allows the declaration of validation and enrichment strategies for events, along with supplemental information such as how events are partitioned in the stream. 
+Events are defined using a structure called an _EventType_, which describes details for a particular kind of event. The EventType declares standard information, such as a name, an owning application (and by implication, an owning team), a well known event category (such as a business process or data change), and a schema defining the event payload. It also allows the declaration of validation and enrichment strategies for events, along with supplemental information such as how events are partitioned in the stream. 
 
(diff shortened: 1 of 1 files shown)


fmueller:
@tfrauenstein Can you add your comments please?

@dehora Can you resolve the conflicts?

events/events.md
```diff
@@ -1,44 +1,56 @@
...
 ## {{ book.must }} Treat Events as part of the service interface
 
-Events are part of a service’s interface to the outside world equivalent in standing to a service’s REST API. Services publishing data for integration must treat their events as a first class design concern, just as they would an API. For example this means approaching events with the "API first" principle in mind [as described in the Introduction](../Introduction.md) and making them available for review.
+Once events pass service boundaries they are considered part of the service API and are subject to the API guidelines. 
+
+Events are part of a service’s interface to the outside world, equivalent in standing to a service’s REST API. Events must be approached with the "API first" principle in mind [as described in the Introduction](../Introduction.md).
+
+## {{ book.must }} Make Events available for review
+
+Services publishing event data for use by others must make the event schema available for review as they would with API representations. Events made be made available for review as Open API YAML or as JSON Schema.
```
ePaul: `made be made` → `must be made` or `may be made`?

ePaul: Do we have some automatic way of converting an OpenAPI-like YAML to the event type definition processed by Nakadi? (Doesn't need to be in the Guidelines, just out of interest.)

events/events.md
```diff
@@ -1,44 +1,56 @@
...
+The [current Open API descriptions](https://github.com/zalando/nakadi/blob/master/api/nakadi-event-bus-api.yaml) include the definitions for three categories:
 
-The service specific data defined for an event is called the _payload_. Only this non-Nakadi defined part of the event is expected to be submitted with the EventType schema.  When defining an EventType's payload schema as part of your API and for review purposes, it's ok to declare the payload as an object definition using Open API, but please note that Nakadi currently expects the EventType's schema to be submitted as JSON Schema and not as Open API JSON or YAML. Further details on how to register EventTypes are available in the Nakadi project's documentation.
+- BusinessEvent, for business process events.
+- DataChangeEvent for data change events.
+- Undefined, for generic 'undefined' type.
+
+The available types of events may be added to in the future.
```
ePaul: Maybe better: "This list of categories might be extended in the future."

events/events.md
dehora: > Do we have some automatic way of converting an OpenAPI-like YAML to the event type definition processed by Nakadi? 

No. Nakadi understands JSON-Schema and OpenAPI is a breaking superset. Also Open API requires paths but not definitions, ie it's a HTTP API definition language first and a format language second.

ePaul: Then maybe we should not write "as Open API YAML" here? (I think there might be some value for doing the review on the YAML syntax – even if it is just JSON schema and not an OpenAPI schema – instead of JSON, but we would need some automated converter to be able to use that as the type definition.

(When using the same schema for both REST API and event API, it might be useful to use just the intersection of both schema languages.)

tfrauenstein:
👍   Thx. for the contribution. 
It is the 1st step for decoupling of event guidelines from Nakadi.
Next step is to master the event category and format definition -- separate PR.
I also have editorial remarks concerning API First and Call for Review obligation -- however, separate PR.

tfrauenstein:
👍

events/events.md
```diff
@@ -1,44 +1,56 @@
...
 ## {{ book.must }} Treat Events as part of the service interface
 
-Events are part of a service’s interface to the outside world equivalent in standing to a service’s REST API. Services publishing data for integration must treat their events as a first class design concern, just as they would an API. For example this means approaching events with the "API first" principle in mind [as described in the Introduction](../Introduction.md) and making them available for review.
+Once events pass service boundaries they are considered part of the service API and are subject to the API guidelines. 
+
+Events are part of a service’s interface to the outside world, equivalent in standing to a service’s REST API. Events must be approached with the "API first" principle in mind [as described in the Introduction](../Introduction.md).
+
+## {{ book.must }} Make Events available for review
+
+Services publishing event data for use by others must make the event schema available for review as they would with API representations. Events may be made available for review as Open API YAML or as JSON Schema.
```
tkrop: @dehora As we see humans as main stakeholder of API definitions, we have been more strictly lately in requiring APIs to be written in YAML. The before sentence contradicts our expectation as it allows JSON Schema.

events/events.md
```diff
@@ -1,44 +1,56 @@
...
+Zalando’s architecture centers around decoupled microservices and in that context we favour asynchronous event driven approaches. The guidelines in this section focus on how to design and publish events intended to be shared for others to consume. 
 
 ## {{ book.must }} Treat Events as part of the service interface
 
-Events are part of a service’s interface to the outside world equivalent in standing to a service’s REST API. Services publishing data for integration must treat their events as a first class design concern, just as they would an API. For example this means approaching events with the "API first" principle in mind [as described in the Introduction](../Introduction.md) and making them available for review.
+Once events pass service boundaries they are considered part of the service API and are subject to the API guidelines.
```
ePaul: I would read that as `considered to be part`.

events/events.md
ePaul: I don't think it makes much sense to review a schema in YAML format and then have to convert it manually to JSON for Nakadi event type creation.

I would really like to have some automated way of doing this (or have Nakadi support a YAML version of JSON schema), but currently we don't have it.

tkrop: @ePaul Yes, I agree. But still, as we go to elide Nakadi, we should stick to the human best readable form. Until Nakadi offers to import YAML, we could automatically convert it to JSON schema in the Nakadi event schema update script any team has to prepare for roll out.

ePaul: So maybe "or as JSON schema (in YAML form)"?

dehora: > Events may be made available for review as Open API YAML or as JSON Schema.

I'm removing the reference to formats here. Formats discussed elsewhere in the section and I want to keep this pr as far as possible an editorial one rather than open up a debate that should happen somewhere else.

dehora: > we should stick to the human best readable form. Until Nakadi offers to import YAML, we could automatically convert it to JSON schema in the Nakadi event schema update script any team has to prepare for roll out.

That doesn't work as a specification option because events are defined in terms of a data format, whereas OpenAPI requires (and only requires) the path elements (object definitions are optional).

This is a general problem with Open API's design and the way the guidelines mandate use of Open API for object definitions, that is outside the scope of this pr.

tkrop:
👍

tfrauenstein:
👍

Issue #231, opened 2017-05-09
Build doesn't work locally with recent node/npm versions


dehora:
The build doesn't appear to work with recent node/npm versions. With node v7.10.0 and npm v4.5.0, I get the dreaded `Error: Cannot find module 'internal/fs'`. I think I've traced this down to gitbook 2.6.7 which seems to rely on older versions of node and graceful-fs. On the other hand upgrading gitbook to 3 breaks restructure-navigation which seems to be fixed to gitbook 2.x (upgrading all other dependencies to their latest seems to be ok, including grunt). I also  tried the restructure-navigation3 version in npm and had no luck. It looks like [restructure-navigation](https://github.com/zalando-incubator/gitbook-restructure-navigation) is something Zalando built for itself (and perhaps for these guidelines, not sure). I haven't tried a workaround with moving the global npm/node versions back yet, but that's suboptimal for contributors.

This isn't breaking builds for publishing, but gitbook 3's been out for a while and staying on old versions in the node/npm ecosystem is usually unwise.

fmueller:
We once decided to stick with old Gitbook version https://github.com/zalando/restful-api-guidelines/issues/106

I think, if the guild decides to go with Gitbook 3 the guidelines need to be reordered to fit in default gitbook chapter structure. I would not adapt custom plugins or implement another plugin. Just stick to the default.

fmueller:
And current build is working with Node 5.1.1 Can we at least hardwire this somewhere?

dehora:
> We once decided to stick with old Gitbook version #106

Got it. I don't know how that stays realistic for (#106 is nearly a year old) in this particular ecosystem. It's disappointing the piece holding back an upgrade is an unmaintained Zalando plugin.

maxim-tschumak:
@fmueller We can specify the [engine version](https://docs.npmjs.com/files/package.json#engines) to the package.json.

ePaul:
**From the meeting:** We should mention our decision to stay at this old Gitbook version in BUILD.md (maybe inviting volunteers to work on improving this). @fmueller could you do this?

fmueller:
Yes @ePaul, will put this into BUILD.md and mention it in README accordingly. Good idea.

Issue #226, opened 2017-05-09
Use rfc2119 keywords to indicate guidance levels


dehora:
The guidelines mandate guideline levels without explaining what they mean. It's customary to use RFC2119 in such cases as they're widely understood in the industry. MSFT and GOOG guidelines both use RFC2119. There doesn't seem to be much value in having  undefined terminology.

This can be added to the introduction:

>Conventions Used in These Guidelines
>
>The requirement level keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" used in this document (case insensitive) are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).

With the following mapping:

| Current term | RFC2119 |
| ------------ | ------------- | 
| Must |  Must | 
| Should | Should |
| Could | May |  

The changes in the text are minimal (`could` is replaced by `may`). Also the `required` term allows the guidelines to indicate and discuss mandates that come from aspects such as rules of play.

tfrauenstein:
👍

ePaul:
From the meeting: We agree this is a good idea.
When changing the word could to may, we need to make sure the build process (Gitbook plugin for highlighting) still works.

@dehora Do you want to work on this?

dehora:
Ok, see https://github.com/zalando/restful-api-guidelines/pull/232

PR #232, opened 2017-05-10
Uses RFC2119 keywords to indicate guidance levels.

dehora:
This adds a section describing the use of RFC2119 to describe the
requirements levels for each guideline. The naming changes to the
guideline headings are:

 - book.must: stays the same
 - book.should: stays the same
 - book.could: becomes book.may

For #226.

Example of the new rendering output:

screen shot 2017-05-10 at 10 11 08

Because of https://github.com/zalando/restful-api-guidelines/issues/231 that was generated using a patched build process. It's probably worth someone else verifying locally.

As a data point for the curious, the guideline levels are distributed as a follows

- must: 62
- should: 33
- could: 14

### DIFF
--- a/Introduction.md
+++ b/Introduction.md
@@ -25,6 +25,10 @@ We encourage our teams to follow them to ensure that our APIs:
 
 Ideally, all Zalando APIs will look like the same author created them.
 
+## Conventions Used in These Guidelines
+
+The requirement level keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" used in this document (case insensitive) are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
+
 ## Zalando specific information
 
 The purpose of our “RESTful API guidelines” is to define standards to successfully establish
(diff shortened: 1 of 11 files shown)


Introduction.md
```diff
@@ -25,6 +25,10 @@ We encourage our teams to follow them to ensure that our APIs:
 
 Ideally, all Zalando APIs will look like the same author created them.
 
+## Conventions Used in These Guidelines
+
+The requirement level keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" used in this document (case insensitive) are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
```
fmueller: Out of curiosity, why do you list keywords which we don't use in our guidelines?

fmueller:
👍

Introduction.md
dehora: > why do you list keywords which we don't use in our guidelines?

1: Because you'd have to explain why there's a subset

2: The guidelines already use them. Ignoring negations, the 3 keywords used in titles, and conversational text (especially for the word required):

- shall: 10
- required: ~6
- optional: 4
- recommended: 5

ePaul:
:+1:

ePaul:
I didn't test this locally to see how it looks (no npm installed). @fmueller did you?

PR #239, opened 2017-05-29
Add definition on how to use the Money object.

Tar-Minyatur:
In several services I encountered a wrong usage of the Money object, where the `amount` was actually a Cent value that needed to be converted into a Euro value, before it could be used. That is the wrong way to do it and can create very confusing situations. This PR adds a small paragraph and some examples about the correct usage.

### DIFF
--- a/common-data-objects/CommonDataObjects.md
+++ b/common-data-objects/CommonDataObjects.md
@@ -11,19 +11,26 @@ Use the following common money structure:
       properties:
         amount:
           type: number
+          description: Amount expressed as a decimal number of major currency units
           format: decimal
           example: 99.95
         currency:
           type: string
+          description: 3 letter currency code as defined by ISO-4217
           format: iso-4217
           example: EUR
       required:
         - amount
         - currency
 
-The `amount` field must not assume that consumers apply any kind of conversion, e.g. division by 100 to get Euros from a Cent value. Decimal values can represent unit and subunit at once, so the only correct representation for 42 Euros and 23 Cents is:
+The decimal values for "amount" describe unit and subunit of the currency in a single value, where
+the digits before the decimal point are for the major unit and the digits after the decimal point are
+for the minor unit. Examples for correct representations (in EUR):
 
-    { "amount": "42.23", "currency": "EUR" }
+- `42.23` = 42 Euros, 23 Cent
+- `0.23` = 23 Cent
+- `42.0` = 42 Euros
+- `1024.42` = 1024 Euros, 42 Cent
 
 Make sure that you don’t convert the “amount” field to `float` / `double` types when implementing
 this interface in a specific language or when doing calculations. Otherwise, you might lose
--- a/common-data-objects/CommonDataObjects.md
+++ b/common-data-objects/CommonDataObjects.md
@@ -25,12 +25,16 @@ Use the following common money structure:
 
 The decimal values for "amount" describe unit and subunit of the currency in a single value, where
 the digits before the decimal point are for the major unit and the digits after the decimal point are
-for the minor unit. Examples for correct representations (in EUR):
+for the minor unit. Note that some business cases (e.g. transactions in Bitcoin) call for a higher
+precision, so applications must be prepared to accept values with unlimited precision, unless
+explicitly stated otherwise in the API specification.
+Examples for correct representations (in EUR):
 
-- `42.23` = 42 Euros, 23 Cent
+- `42.20` or `42.2` = 42 Euros, 20 Cent
 - `0.23` = 23 Cent
-- `42.0` = 42 Euros
+- `42.0` or `42` = 42 Euros
 - `1024.42` = 1024 Euros, 42 Cent
+- `1024.4225` = 1024 Euros, 42.25 Cent
 
 Make sure that you don’t convert the “amount” field to `float` / `double` types when implementing
 this interface in a specific language or when doing calculations. Otherwise, you might lose
--- a/common-data-objects/CommonDataObjects.md
+++ b/common-data-objects/CommonDataObjects.md
@@ -21,6 +21,10 @@ Use the following common money structure:
         - amount
         - currency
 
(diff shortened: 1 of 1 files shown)


common-data-objects/CommonDataObjects.md
```diff
@@ -21,6 +21,10 @@ Use the following common money structure:
         - amount
         - currency
 
+The `amount` field must not assume that consumers apply any kind of conversion, e.g. division by 100 to get Euros from a Cent value. Decimal values can represent unit and subunit at once, so the only correct representation for 42 Euros and 23 Cents is:
```
whiskeysierra: I'd propose to keep this a bit more generic, e.g. 

> The amount is expressed as a decimal in major units.

I'd also suggest that we add proper descriptions to the individual fields in the schema and that we in fact expose it as its own schema file with a stable URL, as we do with problems. That way API designers don't have to copy it all the time.

ePaul: I think the example (with EUR and cents) adds value. Maybe just add the quoted sentence at the start of this paragraph?

ePaul: About putting the schema in its own file, this could be done in a separate pull request.

dehora: > Decimal values can represent unit and subunit at once, so the only correct representation for 42 Euros and 23 Cents is:

I don't think I get the intent of this sentence. The opening clause says "can represent unit and subunit at once", but the following example says there's only one "correct representation" and the sentence says that's a consequence by bridging them with "so" instead of "however"

If the goal is to define how to use the money as per the pr title. The pr needs to be be rewritten to start with specifying there's only one correct representation for this common object, and provide examples of correct representations that cover the main scenarios: 

  - exact euro unit level amounts with nothing in the cent subunit range
  - cent amounts less that 1 euro
  - euro amounts greater than 1000 euros with unit and subunit
  - the representation of deductions (eg whether a client should ever expect a negative sign)

common-data-objects/CommonDataObjects.md
```diff
@@ -21,6 +21,10 @@ Use the following common money structure:
...
+
+    { "amount": "42.23", "currency": "EUR" }
```
dehora: >  { "amount": "42.23", "currency": "EUR" }

This is incorrect, the amount field is a number, not a string.

common-data-objects/CommonDataObjects.md
tkrop: I like @whiskeysierra suggestion of adding descriptions to the schema fields instead of adding a sentence below. The example is already in the schema definition. It should something as follows:

> amount: Amount expressed as a decimal number of major currency units.
> currency: currency name of 3 letters as defined by ISO-4217.

Tar-Minyatur:
It's nice to see that this sparked a discussion. :) I tried to incorporate your suggestions. While I like the idea of having a central location for the Money object definition, I agree that this should be done separately, because it does not help with the problem I want to address.

common-data-objects/CommonDataObjects.md
ePaul: I think the point is that for 42 EUR + 23 cent the `amount` should be `42.23`, not e.g. `4223`.
Of course we can blow this up with many other test cases, but I think this is not necessary.

common-data-objects/CommonDataObjects.md
```diff
@@ -11,16 +11,27 @@ Use the following common money structure:
...
         - amount
         - currency
 
+The decimal values for "amount" describe unit and subunit of the currency in a single value, where
+the digits before the decimal point are for the major unit and the digits after the decimal point are
+for the minor unit. Examples for correct representations (in EUR):
+
+- `42.23` = 42 Euros, 23 Cent
+- `0.23` = 23 Cent
+- `42.0` = 42 Euros
```
ePaul: For completeness, we could add

- `42` = 42 Euros

here (as this is an equivalent alternative way of writing this).

common-data-objects/CommonDataObjects.md
```diff
@@ -11,16 +11,27 @@ Use the following common money structure:
...
+          description: 3 letter currency code as defined by ISO-4217
           format: iso-4217
           example: EUR
       required:
         - amount
         - currency
 
+The decimal values for "amount" describe unit and subunit of the currency in a single value, where
+the digits before the decimal point are for the major unit and the digits after the decimal point are
+for the minor unit. Examples for correct representations (in EUR):
```
ePaul: Some applications need more precision than two digits (e.g. stock valuation at Zalando is currently working in millicent, i.e. 10<sup>-5</sup> EUR. Bitcoins are used in multiples of 0.00000001 BTC.

The examples are all suggesting that there are always only two digits. Should we add some guidance about that? (Like "when you need more than two digits after the decimal point, indicate this in the API definition in a description")?

whiskeysierra: I'd most definitely allow for this. I also like that by default the scale is equal to the default defined by the currency. Anybody deviating from this should document it properly.

whiskeysierra:
I feel like this could be helpful for some discussions around the topic in general: https://github.com/zalando/jackson-datatype-money/blob/master/MONEY.md

common-data-objects/CommonDataObjects.md
tkrop: We could improve the amount description:

> Amount expressed as a decimal number of major currency units. The digits after the decimal point are expressing the minor currency units with potentially infinite precision, e.g. for bitcoins (BTC).

or we could add before the example:

> Client and service implementers should be prepared to support infinite precision values, e.g. for BTC.

Anyhow, I still think we overdoing the additional description and example part here, as the connection JSON and decimal number in my opinion implies infinite precision numbers - doubles and floats are no option for parsing unknown JSON correctly. May be we should add a rule to our JSON guideline section:

> Must: parse decimal numbers with infinite precision.
>
> API implementers must be prepared to parse numbers with infinite precision integer and decimal values unless stated otherwise by the schema description.

ePaul: @tkrop infinite → unlimited. Infinite precision sounds like impossible.

tkrop:
👍

maxim-tschumak:
:+1:

PR #241, opened 2017-06-06
Issue #227: Mark appendixes as such

ePaul:
This is a part of #227. (The other part is #242.)

I also added some change log entries for other recently merged PRs.

Issue #227, opened 2017-05-09
Clarify compatibility status of Open API definitions and generated code


dehora:
It's possible to use multiple Open API definitions to define the same JSON structure. 

The guidelines talk about compatibility in terms of the JSON at "runtime" and not about compatibility of object definitions. The guidelines also say code generation is fine. From [here](http://zalando.github.io/restful-api-guidelines/tooling/Tooling.html):

> The following frameworks were specifically designed to support the API First workflow with OpenAPI YAML files
>...
> - Swagger Codegen: template-driven engine to generate client code in different languages by parsing Swagger Resource Declaration
> - Swagger Codegen Tooling: plugin for Maven that generates pieces of code from OpenAPI specification

Nothing is said about the compatibility of the definition itself with respect to generated code. Changing a definition file such that client code breaks but the runtime output is the same might lead to confusion around what a compatibility break is.

It could be argued that the emphasis on the REST plus API-first principles, and the compatibility section would make it obvious that as long as the runtime JSON doesn't break, Open API changes are compatible. I'm not sure this is sufficient because it's implicit.

ePaul:
From the meeting:
* We agree that the compatibility guarantees are just for the "on-the-wire" (JSON) format described by OpenAPI, in the sense that clients following the old API definitions still work with the new API. We don't agree source or byte code or similar compatibility for anything generated from it – updating a client to a new API definition version would usually imply changing some of the using code (if you are using a generator at all).
* The whole Tooling section is not really part of the guidelines, more some kind of appendix. We might need to think about how to make this clearer.

ePaul:
I opened pull requests #141 + #142 for this.

ePaul:
@dehora could you have a look at the pull requests?

ePaul:
With both PRs being merged, I think this can be closed. @dehora please reopen if you think I missed something.

dehora:
> With both PRs being merged, I think this can be closed.

👍  Nice work!

PR #242, opened 2017-06-06
Issue #227: make clear that compatibility is just for on-the-wire format

ePaul:
This is the second half of #227 (the first one is #241).

### DIFF
--- a/compatibility/Compatibility.md
+++ b/compatibility/Compatibility.md
@@ -18,12 +18,14 @@ allow us to make compatible changes without versioning.
 
 Please note that the compatibility guarantees are for the on-the-wire format,
 i.e. they ensure that consumers relying on the old API definition will work
-with servers which already implement a newer API definitions, assuming both API
+with servers which already implement a newer API definition, assuming both API
 designers and client and server implementors adhere to the following rules.
 
-(Binary or source compatibility of code generated from an API definition is not covered
-by this – if you update the definition to a new version, it is expected that code changes
-are necessary.)
+**Note:** Binary or source compatibility of code generated from an API definition
+is not covered by these rules.  If client implementations update their generation
+process to a new version of the API definition, it has to be expected that code
+changes are necessary.
+
 
 ## {{ book.should }} Prefer Compatible Extensions
 
--- a/compatibility/Compatibility.md
+++ b/compatibility/Compatibility.md
@@ -13,8 +13,17 @@ There are two techniques to change APIs without breaking them:
 - introduce new API versions and still support older versions
 
 We strongly encourage using compatible API extensions and discourage versioning.
-With Postel’s Law in mind, here are some rules for providers and consumers that
-allow us to make compatible changes without versioning:
+With Postel’s Law in mind, the following some rules for providers and consumers
+allow us to make compatible changes without versioning.
+
+Please note that the compatibility guarantees are for the on-the-wire format,
+i.e. they ensure that consumers relying on the old API definition will work
+with servers which already implement a newer API definitions, assuming both API
+designers and client and server implementors follow the following rules.
+
+(Binary or source compatibility of code generated from an API definition is not covered
+by this – if you update the definition to a new version, it is expected that code changes
+are necessary.)
 
 ## {{ book.should }} Prefer Compatible Extensions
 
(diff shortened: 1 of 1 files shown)


compatibility/Compatibility.md
```diff
@@ -13,8 +13,17 @@ There are two techniques to change APIs without breaking them:
 - introduce new API versions and still support older versions
 
 We strongly encourage using compatible API extensions and discourage versioning.
-With Postel’s Law in mind, here are some rules for providers and consumers that
-allow us to make compatible changes without versioning:
+With Postel’s Law in mind, the following some rules for providers and consumers
```
whiskeysierra: > the following rules

compatibility/Compatibility.md
```diff
@@ -13,8 +13,17 @@ There are two techniques to change APIs without breaking them:
...
+allow us to make compatible changes without versioning.
+
+Please note that the compatibility guarantees are for the on-the-wire format,
+i.e. they ensure that consumers relying on the old API definition will work
+with servers which already implement a newer API definitions, assuming both API
+designers and client and server implementors follow the following rules.
```
whiskeysierra: > adhere to the following rules?

compatibility/Compatibility.md
ePaul: Hmm, I wanted to explain that I mean just some rules, not all of them. Though I guess then I should have gone and counted (`the following 5 rules`).

compatibility/Compatibility.md
ePaul: Updated.

compatibility/Compatibility.md
```diff
@@ -13,8 +13,17 @@ There are two techniques to change APIs without breaking them:
...
-allow us to make compatible changes without versioning:
+With Postel’s Law in mind, the following five rules for providers and consumers
+allow us to make compatible changes without versioning.
+
+Please note that the compatibility guarantees are for the on-the-wire format,
+i.e. they ensure that consumers relying on the old API definition will work
+with servers which already implement a newer API definitions, assuming both API
+designers and client and server implementors adhere to the following rules.
+
+(Binary or source compatibility of code generated from an API definition is not covered
```
tkrop: I would express this less reluctant:

**Note:** compatibility of code generated from an API definition is not covered by the rules. If you update the definition, it has to be expected that code changes are necessary.

compatibility/Compatibility.md
```diff
@@ -13,8 +13,17 @@ There are two techniques to change APIs without breaking them:
...
 We strongly encourage using compatible API extensions and discourage versioning.
-With Postel’s Law in mind, here are some rules for providers and consumers that
-allow us to make compatible changes without versioning:
+With Postel’s Law in mind, the following five rules for providers and consumers
+allow us to make compatible changes without versioning.
+
+Please note that the compatibility guarantees are for the on-the-wire format,
+i.e. they ensure that consumers relying on the old API definition will work
+with servers which already implement a newer API definitions, assuming both API
+designers and client and server implementors adhere to the following rules.
```
dehora: > +with servers which already implement a newer API definitions,

Can remove the plural here - 

"""
with servers which already implement a newer API definition, 
"""

compatibility/Compatibility.md
```diff
@@ -13,8 +13,17 @@ There are two techniques to change APIs without breaking them:
...
+
+(Binary or source compatibility of code generated from an API definition is not covered
+by this – if you update the definition to a new version, it is expected that code changes
+are necessary.)
```
dehora: I don't think the parens here are needed, it's a constraint on API clients and can be phrased as a specification - 

"""
Binary or source compatibility of code generated from an API definition is not covered by this – API designers MAY update their Open API definition to a new version such that code changes by client generators would be necessary, as long as the runtime representations remain compatible.
"""

compatibility/Compatibility.md
ePaul: Done.

compatibility/Compatibility.md
ePaul: Done (slightly adapted).

compatibility/Compatibility.md
ePaul: I think even more that API designers should not care about what client code generators are doing with this, as long as they adhere to the rules here.

I removed the parentheses, and reworded it a bit, mostly following the proposal of @tkrop.

ePaul:
:+1:

tkrop:
👍

PR #244, opened 2017-06-14
Remove reference to json-fields library

### LINKED ISSUE #224: Remove library level guidance dependency on json-fields
From the section [Performance.md#-bookshould--support-filtering-of-resource-fields](https://github.com/zalando/restful-api-guidelines/blob/master/performance/Performance.md#-bookshould--support-filtering-of-resource-fields)

> The approach we recommend for field filtering is a Zalando Github project, json-fields. It defines a formal grammar for the ANTLR parser generator and provides a ready-to use library for Java / Jackson based projects (Maven link). Teams that use other JSON serializers are encouraged to contribute to the open source project and create their own parser / framework based on this grammar.

This puts a dependency on a library into the guidelines. That seems to be contrary to API-first. It's not clear how widely adopted the standard or library [1] is even within Zalando or that it's been known this design is effective and successfully adopted. The concern is that this approach was  standardized ahead of sufficient experience.

In any case, the guidelines should specify in a way that doesn't require a library and a grammar file pinned to a particular runtime. 

----

[1] The library itself has problems, it introduces a runtime dependency on guava (and antlr), but that's best filed upstream on the library itself.

duergner:
Instead of referencing the external [json-fields](https://github.com/zalando-incubator/json-fields) library the section about filtering fields on the response for performance reasons now defines the desired format as a BNF grammar inside the section.

fixes #224

### DIFF
--- a/performance/Performance.md
+++ b/performance/Performance.md
@@ -70,17 +70,34 @@ Content-Type: application/json
 }
 ```
 
-The approach we recommend for field filtering is a Zalando Github project,
-[json-fields](https://github.com/zalando/json-fields). It defines a formal grammar for the ANTLR
- parser generator and provides a ready-to use library for Java / Jackson based projects
- ([Maven link](http://mvnrepository.com/artifact/org.zalando.guild.api/json-fields-jackson)).
-Teams that use other JSON serializers are encouraged to contribute to the open source project and
-create their own parser / framework based on this grammar.
+The approach we recommend for field filtering is adding a request parameter like e.g. `fields` 
+which value follows the following [BNF](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form) grammar.
 
-Other approaches we have considered are JSONPath or GraphQL. While they have advantages, neither of
-them can easily be plugged into an existing serialization process, so they require an additional,
-manual serialization process, whereas the above solution addresses our main filter use cases and
-can easily be introduced with a minimum of effort.
+```
+<fields> ::= <negation> <fields_expression> | <fields_expression>
+
+<negation> ::= "!"
+
+<fields_expression> ::= "(" <field_set> ")"
+
+<field_set> ::= <qualified_field> | <qualified_field> "," <field_set>
+
+<qualified_field> ::= <field> | <field> <fields_expression>
+
+<field> ::= <DASH_LETTER_DIGIT> | <DASH_LETTER_DIGIT> <field>
+
+<DASH_LETTER_DIGIT> ::= <DASH> | <LETTER> | <DIGIT>
+
+<DASH> ::= "-" | "_"
+
+<LETTER> ::= "A" | "B" | "C" | "D" | "E" | "F" | "G" | "H" | "I" | "J" | "K" | "L" | "M" | "N" | "O" | "P" | "Q" | "R" | "S" | "T" | "U" | "V" | "W" | "X" | "Y" | "Z" | "a" | "b" | "c" | "d" | "e" | "f" | "g" | "h" | "i" | "j" | "k" | "l" | "m" | "n" | "o" | "p" | "q" | "r" | "s" | "t" | "u" | "v" | "w" | "x" | "y" | "z"
+
+<DIGIT> ::= "0" | "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8" | "9"
+```
+
+A `fields_expression` as defined by the grammar describes the properties of an object, i.e. `(name)` would define to
+return only the `name` property of the root object. `(name,partner(name))` would define to return the `name` and 
+`partner` properties where `partner` itself would also be an object and only its `name` property should be returned.
 
 Hint: OpenAPI doesn't allow you to formally specify whether depending on a given parameter will
 return different parts of the specified result schema. Explain this in English in the parameter
@@ -93,7 +110,9 @@ cases where clients know upfront that they need some related resources they can 
 data eagerly. Whether this is optimized on the server, e.g. a database join, or done in a generic way, e.g. an HTTP
 proxy that transparently embeds resources, is up to the implementation.
 
-See [*Conventional Query Strings*](../naming/Naming.md#could-use-conventional-query-strings) for naming. Please use [json-fields](https://github.com/zalando/json-fields) library, already mentioned above for filtering, when it comes to an embedding query syntax.
+See [*Conventional Query Strings*](../naming/Naming.md#could-use-conventional-query-strings) for naming. Please use the 
+[BNF](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form) grammar, as already defined above for filtering, when it 
+comes to an embedding query syntax.
 
 Embedding a sub-resource can possibly look like this where an order resource has its order items as sub-resource (/order/{orderId}/items):
 
--- a/performance/Performance.md
+++ b/performance/Performance.md
@@ -70,8 +70,8 @@ Content-Type: application/json
 }
 ```
 
-The approach we recommend for field filtering is adding a request parameter like e.g. `fields` 
-which value follows the following [BNF](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form) grammar.
+As illustrated by this example, field filtering should be done via request parameter "fields" with value range defined 
+by the following [BNF](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form) grammar.
 
 ```
 <fields> ::= <negation> <fields_expression> | <fields_expression>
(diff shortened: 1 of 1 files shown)


performance/Performance.md
```diff
@@ -70,17 +70,34 @@ Content-Type: application/json
...
- parser generator and provides a ready-to use library for Java / Jackson based projects
- ([Maven link](http://mvnrepository.com/artifact/org.zalando.guild.api/json-fields-jackson)).
-Teams that use other JSON serializers are encouraged to contribute to the open source project and
-create their own parser / framework based on this grammar.
-
-Other approaches we have considered are JSONPath or GraphQL. While they have advantages, neither of
-them can easily be plugged into an existing serialization process, so they require an additional,
-manual serialization process, whereas the above solution addresses our main filter use cases and
-can easily be introduced with a minimum of effort.
+The approach we recommend for field filtering is adding a request parameter like e.g. `fields`
```
tfrauenstein: editorial: "As illustrated by this example, field filtering should be done via request parameter "fields" with value range defined by the following..."

performance/Performance.md
```diff
@@ -93,7 +110,9 @@ cases where clients know upfront that they need some related resources they can
 data eagerly. Whether this is optimized on the server, e.g. a database join, or done in a generic way, e.g. an HTTP
 proxy that transparently embeds resources, is up to the implementation.
 
-See [*Conventional Query Strings*](../naming/Naming.md#could-use-conventional-query-strings) for naming. Please use [json-fields](https://github.com/zalando/json-fields) library, already mentioned above for filtering, when it comes to an embedding query syntax.
+See [*Conventional Query Strings*](../naming/Naming.md#could-use-conventional-query-strings) for naming. Please use the
```
tfrauenstein: editorial:  "Conventional Query Strings" --> "Conventional Query Parameters" 
editorial:  "for naming." --> "for naming, e.g. "embed" for steering of embedded resource expansion."

performance/Performance.md
```diff
@@ -70,17 +70,34 @@ Content-Type: application/json
...
+<DASH> ::= "-" | "_"
+
+<LETTER> ::= "A" | "B" | "C" | "D" | "E" | "F" | "G" | "H" | "I" | "J" | "K" | "L" | "M" | "N" | "O" | "P" | "Q" | "R" | "S" | "T" | "U" | "V" | "W" | "X" | "Y" | "Z" | "a" | "b" | "c" | "d" | "e" | "f" | "g" | "h" | "i" | "j" | "k" | "l" | "m" | "n" | "o" | "p" | "q" | "r" | "s" | "t" | "u" | "v" | "w" | "x" | "y" | "z"
+
+<DIGIT> ::= "0" | "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8" | "9"
+```
+
+A `fields_expression` as defined by the grammar describes the properties of an object, i.e. `(name)` would define to
+return only the `name` property of the root object. `(name,partner(name))` would define to return the `name` and 
+`partner` properties where `partner` itself would also be an object and only its `name` property should be returned.
```
tfrauenstein: we should add: 
"Hint: You may use json-field/REF/ open source library as a Java parser for the field filter syntax."

performance/Performance.md
tkrop: I think, we should require it to be named `fields`. May be we should add a `fields` parameter template with an expected description that can be copy&pasted by API designers. May be we can refer in the example to the guideline for the details of the supported syntax.

performance/Performance.md
```diff
@@ -70,17 +70,34 @@ Content-Type: application/json
...
-
-Other approaches we have considered are JSONPath or GraphQL. While they have advantages, neither of
-them can easily be plugged into an existing serialization process, so they require an additional,
-manual serialization process, whereas the above solution addresses our main filter use cases and
-can easily be introduced with a minimum of effort.
+The approach we recommend for field filtering is adding a request parameter like e.g. `fields` 
+which value follows the following [BNF](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form) grammar.
+
+```
+<fields> ::= <negation> <fields_expression> | <fields_expression>
```
tkrop: Do we really want/need to enforce brackets? In my opinion `<field_set>` would be sufficient for most use cases:

```
<fields> ::= <negation> <fields_expression> | <fields_expression> | <field_set>
```

performance/Performance.md
duergner: No objections from my side here for both proposals.

Can we maybe host the definition for that parameter in a way to be directly referenced? Similar to the reference to Problem JSON?

performance/Performance.md
duergner: Enforcing the brackets makes it easier to parse IMHO for child entities. But I wouldn't object removing the outer brackets requirement.

performance/Performance.md
tkrop: I'm not absolutely sure, but I assume

```
parameters:
  Fields:
    $ref https://zalando.github.io/common-definitions/schema.yaml#/parameters/Fields
```

would work (maybe @ePaul can say for sure). Anyhow, I'm not sure whether we want to refer for common query parameters and definitions like this. It would make the API reading more difficult for new developers. On the other hand we could put more context information and experienced readers could just skip the details.

performance/Performance.md
tkrop: How did you outdate the discussion.

duergner: I don't have any idea - was not intentional at all

performance/Performance.md
duergner: Agree on the up- and downsides for both new as well as experienced developers. Given that people are normally quite good at pattern matching I feel the downside for new developers could be acceptable.

performance/Performance.md
```diff
@@ -70,17 +70,34 @@ Content-Type: application/json
...
- ([Maven link](http://mvnrepository.com/artifact/org.zalando.guild.api/json-fields-jackson)).
-Teams that use other JSON serializers are encouraged to contribute to the open source project and
-create their own parser / framework based on this grammar.
-
-Other approaches we have considered are JSONPath or GraphQL. While they have advantages, neither of
-them can easily be plugged into an existing serialization process, so they require an additional,
-manual serialization process, whereas the above solution addresses our main filter use cases and
-can easily be introduced with a minimum of effort.
+As illustrated by this example, field filtering should be done via request parameter "fields" with value range defined 
+by the following [BNF](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form) grammar.
```
dehora: > ` [BNF](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form)`

is there a reason this doesn't reference an RFC?

performance/Performance.md
```diff
@@ -70,17 +70,34 @@ Content-Type: application/json
...
+<DASH> ::= "-" | "_"
+
+<LETTER> ::= "A" | "B" | "C" | "D" | "E" | "F" | "G" | "H" | "I" | "J" | "K" | "L" | "M" | "N" | "O" | "P" | "Q" | "R" | "S" | "T" | "U" | "V" | "W" | "X" | "Y" | "Z" | "a" | "b" | "c" | "d" | "e" | "f" | "g" | "h" | "i" | "j" | "k" | "l" | "m" | "n" | "o" | "p" | "q" | "r" | "s" | "t" | "u" | "v" | "w" | "x" | "y" | "z"
+
+<DIGIT> ::= "0" | "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8" | "9"
+```
+
+A `fields_expression` as defined by the grammar describes the properties of an object, i.e. `(name)` would define to
+return only the `name` property of the root object. `(name,partner(name))` would define to return the `name` and 
+`partner` properties where `partner` itself would also be an object and only its `name` property should be returned.
```
dehora: > +A `fields_expression` as defined by the grammar describes the properties of an object, i.e. `(name)` would define to
 +return only the `name` property of the root object. `(name,partner(name))` would define to return the `name` and 
 +`partner` properties where `partner` itself would also be an object and only its `name` property should be returned.

The conditional (would/should) phrasing can be dropped:

"""
A `fields_expression` as defined by the grammar describes the properties of an object, i.e. `(name)` returns only the `name` property of the root object. The `(name,partner(name))` expression returns the `name` and  `partner` properties, where `partner` itself is also an object and only its `name` property is returned.
"""

performance/Performance.md
duergner: @dehora no specific reason. I'm open to change it to a RFC reference if you have in mind?

dehora:
👍

ePaul:
:+1:

duergner:
I'm open to do a follow-up PR with changing the reference to the RFC.

I have a stash locally with an idea for that standard reference. Didn't add it as there's been no additional request / support for it. Shall we put that next weeks guild meeting agenda? Might make sense to look for additional similar use cases as well (I think the common money object might be a similar candidate, right?)

Issue #199, opened 2017-03-06
Make guideline for links more liberal


fmueller:
Currently we define in our guidelines to place all links in `_links` field which also has to be an array. This adds quite a lot of boilerplate to json payloads. And we separate metadata like links from actual context they are part of. We do this - I guess - for more generic client implementations. On the other hand we don't believe in generic clients which is one of our arguments against HATEOAS.

I propose to be more liberal with links:
- stick to Link object in our guidelines
- links can be placed anywhere, at best in their context

dehora:
> Currently we define in our guidelines to place all links in _links field which also has to be an array.

According to [Must: Use a well-defined subset of HAL](http://zalando.github.io/restful-api-guidelines/hyper-media/Hypermedia.html#must-use-a-welldefined-subset-of-hal) and the examples there,

- it's an object: 

```yaml
Links:
  type: object
```

- links can (must) live anywhere:

> Clients and Servers are required to support _links with its href and rel attributes, not only at the root level but also in nested objects.


It sounds like you're asking for fields that can have link objects as values. What's wrong with using `_links` at any level? I'm not convinced about HAL but the guild is (see #147) and allowing links anywhere seems to pretty much defeat the point of using it.

fmueller:
Sorry for misreading this. Right, _links is an object. What's the benefit of wrapping links always inside _links? I only see the 'generic client' argument here and this should not be used by us for this since we used it in past to discourage HATEOAS. You see my point?

Btw: We had a big talk in last API guild meetup about some cons of HAL.

ePaul:
> On the other hand we don't believe in generic clients which is one of our arguments against HATEOAS.

Just for the record, I think a schema-aware semi-generic client (like what was presented in the last meeting – sorry that we can't share the details publicly) could use links in a HATEOASy way even if they are not in a `_links` object. (I'll have to work out whether this can be done with OpenAPI 3.0's new `links` feature or if it needs an extension.)

> What's the benefit of wrapping links always inside _links?

Our current use of `_links` already deviates from HAL (`_links` is allowed just at top level for HAL, but also in nested objects for us), just sharing parts of the syntax with it. So it allows having a link nested in the correct part of your response (or less often request), but with some syntactic overhead.

The `_links` object has the advantage of making the link relation type explicit, instead of hiding it in a description of a property.

tkrop:
The Ideas are not that difficult to share - e.g. Nokias Here Places API is public:

[...]

PR #246, opened 2017-06-15
First suggestion for issue #199: Replace HAL enforcement with simple hypertext control

meshcalero:
A separate pull request for "self-describing hypertext controls" will follow.

### DIFF
--- a/hyper-media/Hypermedia.md
+++ b/hyper-media/Hypermedia.md
@@ -36,19 +36,19 @@ When designing an API that contains any hypertext controls, the attribute name `
 
 The schema for hypertext controls can be derived from this model:
 
-    HttpLink:
-      description: |
-        A base type of objects representing links to resources.
-      type: object
-      properties:
-        href:
-          description: |
-            Any URI that is using http or https protocol
-          type: string
-          format: uri
-      required: [ "href" ]
- 
-The name of an attribute holding such a HttpLink object specifies the relation between the object that 
+```yaml
+HttpLink:
+  description: A base type of objects representing links to resources.
+  type: object
+  properties:
+    href:
+      description: Any URI that is using http or https protocol
+      type: string
+      format: uri
+  required: [ "href" ]
+```
+
+The name of an attribute holding such a `HttpLink` object specifies the relation between the object that 
 contains the link and the linked resource. Implementations should use names from the 
 [IANA Link Relation Registry](http://www.iana.org/assignments/link-relations/link-relations.xhtml) 
 whenever appropriate.
@@ -56,18 +56,21 @@ whenever appropriate.
 Specific link objects may extend the basic link type with additional attributes, to give additional information
 related to the linked resource or the relationship between the source resource and the linked one.
 
-E.g. when modeling a person that is married with some other person you could have attributes in the hypertext control that describe both, the spouse and the relationship between the two persons:
+E.g. a service providing "Person" resources could model a person who is married with some other person with a hypertext control that contains attributes which describe the other person (`id`, `name`) but also the relationship "spouse" between between the two persons (`since`):
 
-    {
-      "id": "446f9876-e89b-12d3-a456-426655440000",
-      "name": "Peter Mustermann",
-      "spouse": {
-        "href": "https://...",
-        "since": "1996-12-19",
-        "id": "123e4567-e89b-12d3-a456-426655440000",
-        "name": "Linda Mustermann"
-      }
-    }
+
+```json
+{
+  "id": "446f9876-e89b-12d3-a456-426655440000",
+  "name": "Peter Mustermann",
+  "spouse": {
+    "href": "https://...",
+    "since": "1996-12-19",
+    "id": "123e4567-e89b-12d3-a456-426655440000",
+    "name": "Linda Mustermann"
+  }
+}
+```
 
 Hypertext controls are allowed anywhere within a JSON model. While this specification would
 allow [HAL](http://stateless.co/hal_specification.html), we actually don't recommend/enforce the 
--- a/hyper-media/Hypermedia.md
+++ b/hyper-media/Hypermedia.md
@@ -32,7 +32,7 @@ When embedding links to other resources into representations you must use the co
 
 * `href`: The URI of the resource the hypertext control is linking to. All our API are using HTTP(s) as URI scheme. 
 
-When designing an API that contains any hypertext controls, the attribute name `href` should be reserved for usage within hypertext controls.
+In API that contain any hypertext controls, the attribute name `href` is reserved for usage within hypertext controls.
 
 The schema for hypertext controls can be derived from this model:
 
--- a/hyper-media/Hypermedia.md
+++ b/hyper-media/Hypermedia.md
@@ -26,62 +26,53 @@ with APIs explicitly defined outside the code with standard specification langua
 
 However, we do not forbid HATEOAS; you could use it, if you checked its limitations and still see clear value for your usage scenario that justifies its additional complexity. If you use HATEOAS please share experience and present your findings in the [API Guild \[internal link\]](https://techwiki.zalando.net/display/GUL/API+Guild).
 
-## {{ book.must }} Use a well-defined subset of HAL
+## {{ book.must }} Use common hypertext controls
 
-Links to other resources must be defined exclusively using [HAL](http://stateless.co/hal_specification.html) and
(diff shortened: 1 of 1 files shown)


hyper-media/Hypermedia.md
```diff
@@ -26,62 +26,53 @@ with APIs explicitly defined outside the code with standard specification langua
 
 However, we do not forbid HATEOAS; you could use it, if you checked its limitations and still see clear value for your usage scenario that justifies its additional complexity. If you use HATEOAS please share experience and present your findings in the [API Guild \[internal link\]](https://techwiki.zalando.net/display/GUL/API+Guild).
 
-## {{ book.must }} Use a well-defined subset of HAL
```
dehora: I think this pr is going to need to update "Use Custom Link Relations" (further down in the section) as well - that section has a direct HAL reference and uses `_links` in its example.

hyper-media/Hypermedia.md
```diff
@@ -26,62 +26,53 @@ with APIs explicitly defined outside the code with standard specification langua
...
-| [Collection+JSON](http://amundsen.com/media-types/collection/) | ✗          | ✗             | ✗        | Collections and queries | ✓    |
-
-We define HAL links to be extensible, i.e. to contain additional properties if needed. For consistency extensions should reuse attributes from the [`Link` header](https://tools.ietf.org/html/rfc5988#section-5).
-
-Interesting articles for comparisons of different hypermedia formats:
-* [Kevin Sookocheff’s On choosing a hypermedia type for your API](http://sookocheff.com/post/api/on-choosing-a-hypermedia-format/)
-* [Mike Stowe's API Best Practices: Hypermedia](http://blogs.mulesoft.com/dev/api-dev/api-best-practices-hypermedia-part-3/)
+## {{ book.must }} Use common hypertext controls
+
+When embedding links to other resources into representations you must use the common hypertext control object. It contains at least one attribute:
```
dehora: > you must use the common hypertext control object

The ability to easily inline hrefs is neat, but as I understand this text, hrefs inside an object will be the only allowed way to declare a link, That doesn't leave remove for describing relations, eg in the example, `spouse` is regular data and not a rel:

```json
{
   "spouse": {
         "href": "https://...",
         "since": "1996-12-19",
         "id": "123e4567-e89b-12d3-a456-426655440000",
         "name": "Linda Mustermann"
       }
}
```

a rel could be emulated if chosen to nest another level:

```json
{
   "spouse": {
        "self": {
           "href": "https://...",
         },
         "since": "1996-12-19",
         "id": "123e4567-e89b-12d3-a456-426655440000",
         "name": "Linda Mustermann"
       }
}
```

That means relations can't be automatically distinguished from regular data anymore. This is probably fine for now, and maybe it gets addressed in the follow up as it'll possibly be relevant to structures like pagination. 

Speaking of pagination, the guideline "May: Use Pagination Links Where Applicable" in [Pagination](http://zalando.github.io/restful-api-guidelines/pagination/Pagination.html) has an implicit HAL  reference via `_links` but not HAL directly, so probably can be left out of scope for now.

hyper-media/Hypermedia.md
meshcalero: Agreed. I was just not sure if we could remove the section completely (as the reference to standard link relation is already in the new hypertext control section), or if we still want to have a dedicated section for it

hyper-media/Hypermedia.md
meshcalero: For this specific example I was assuming we only have "person" resources and "spouse" is the relation between two persons. But still often a relation has additional data attached (in this case only the "since" attribute, as "id" and "name" are actually attributes of the referenced person) and I believe this is a pragmatic way to model it that matches the mental model of the consumer. This should not require to model each relation as its own resource. 

For pagination I believe the consequence would be that relations with a carnality larger than 1 would require a dedicated "items" property, e.g.

```
    {
      "id": "446f9876-e89b-12d3-a456-426655440000",
      "name": "Peter Mustermann",
      "children": {
        "next": {
          "href": "https://..."
        }
        "items": [
          {
            "href": "https://...",
            "id": "123e4567-e89b-12d3-a456-426655440000",
            "name": "Max Mustermann"
          }
        ]
      }
    }
```

hyper-media/Hypermedia.md
```diff
@@ -26,62 +26,53 @@ with APIs explicitly defined outside the code with standard specification langua
...
+      type: object
+      properties:
+        href:
+          description: |
+            Any URI that is using http or https protocol
+          type: string
+          format: uri
+      required: [ "href" ]
+ 
+The name of an attribute holding such a HttpLink object specifies the relation between the object that
```
tkrop: I would like mark `HttpLink` marked as code here.

tkrop:
Very concise, is this every aspect of hypertext control we should cover?

ePaul:
One of the complaints about our HAL rule before was that we always had to unwrap several layers of objects to see whether there is a next-link or not (pagination was about the only use of links).

Previously we had this:

```json
{
   "_links": { "next": { "href" : "https://..." } },
   "items": [ ... ]
}
```
Which needed something like this to access:
```java
public String getNextLink(SomeListDto list) {
   if(list.getLinks() != null && list.getLinks().getNext() != null && list.getNext().getHref() != null) {
      return list.getLinks().getNext().getHref();
   } else {
     return null;
   }
}
```

I guess now we have this:
```json
{
   "next": { "href" : "https://..." },
   "items": [ ... ]
}
```
... and can access it like this:
```java
public String getNextLink(SomeListDto list) {
   if(list.getNext() != null && list.getHref() != null) {
      return list.getNext().getHref();
   } else {
     return null;
   }
}
```
I'm not sure this is much better. Any chance we can simply allow this?
```json
{
   "next": "https://...",
   "items": [ ... ]
}
```

meshcalero:
@tkrop As said, there will be another PR that will propose a richer hypertext control object for more self-describing API, that will be an extension of this minimalitic one. Here I focus on API that use documentation to describe how and when to use a given hyperlink.

meshcalero:
@ePaul I think that URI-only hypertext controls are possible. E.g. before I used the following four preconditions: 

- the name of a hypertext control is a IANA-registered rel-type
- the HTTP method is GET 
- the linked resource returns the same representation as the object enclosing the hypertext control
- the linked resource accepts the same authorization as the source resource

I haven't put this in this PR as it relates also to the extensions for the richer self-describing hypertext control and I believe it makes sense to discuss the two topics together.

ePaul:
In Zalando (with our OpenAPI definitions), we don't have a full hypertext with generic links to resources of arbitrary types anyways.
Whenever there is a link to another resource, we know what kind of link there is (or might be) from the API definition. So I don't see much use of the requirements to put the links into "hypertext control" objects.

tkrop:
I like @ePaul very short version for `next`, `prev`, `last`, `first`, etc, and I think we should include this in a hypertext controls special subsection for pagination. Anyhow, for now we should go with the simple hypertext controls as described in the PR.

tkrop:
👍

hyper-media/Hypermedia.md
```diff
@@ -26,62 +26,56 @@ with APIs explicitly defined outside the code with standard specification langua
...
 
-Links to other resources must be defined exclusively using [HAL](http://stateless.co/hal_specification.html) and
-preferably using standard [link relations](http://www.iana.org/assignments/link-relations/link-relations.xml).
+When embedding links to other resources into representations you must use the common hypertext control object. It contains at least one attribute:
 
-Clients and Servers are required to support `_links` with its `href` and `rel` attributes, not only at the root level
-but also in nested objects. To reduce the effort needed by clients to process hypertext data from servers it's not recommended to serve data with CURIEs, URI templates or embedded resources. Nor is it required to support the HAL media type `application/hal+json`. The following snippet specifies the HAL link structure in JSONSchema:
+* `href`: The URI of the resource the hypertext control is linking to. All our API are using HTTP(s) as URI scheme. 
+
+When designing an API that contains any hypertext controls, the attribute name `href` should be reserved for usage within hypertext controls.
```
dehora: > should be reserved 

is reserved (avoids anyone getting confused about the optionality, since the guideline itself is a must)

tkrop: Good catch 👍

tkrop:
👍

dehora:
:+1:

whiskeysierra:
Since this wasn't discussed before, how do we deal with link relations that violate our naming conventions, e.g. `version-history`?

PR #248, opened 2017-06-20
Refinement of ignore-unknown-fields guideline

### LINKED ISSUE #165: Refinement of "ignore unknown fields" guideline
We currently have the general rule "Should: Prefer Compatible Extensions" and the more client specific rule "Must: Prepare Clients for Compatible API Extensions (the Robustness Principle)".
Both are partially overlapping -- I think we should specialise the first to be server specific and avoid redundancies and even inconsistencies. 

For the server side, I'm not sure whether here robustness principle generally applies; there are some problems, for instance...
a) might not be compatible with semantics of operation, e.g. PUT for resource storage service should not ignore unknown input fields and do not store it without any feedback 
b) ignore unknown input fields may have deferred impact to future extensions which are supposed to be compatible, however, might break clients that use a different type here

Having this in mind, I propose on server side "ignore unknown fields" is not a "should", but a "could" rule.
If robustness will not be provided, I would expect that still additional properties is true -- because I want to have the option to extend the data structure in the future (see issue #164 ) -- but a generic error code (which one?) is returned. 

Yes, different services might want to take different approaches with "unknown fields" -- but the service provider should be aware of the problems mentioned above and be explicit if it is not supported.

tfrauenstein:
Fixes #165.

### DIFF
--- a/changelog/Changelog.md
+++ b/changelog/Changelog.md
@@ -8,6 +8,7 @@ To see a list of all changes, please have a look at the [commit list in Github](
 
 ## Rule Changes
 
+* `2017-06-20:` Be more precise on client vs. server obligations for compatible API extensions.
 * `2017-06-06:` Made money object guideline clearer.
 * `2017-05-17:` Added guideline on query parameter collection format.
 * `2017-05-10:` Added the convention of using RFC2119 to describe guideline 
--- a/changelog/Changelog.md
+++ b/changelog/Changelog.md
@@ -8,7 +8,7 @@ To see a list of all changes, please have a look at the [commit list in Github](
 
 ## Rule Changes
 
-* `2017-06-20:` Be more precise on client vs. server obligations for compatible API extensions.
+* `2017-07-20:` Be more precise on client vs. server obligations for compatible API extensions.
 * `2017-06-06:` Made money object guideline clearer.
 * `2017-05-17:` Added guideline on query parameter collection format.
 * `2017-05-10:` Added the convention of using RFC2119 to describe guideline 
--- a/compatibility/Compatibility.md
+++ b/compatibility/Compatibility.md
@@ -49,7 +49,7 @@ API designers should apply the following rules to evolve RESTful APIs for servic
 * Support redirection in case an URL has to change
   ([301 Moved Permanently](https://en.wikipedia.org/wiki/HTTP_301)).
 
-## {{ book.must }} Clients Do Not Crash With Compatible API Extensions
+## {{ book.must }} Prepare Clients To Not Crash On Compatible API Extensions
 
 Service clients apply the robustness principle and must be prepared for compatible API extensions of service providers:
 
--- a/compatibility/Compatibility.md
+++ b/compatibility/Compatibility.md
@@ -31,8 +31,11 @@ changes are necessary.
 
 API designers should apply the following rules to evolve RESTful APIs for services in a backward-compatible way:
 
-* Add only optional, never mandatory fields
-* Never change the meaning of a field (e.g. by being more restrictive).
+* Add only optional, never mandatory fields.
+* Never change the meaning of a field (e.g. switching to customer-number 
+  semantics for customer-id, both being unique customer keys).
+* Input fields may have (complexer) constaints being validated via server-side business logic.
+  Never change the validation logic to be more restrictive and make sure that constraints a clearly defined in description. 
 * Enum ranges can be reduced when used as input parameters, only if the server 
   is ready to accept and handle old range values too. Enum range can be reduced 
   when used as output parameters.
@@ -44,19 +47,19 @@ API designers should apply the following rules to evolve RESTful APIs for servic
   be extended with growing functionality. It defines an open list of explicit
   values and clients must be agnostic to new values.
 * Support redirection in case an URL has to change
-  ([301 Moved Permanently](https://en.wikipedia.org/wiki/HTTP_301))
+  ([301 Moved Permanently](https://en.wikipedia.org/wiki/HTTP_301)).
 
 ## {{ book.must }} Clients Do Not Crash With Compatible API Extensions
 
 Service clients apply the robustness principle and must be prepared for compatible API extensions of service providers:
 
 * Ignore new and unknown fields in the payload (see also Fowler’s
-  “[TolerantReader](http://martinfowler.com/bliki/TolerantReader.html)” post)
+  “[TolerantReader](http://martinfowler.com/bliki/TolerantReader.html)” post).
 * Be prepared that [`x-extensible-enum`](#should-used-openended-list-of-values-xextensibleenum-instead-of-enumerations)
   return parameter may deliver new values; either be agnostic or provide default behavior for unknown values. 
 * Be prepared to handle HTTP status codes not explicitly specified in endpoint definitions. 
   Note also, that status codes are  extensible. Default handling is how you would treat the 
-  corresponding x00 code (see [RFC7231  Section 6](https://tools.ietf.org/html/rfc7231#section-6))
+  corresponding x00 code (see [RFC7231  Section 6](https://tools.ietf.org/html/rfc7231#section-6)).
 * Follow the redirect when the server returns HTTP status
   [301 Moved Permanently](https://en.wikipedia.org/wiki/HTTP_301).
 
@@ -65,11 +68,11 @@ Service clients apply the robustness principle and must be prepared for compatib
 Designers of service provider APIs should be conservative and accurate in what they accept from clients:
 
 * Unknown input fields in payload or URL should not be ignored; 
-  provide error feedback to client via HTTP 400 return code.
+  provide error feedback to clients via HTTP 400 return code.
 * Be accurate in defining input data constraints (like formats, ranges, lengths etc.) — and 
-  checking the constraints with dedicated error return information in case of violations. 
-* Prefer being more specific and restrictive, if compliant to functional requirements, 
-  e.g. by defining length range of strings; it may simplify implementation while providing 
+  check constraints and return dedicated error information in case of violations. 
+* Prefer being more specific and restrictive (if compliant to functional requirements), 
+  e.g. by defining length range of strings. It may simplify implementation while providing 
   freedom for further evolution as compatible extensions. 
 
 Not ignoring unknown input fields is a specific deviation from Postel's Law and a strong recommendation.
@@ -82,18 +85,17 @@ problems and be explicit in what is supported:
   (see [RFC7231  Section 4.3.4](https://tools.ietf.org/html/rfc7231#section-4.3.4)).
   Note, accepting (i.e. not ignoring) unknown input fields and returning it in subsequent 
   GET responses is a different situation and compliant to PUT semantics. 
-* Certain client errors cannot be recognized by server, e.g. attribute name typing 
-  errors will be ignored without server error feedback. The server cannot differ between 
+* Certain client errors cannot be recognized by servers, e.g. attribute name typing 
+  errors will be ignored without server error feedback. The server cannot differentiate between 
   client provided intentionally an additional field vs. client field name typing error 
   where client intentionally wanted to provide a specific optional input field.
 * Future extensions of the input data structure might be in conflict with already 
   ignored fields and, hence, will not be compatible, i.e. break clients that already 
   use this field but with different type.
 
-In specific situations, where an (known) input field is not needed anymore
-it either can stay in the API definition with "not used anymore" description or 
-it can be removed from the API definition as long as the server ignores 
-this specific parameter to not break the clients. 
+In specific situations, where a (known) input field is not needed anymore, it either can 
+stay in the API definition with "not used anymore" description or can be removed from 
+the API definition as long as the server ignores this specific parameter. 
 
 ## {{ book.must }} Always Return JSON Objects As Top-Level Data Structures To Support Extensibility
 
--- a/compatibility/Compatibility.md
+++ b/compatibility/Compatibility.md
@@ -2,10 +2,10 @@
 
 ## {{ book.must }} Don’t Break Backward Compatibility
 
-Change APIs, but keep all consumers
+Change APIs, but keep all consumers 
 running. Consumers usually have independent release lifecycles, focus on
 stability, and avoid changes that do not provide additional value. APIs are
-service contracts between service provider and its clients that cannot be broken via unilateral decisions.
+contracts between service providers and service consumers that cannot be broken via unilateral decisions.
 
 There are two techniques to change APIs without breaking them:
 
@@ -13,8 +13,8 @@ There are two techniques to change APIs without breaking them:
 - introduce new API versions and still support older versions
 
 We strongly encourage using compatible API extensions and discourage versioning.
-The guideline rules below for providers and consumers enable us with Postel’s Law 
-in mind to make compatible changes without versioning.
+The below guideline rules for service providers and consumers enable us (having Postel’s Law in mind) 
+to make compatible changes without versioning.
 
 **Hint:** Please note that the compatibility guarantees are for the on-the-wire format,
 i.e. they ensure that consumers relying on the old API definition will work
@@ -27,9 +27,9 @@ process to a new version of the API definition, it has to be expected that code
 changes are necessary.
 
 
-## {{ book.should }} Service Provider Prefer Compatible Extensions
+## {{ book.should }} Prefer Compatible Extensions
 
-Service provider should apply the following rules to evolve RESTful APIs in a backward-compatible way:
+API designers should apply the following rules to evolve RESTful APIs for services in a backward-compatible way:
 
 * Add only optional, never mandatory fields
 * Never change the meaning of a field (e.g. by being more restrictive).
@@ -46,7 +46,7 @@ Service provider should apply the following rules to evolve RESTful APIs in a ba
 * Support redirection in case an URL has to change
   ([301 Moved Permanently](https://en.wikipedia.org/wiki/HTTP_301))
 
-## {{ book.must }} Service Clients Do Not Crash With Compatible API Extensions
+## {{ book.must }} Not Crash Clients With Compatible API Extensions
 
 Service clients apply the robustness principle and must be prepared for compatible API extensions of service providers:
 
@@ -60,19 +60,21 @@ Service clients apply the robustness principle and must be prepared for compatib
 * Follow the redirect when the server returns HTTP status
   [301 Moved Permanently](https://en.wikipedia.org/wiki/HTTP_301).
 
-## {{ book.should }} Service Provider Are Conservative
+## {{ book.should }} Design Service Providers Conservatively 
 
-Service provider should be conservative and accurate in what they accept from clients:
+Designers of service APIs should be conservative and accurate in what they accept from clients:
 
-* Unknown fields in payload or URL should not be ignored.
+* Unknown fields in payload or URL should not be ignored; 
+  provide error feedback to client via HTTP 400 return code.
 * Be accurate in defining input data constraints (like formats, ranges, lengths etc.) — and 
   checking the constraints with dedicated error return information in case of violations. 
 * Prefer being more specific and restrictive, if compliant to functional requirements, 
   e.g. by defining length range of strings; it may simplify implementation while providing 
   freedom for further evolution as compatible extensions. 
 
-This is a strong recommendation; servers might want to take different approach with 
-"unknown fields" but should be aware of the following problems and be explicit in what is supported: 
+Not ignoring unknown fields is different from Postel's Law and a strong recommendation.
+Servers might want to take different approach but should be aware of the following 
+problems and be explicit in what is supported: 
 
 * Ignoring unknown fields is actually no option for PUT, since it becomes asymmetric 
   with subsequent GET response and HTTP is pretty clear about the PUT "replace" 
@@ -86,6 +88,10 @@ This is a strong recommendation; servers might want to take different approach w
   ignored fields and, hence, will not be compatible, i.e. break clients that already 
   use this field but with different type.
 
+In specific situations, where a (known) field is not needed anymore as input parameter
+it either can stay in the API definition with "not used anymore" description or it can 
+be removed from the API definition as long as the server ignores this specific parameter 
+to not break the clients. 
 
 ## {{ book.must }} Always Return JSON Objects As Top-Level Data Structures To Support Extensibility
 
--- a/compatibility/Compatibility.md
+++ b/compatibility/Compatibility.md
@@ -46,7 +46,7 @@ API designers should apply the following rules to evolve RESTful APIs for servic
 * Support redirection in case an URL has to change
   ([301 Moved Permanently](https://en.wikipedia.org/wiki/HTTP_301))
 
-## {{ book.must }} Not Crash Clients With Compatible API Extensions
+## {{ book.must }} Clients Do Not Crash With Compatible API Extensions
 
 Service clients apply the robustness principle and must be prepared for compatible API extensions of service providers:
 
@@ -60,11 +60,11 @@ Service clients apply the robustness principle and must be prepared for compatib
 * Follow the redirect when the server returns HTTP status
   [301 Moved Permanently](https://en.wikipedia.org/wiki/HTTP_301).
 
-## {{ book.should }} Design Service Providers Conservatively 
+## {{ book.should }} Design APIs Conservatively 
 
-Designers of service APIs should be conservative and accurate in what they accept from clients:
+Designers of service provider APIs should be conservative and accurate in what they accept from clients:
 
-* Unknown fields in payload or URL should not be ignored; 
+* Unknown input fields in payload or URL should not be ignored; 
   provide error feedback to client via HTTP 400 return code.
 * Be accurate in defining input data constraints (like formats, ranges, lengths etc.) — and 
   checking the constraints with dedicated error return information in case of violations. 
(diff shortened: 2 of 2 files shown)


compatibility/Compatibility.md
```diff
@@ -5,61 +5,87 @@
...
 changes are necessary.
 
 
-## {{ book.should }} Prefer Compatible Extensions
+## {{ book.should }} Service Provider Prefer Compatible Extensions
 
-Apply the following rules to evolve RESTful APIs in a backward-compatible way:
+Service provider should apply the following rules to evolve RESTful APIs in a backward-compatible way:
 
-* Ignore unknown fields in the payload
```
ePaul: I think this was needed here to allow removing fields which are not needed anymore.
If we remove that part of the rule, we need to have something about removing fields in a compatible way.

compatibility/Compatibility.md
```diff
@@ -5,61 +5,87 @@
...
 is not covered by these rules.  If client implementations update their generation
 process to a new version of the API definition, it has to be expected that code
 changes are necessary.
 
 
-## {{ book.should }} Prefer Compatible Extensions
+## {{ book.should }} Service Provider Prefer Compatible Extensions
 
-Apply the following rules to evolve RESTful APIs in a backward-compatible way:
+Service provider should apply the following rules to evolve RESTful APIs in a backward-compatible way:
```
ePaul: Grammar: subject (singular) and verb (plural) don't agree here.

Also, I'm not sure whether we should write something about the fact that those guideline are for "service provider = API designer". We have some cases where the API is defined by the consumer, which is then accessing multiple service providers (each implementing the same API). In that case the API evolution will happen with client/server roles switched around.

compatibility/Compatibility.md
```diff
@@ -5,61 +5,87 @@
...
+  Note also, that status codes are  extensible. Default handling is how you would treat the 
+  corresponding x00 code (see [RFC7231  Section 6](https://tools.ietf.org/html/rfc7231#section-6))
+* Follow the redirect when the server returns HTTP status
+  [301 Moved Permanently](https://en.wikipedia.org/wiki/HTTP_301).
+
+## {{ book.should }} Service Provider Are Conservative
+
+Service provider should be conservative and accurate in what they accept from clients:
+
+* Unknown fields in payload or URL should not be ignored.
```
ePaul: Can we note what should be done instead? (Return 400, log it as an error, crash the server ;-), ...)

compatibility/Compatibility.md
```diff
@@ -5,61 +5,87 @@
...
-* Be prepared to handle HTTP status codes not explicitly specified in endpoint definitions! Note also, that status codes are extensible -- default handling is how you would treat the corresponding x00 code (see [RFC7231  Section 6](https://tools.ietf.org/html/rfc7231#section-6))
+* Be prepared that [`x-extensible-enum`](#should-used-openended-list-of-values-xextensibleenum-instead-of-enumerations)
+  return parameter may deliver new values; either be agnostic or provide default behavior for unknown values. 
+* Be prepared to handle HTTP status codes not explicitly specified in endpoint definitions. 
+  Note also, that status codes are  extensible. Default handling is how you would treat the 
+  corresponding x00 code (see [RFC7231  Section 6](https://tools.ietf.org/html/rfc7231#section-6))
+* Follow the redirect when the server returns HTTP status
+  [301 Moved Permanently](https://en.wikipedia.org/wiki/HTTP_301).
+
+## {{ book.should }} Service Provider Are Conservative
```
ePaul: Can we find a better title for this guideline? If I see it in the table of contents, I have no idea what this is about.

By the way, this (like the one before) is more of an API implementation guideline, not so much about the API definition. Not sure if we need to point this out.

compatibility/Compatibility.md
```diff
@@ -5,61 +5,87 @@
...
-  values and clients must be agnostic to new values (e.g. via casuistic with
-  default behavior).
+  values and clients must be agnostic to new values.
 * Support redirection in case an URL has to change
- ([301 Moved Permanently](https://en.wikipedia.org/wiki/HTTP_301))
+  ([301 Moved Permanently](https://en.wikipedia.org/wiki/HTTP_301))
 
+## {{ book.must }} Service Clients Do Not Crash With Compatible API Extensions
 
-## {{ book.must }} Prepare Clients for Compatible API Extensions (the Robustness Principle)
```
ePaul: Wording: Previously most guidelines had a title that is readable/speakable when the requirements verb (MUST/SHOULD/MAY) is included, in an imperative mode of speech (i.e. beginning with the verb, without a subject).

Now we have something which is a full statement sentence with subject, and the requirements verb in front of it. I don't think this fits with the style we have elsewhere. (This applies to all the new titles changed by this PR.)

I don't have an idea how to word this better, unfortunately.

compatibility/Compatibility.md
```diff
@@ -5,61 +5,87 @@
...
+* Follow the redirect when the server returns HTTP status
+  [301 Moved Permanently](https://en.wikipedia.org/wiki/HTTP_301).
+
+## {{ book.should }} Service Provider Are Conservative
+
+Service provider should be conservative and accurate in what they accept from clients:
+
+* Unknown fields in payload or URL should not be ignored.
+* Be accurate in defining input data constraints (like formats, ranges, lengths etc.) — and 
+  checking the constraints with dedicated error return information in case of violations.
```
ePaul: Related: #177.

compatibility/Compatibility.md
```diff
@@ -5,61 +5,87 @@
...
 
-**Note:** Binary or source compatibility of code generated from an API definition
+Hence, binary or source compatibility of code generated from an API definition
 is not covered by these rules.  If client implementations update their generation
 process to a new version of the API definition, it has to be expected that code
 changes are necessary.
 
 
-## {{ book.should }} Prefer Compatible Extensions
+## {{ book.should }} Service Provider Prefer Compatible Extensions
```
tkrop: Heading sounds like a query and not like a guideline ... 😸 How about:

`## {{ book.should }} Prefer Compatible Extensions with Respect to Service Providers`

compatibility/Compatibility.md
tkrop: How about:

`Service provider should apply` -> `With respect to service providers API designers should apply`?

Is it to longish?

compatibility/Compatibility.md
tkrop: I would stick to the old title here without the robustness principle:

`## {{ book.must }} Prepare Clients for Compatible API Extensions`

compatibility/Compatibility.md
tkrop: My 2 cents here:  we should remove the reference to Postel's Law in this chapter: This section is definitely going against the robustness principle, and we have a good reason to do so.

How about `## {{ book.should }} Design Service Provider Conservative`?

compatibility/Compatibility.md
```diff
@@ -5,61 +5,87 @@
...
+* Be accurate in defining input data constraints (like formats, ranges, lengths etc.) — and 
+  checking the constraints with dedicated error return information in case of violations. 
+* Prefer being more specific and restrictive, if compliant to functional requirements, 
+  e.g. by defining length range of strings; it may simplify implementation while providing 
+  freedom for further evolution as compatible extensions. 
+
+This is a strong recommendation; servers might want to take different approach with 
+"unknown fields" but should be aware of the following problems and be explicit in what is supported: 
+
+* Ignoring unknown fields is actually no option for PUT, since it becomes asymmetric
```
whiskeysierra: I disagree. We have read-only properties that the client will ignore (and bascially not update). There are also audit properties, e.g. last modified. From the linked section of the RFC:

> A successful PUT of a given
   representation would suggest that a subsequent GET on that same
   target resource will result in an **equivalent representation** being
   sent in a 200 (OK) response.  However, there is no guarantee that
   such a state change will be observable, since the target resource
   might be acted upon by other user agents in parallel, or might be
   **subject to dynamic processing by the origin server**, before any
   subsequent GET is received.

Ignoring unknown fields sounds like a valid approach to me.

compatibility/Compatibility.md
tfrauenstein: Ok, changed title and related text to service API designers.

compatibility/Compatibility.md
tfrauenstein: yes, added, thx.

compatibility/Compatibility.md
tfrauenstein: 1) yes, changed tittle as proposed. 
2) added hint that recommendation is different than Postel's law.

dehora: > Can we find a better title for this guideline?

```
 ## {{ book.should }} Avoid accepting undefined data from clients
```

Removes the service provider concept, but imo that's ok because this is about data definitions, insofar as clients should expect to have put/post requests rejected with data that is not defined in the API file. 

The obvious exception to this are systems that hold arbitrary data such as storage apis. Another are APIs where certain objects within definitions are defined as as free-form. 

An important class that this guideline doesn't benefit, are saas products such as analytics systems who don't know what information is being sent by customers (this is arguably an industry category). So I'm glad this is should rather than must guidance.

compatibility/Compatibility.md
tfrauenstein: ok, changed it to: "{{ book.must }} Not Crash Clients With Compatible API Extensions"

compatibility/Compatibility.md
tfrauenstein: yes, but since #177 is still pending, we have at least already some hints here.

compatibility/Compatibility.md
```diff
@@ -5,61 +5,87 @@
 Change APIs, but keep all consumers
 running. Consumers usually have independent release lifecycles, focus on
 stability, and avoid changes that do not provide additional value. APIs are
-service contracts that cannot be broken via unilateral decisions.
+service contracts between service provider and its clients that cannot be broken via unilateral decisions.
```
tfrauenstein: done

compatibility/Compatibility.md
```diff
@@ -5,61 +5,87 @@
...
 
 There are two techniques to change APIs without breaking them:
 
 - follow rules for compatible extensions
 - introduce new API versions and still support older versions
 
 We strongly encourage using compatible API extensions and discourage versioning.
-With Postel’s Law in mind, the following five rules for providers and consumers
-allow us to make compatible changes without versioning.
+The guideline rules below for providers and consumers enable us with Postel’s Law
```
tfrauenstein: done

compatibility/Compatibility.md
tfrauenstein: yes, deleted "Service Provider"

compatibility/Compatibility.md
dehora: > My 2 cents here: we should remove the reference to Postel's Law in this chapter: This section is definitely going against the robustness principle,

It might do harm to remove the reference as it might result in pointless debates, but it's not quite against the principle. more of a refinement. The guidelines are describing sender and receiver behaviour separately, which is what a document that is  trying to "standardize" behaviour should be doing (see [rfc2360](https://tools.ietf.org/html/rfc2360#section-2.9)).

compatibility/Compatibility.md
tfrauenstein: Yes, the server now generally should not ignore unknown fields -- see rule below. 
The guidelines now differ clearly between server and client view. 
 
And yes, good hint: there might be the specific situation that the server wants to remove specific fields not needed anymore -- I added a corresponding hint at end of the "be conservative" rule.

compatibility/Compatibility.md
tfrauenstein: as discussed: ignoring read-only, known input fields is ok, and compliant with PUT semantics. 
Here it is about ignoring unknown input fields  -- this is not compliant with PUT semantics. 
Supporting unknown fields as input *and* GET output would be symmetric and compliant -- I added corresponding hint here.

compatibility/Compatibility.md
```diff
@@ -2,64 +2,98 @@
...
-  values and clients must be agnostic to new values (e.g. via casuistic with
-  default behavior).
+  values and clients must be agnostic to new values.
 * Support redirection in case an URL has to change
- ([301 Moved Permanently](https://en.wikipedia.org/wiki/HTTP_301))
-
+  ([301 Moved Permanently](https://en.wikipedia.org/wiki/HTTP_301))
 
-## {{ book.must }} Prepare Clients for Compatible API Extensions (the Robustness Principle)
+## {{ book.must }} Clients Do Not Crash With Compatible API Extensions
```
tkrop: This heading does not work spoken out correctly. How about:

```
## {{ book.must }} Prepare Clients To Not Crash On Compatible API Extensions
```

compatibility/Compatibility.md
```diff
@@ -2,64 +2,98 @@
...
+  corresponding x00 code (see [RFC7231  Section 6](https://tools.ietf.org/html/rfc7231#section-6))
+* Follow the redirect when the server returns HTTP status
+  [301 Moved Permanently](https://en.wikipedia.org/wiki/HTTP_301).
+
+## {{ book.should }} Design APIs Conservatively 
+
+Designers of service provider APIs should be conservative and accurate in what they accept from clients:
+
+* Unknown input fields in payload or URL should not be ignored; 
+  provide error feedback to client via HTTP 400 return code.
```
tkrop: `client` -> `clients`

compatibility/Compatibility.md
```diff
@@ -2,64 +2,98 @@
...
+* Be accurate in defining input data constraints (like formats, ranges, lengths etc.) — and 
+  checking the constraints with dedicated error return information in case of violations.
```
tkrop: sentence does not work: `checking` ... `constraints` ... `in case of violations`.

May be:
`checking the constraints with dedicated error return information in case of violations.` -> 
`check constraints and return dedicated error information in case of violations.`

compatibility/Compatibility.md
```diff
@@ -2,64 +2,98 @@
...
+
+## {{ book.should }} Design APIs Conservatively 
+
+Designers of service provider APIs should be conservative and accurate in what they accept from clients:
+
+* Unknown input fields in payload or URL should not be ignored; 
+  provide error feedback to client via HTTP 400 return code.
+* Be accurate in defining input data constraints (like formats, ranges, lengths etc.) — and 
+  checking the constraints with dedicated error return information in case of violations. 
+* Prefer being more specific and restrictive, if compliant to functional requirements,
```
tkrop: Split sentence here.

compatibility/Compatibility.md
```diff
@@ -2,64 +2,98 @@
...
+Servers might want to take different approach but should be aware of the following 
+problems and be explicit in what is supported: 
+
+* Ignoring unknown input fields is actually no option for PUT, since it becomes asymmetric 
+  with subsequent GET response and HTTP is pretty clear about the PUT "replace" 
+  semantics and default roundtrip expectations 
+  (see [RFC7231  Section 4.3.4](https://tools.ietf.org/html/rfc7231#section-4.3.4)).
+  Note, accepting (i.e. not ignoring) unknown input fields and returning it in subsequent 
+  GET responses is a different situation and compliant to PUT semantics. 
+* Certain client errors cannot be recognized by server, e.g. attribute name typing
```
tkrop: `server` -> `the server` or `the service`

Maybe we should in general talk more often about a `service` instead of `server` in this chapter.

compatibility/Compatibility.md
```diff
@@ -2,64 +2,98 @@
...
+problems and be explicit in what is supported: 
+
+* Ignoring unknown input fields is actually no option for PUT, since it becomes asymmetric 
+  with subsequent GET response and HTTP is pretty clear about the PUT "replace" 
+  semantics and default roundtrip expectations 
+  (see [RFC7231  Section 4.3.4](https://tools.ietf.org/html/rfc7231#section-4.3.4)).
+  Note, accepting (i.e. not ignoring) unknown input fields and returning it in subsequent 
+  GET responses is a different situation and compliant to PUT semantics. 
+* Certain client errors cannot be recognized by server, e.g. attribute name typing 
+  errors will be ignored without server error feedback. The server cannot differ between
```
tkrop: `differ` -> `differentiate`?

compatibility/Compatibility.md
```diff
@@ -2,64 +2,98 @@
...
+  GET responses is a different situation and compliant to PUT semantics. 
+* Certain client errors cannot be recognized by server, e.g. attribute name typing 
+  errors will be ignored without server error feedback. The server cannot differ between 
+  client provided intentionally an additional field vs. client field name typing error 
+  where client intentionally wanted to provide a specific optional input field.
+* Future extensions of the input data structure might be in conflict with already 
+  ignored fields and, hence, will not be compatible, i.e. break clients that already 
+  use this field but with different type.
+
+In specific situations, where an (known) input field is not needed anymore
```
tkrop: `an (known) input` -> `a (known) input`

compatibility/Compatibility.md
```diff
@@ -2,64 +2,98 @@
...
+it either can stay in the API definition with "not used anymore" description or 
+it can be removed from the API definition as long as the server ignores 
+this specific parameter to not break the clients.
```
tkrop: `this specific parameter to not break the clients.` -> `this specific parameter.`

compatibility/Compatibility.md
tfrauenstein: done

compatibility/Compatibility.md
tfrauenstein: done, thx

compatibility/Compatibility.md
tfrauenstein: done

compatibility/Compatibility.md
tfrauenstein: done

compatibility/Compatibility.md
tfrauenstein: done

compatibility/Compatibility.md
tfrauenstein: done

compatibility/Compatibility.md
tfrauenstein: done

compatibility/Compatibility.md
```diff
@@ -2,64 +2,100 @@
...
-  values and clients must be agnostic to new values (e.g. via casuistic with
-  default behavior).
+  values and clients must be agnostic to new values.
 * Support redirection in case an URL has to change
- ([301 Moved Permanently](https://en.wikipedia.org/wiki/HTTP_301))
-
+  ([301 Moved Permanently](https://en.wikipedia.org/wiki/HTTP_301)).
 
-## {{ book.must }} Prepare Clients for Compatible API Extensions (the Robustness Principle)
+## {{ book.must }} Clients Do Not Crash With Compatible API Extensions
```
tkrop: This heading does not work spoken out correctly. How about:

```
## {{ book.must }} Prepare Clients To Not Crash On Compatible API Extensions
```

(Copied from last review - maybe you have unintentional undone your change).

tfrauenstein: done

compatibility/Compatibility.md
```diff
@@ -2,64 +2,100 @@
...
-* Never change the meaning of a field.
-* Enum ranges cannot not be extended when used for output parameters — clients
-  may not be prepared to handle it. However, enum ranges can be extended when
-  used for input parameters.
-* Enum ranges can be reduced when used as input parameters, only if the server
-  is ready to accept and handle old range values too. Enum values can be reduced
+* Add only optional, never mandatory fields.
+* Never change the meaning of a field (e.g. switching to customer-number 
+  semantics for customer-id, both being unique customer keys).
+* Input fields may have (complexer) constaints being validated via server-side business logic.
```
dehora: > (complexer)

"""
complex
""""

compatibility/Compatibility.md
```diff
@@ -2,64 +2,100 @@
...
+
+* Unknown input fields in payload or URL should not be ignored; 
+  provide error feedback to clients via HTTP 400 return code.
+* Be accurate in defining input data constraints (like formats, ranges, lengths etc.) — and 
+  check constraints and return dedicated error information in case of violations. 
+* Prefer being more specific and restrictive (if compliant to functional requirements), 
+  e.g. by defining length range of strings. It may simplify implementation while providing 
+  freedom for further evolution as compatible extensions. 
+
+Not ignoring unknown input fields is a specific deviation from Postel's Law and a strong recommendation.
```
dehora: This might be a useful link: https://cacm.acm.org/magazines/2011/8/114933-the-robustness-principle-reconsidered/fulltext - it discusses examples where ignoring unknown input presents issues and postel's law presents issues

compatibility/Compatibility.md
```diff
@@ -2,64 +2,100 @@
...
+Servers might want to take different approach but should be aware of the following 
+problems and be explicit in what is supported: 
+
+* Ignoring unknown input fields is actually no option for PUT, since it becomes asymmetric
```
dehora: > actually no option for PUT

"""
actually not an option for PUT
"""

compatibility/Compatibility.md
```diff
@@ -2,64 +2,100 @@
...
+* Prefer being more specific and restrictive (if compliant to functional requirements), 
+  e.g. by defining length range of strings. It may simplify implementation while providing 
+  freedom for further evolution as compatible extensions. 
+
+Not ignoring unknown input fields is a specific deviation from Postel's Law and a strong recommendation.
+Servers might want to take different approach but should be aware of the following 
+problems and be explicit in what is supported: 
+
+* Ignoring unknown input fields is actually no option for PUT, since it becomes asymmetric 
+  with subsequent GET response and HTTP is pretty clear about the PUT "replace"
```
dehora: > is pretty clear 

"""
is clear 
"""

compatibility/Compatibility.md
```diff
@@ -2,64 +2,100 @@
...
+* Ignoring unknown input fields is actually no option for PUT, since it becomes asymmetric 
+  with subsequent GET response and HTTP is pretty clear about the PUT "replace" 
+  semantics and default roundtrip expectations 
+  (see [RFC7231  Section 4.3.4](https://tools.ietf.org/html/rfc7231#section-4.3.4)).
+  Note, accepting (i.e. not ignoring) unknown input fields and returning it in subsequent 
+  GET responses is a different situation and compliant to PUT semantics. 
+* Certain client errors cannot be recognized by servers, e.g. attribute name typing 
+  errors will be ignored without server error feedback. The server cannot differentiate between 
+  client provided intentionally an additional field vs. client field name typing error 
+  where client intentionally wanted to provide a specific optional input field.
```
dehora: > The server cannot differentiate between 
  client provided intentionally an additional field vs. client field name typing error 
  where client intentionally wanted to provide a specific optional input field.

"""
The server cannot differentiate between the client intentionally providing an additional field versus the client sending a mistakenly named field, when the client's actual intent was to provide an optional input field.
"""

compatibility/Compatibility.md
```diff
@@ -2,64 +2,100 @@
...
+  corresponding x00 code (see [RFC7231  Section 6](https://tools.ietf.org/html/rfc7231#section-6)).
+* Follow the redirect when the server returns HTTP status
+  [301 Moved Permanently](https://en.wikipedia.org/wiki/HTTP_301).
+
+## {{ book.should }} Design APIs Conservatively 
+
+Designers of service provider APIs should be conservative and accurate in what they accept from clients:
+
+* Unknown input fields in payload or URL should not be ignored; 
+  provide error feedback to clients via HTTP 400 return code.
```
dehora: > Unknown input fields in payload or URL should not be ignored;  provide error feedback to clients via HTTP 400 return code.

"""
Unknown input fields in payload or URL should not be ignored; servers should provide error feedback to clients via an HTTP 400 response code.
"""

compatibility/Compatibility.md
tfrauenstein: done

compatibility/Compatibility.md
tfrauenstein: done

compatibility/Compatibility.md
tfrauenstein: interesting link -- inserted; thx!

compatibility/Compatibility.md
tfrauenstein: done

compatibility/Compatibility.md
tfrauenstein: done

compatibility/Compatibility.md
tfrauenstein: done and thx.

dehora:
👍

changelog/Changelog.md
```diff
@@ -8,6 +8,7 @@ To see a list of all changes, please have a look at the [commit list in Github](
 
 ## Rule Changes
 
+* `2017-06-20:` Be more precise on client vs. server obligations for compatible API extensions.
```
duergner: We should update the date here once we're ready to merge, shouldn't we?

compatibility/Compatibility.md
```diff
@@ -2,64 +2,102 @@
...
-* Never change the meaning of a field.
-* Enum ranges cannot not be extended when used for output parameters — clients
-  may not be prepared to handle it. However, enum ranges can be extended when
-  used for input parameters.
-* Enum ranges can be reduced when used as input parameters, only if the server
-  is ready to accept and handle old range values too. Enum values can be reduced
+* Add only optional, never mandatory fields.
+* Never change the meaning of a field (e.g. switching to customer-number 
+  semantics for customer-id, both being unique customer keys).
+* Input fields may have (complex) constaints being validated via server-side business logic.
```
duergner: Editorial: `(complex) constaints being` -> `(complex) constraints being`

tkrop: Good catch 👍

changelog/Changelog.md
tkrop: Yes, this looks quite outdated now 😸

tfrauenstein: done

compatibility/Compatibility.md
tfrauenstein: done

tkrop:
👍

dehora:
:+1:

PR #250, opened 2017-06-20
Combine common attributes name and address attribute names in a single section #203

### LINKED ISSUE #203: Common Data Objects: Separate Receipient from Address
Currently the Address model covers two aspects WHO and WHERE. If you want to model recipients with multiple addresses, or add additional information about the person receiving something that has nothing to do with its address, you can't nicely do this.

If we would instead have two distinct models (e.g. Party and Address) for the two aspects we could overcome that problem.

To avoid misunderstanding, we could also rename the attribute business_name within address to something like care_of, to make it clear that the business is not the recipient, but only the place where you find the recipient. If we want to model businesses as recipients (AFAIK we don't cover B2C business yet), we could still keep business_name in recipient.

### LINKED ISSUE #263: Mention "Link object" in "14 Common Data Objects" (improvement)
Requirement "Must: Use common hypertext controls" in section "12. Hypermedia" mentions "Link Object". It looks helpful to refer or mention it in section "14 Common Data Objects" also.

meshcalero:
Fixes #203.
Fixes #263.

### DIFF
--- a/common-data-types/CommonDataTypes.md
+++ b/common-data-types/CommonDataTypes.md
@@ -60,9 +60,9 @@ There are some data fields that come up again and again in API data:
 
 - `xyz_id`: an attribute within one object holding the identifier of another object must use a name that corresponds to the type of the referenced object or the relationship to the referenced object followed by `_id` (e.g. `customer_id` not `customer_number`; `parent_node_id` for the reference to a parent node from a child node, even if both have the type `Node`)
 
-- `created`: when the object was created. If used, this must be a date-time construct.
+- `created`: when the object was created. If used, this must be a `date-time` construct.
 
-- `modified`: when the object was updated. If used, this must be a date-time construct.
+- `modified`: when the object was updated. If used, this must be a `date-time` construct.
 
 - `type`: the kind of thing this object is. If used, the type of this field should be a string. Types allow runtime information on the entity provided that otherwise requires examining the Open API file.
 
@@ -103,22 +103,6 @@ Address structures play a role in different functional and use-case contexts, in
 variances. All attributes that relate to address information should follow the naming
 and semantics defined below.
 
-- `salutation`: a salutation and/or title used for personal contacts to some addressee.
-    __Hint__: not to be confused with the gender information that is stored per customer account
-
-- `first_name`: given name(s) or first name(s) of a person; may also include the middle names.
-- `last_name`: family name(s) or surname(s) of a person
-- `business_name`: company name of the business organization. Used when a business is the actual addressee. For personal shipments to office addresses, use `care_of` instead.
-- `care_of` (_c/o_): the person that resides at the address, if different from addressee. E.g. used when sending a personal parcel to the office / someone else's home where the addressee resides temporarily.
-- `street`: the full street address including house number and street name
-- `additional`: further details like building name, suite, apartment number, etc.
-- `city`: name of the city / locality
-- `zip`: zip code or postal code
-- `country_code`: the country code according to [iso-3166-1-alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2)
-
-Grouping and cardinality of fields in specific data types may vary based on the specific use case
-(e.g. combining addressee and address fields into a single type when modeling an address label vs distinct addressee and address types when modeling users and their addresses).
-
     addressee:
         description:
           a (natural or legal) person that gets addressed
@@ -132,15 +116,25 @@ Grouping and cardinality of fields in specific data types may vary based on the 
           - country_code
         properties:
           salutation:
+            description: |
+              a salutation and/or title used for personal contacts to some addressee;
+              not to be confused with the gender information!
             type: string
             example: Mr
           first_name:
+            description: |
+              given name(s) or first name(s) of a person; may also include the middle names.
             type: string
             example: Hans Dieter
           last_name:
+            description: |
+              family name(s) or surname(s) of a person
             type: string
             example: Mustermann
           business_name:
+            description: |
+              company name of the business organization. Used when a business is the actual
+              addressee; for personal shipments to office addresses, use `care_of` instead.
             type: string
             example: Consulting Services GmbH
         required:
@@ -153,21 +147,36 @@ Grouping and cardinality of fields in specific data types may vary based on the 
         type: object
         properties:
           care_of:
+            description: |
+              (aka c/o) the person that resides at the address, if different from
+              addressee. E.g. used when sending a personal parcel to the 
+              office /someone else's home where the addressee resides temporarily
             type: string
             example: Consulting Services GmbH
           street:
+            description: |
+              the full street address including house number and street name
             type: string
             example: Schönhauser Allee 103
           additional:
+            description: |
+              further details like building name, suite, apartment number, etc.
             type: string
             example: 2. Hinterhof rechts
           city:
+            description: |
+              name of the city / locality
             type: string
             example: Berlin
           zip:
+            description: |
+              zip code or postal code
             type: string
             example: 14265
           country_code:
+            description: |
+              the country code according to 
+              [iso-3166-1-alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2)
             type: string
             example: DE
         required:
@@ -175,6 +184,14 @@ Grouping and cardinality of fields in specific data types may vary based on the 
           - city
           - zip
           - country_code
+
+Grouping and cardinality of fields in specific data types may vary based on the specific use case
+(e.g. combining addressee and address fields into a single type when modeling an address label vs distinct addressee and address types when modeling users and their addresses).
+
+## {{ book.must }} Follow Hypertext Control Conventions
+
+API that provide hypertext controls (links) to interconnect API resources must follow
+the conventions for naming and modeling of hypertext controls as defined in section. [Hypermedia](../hyper-media/Hypermedia.html). 
 
 ## {{ book.must }} Use Problem JSON
 
--- a/common-data-types/CommonDataTypes.md
+++ b/common-data-types/CommonDataTypes.md
@@ -58,13 +58,13 @@ There are some data fields that come up again and again in API data:
 
 - `id`: the identity of the object. If used, IDs must opaque strings and not numbers. IDs are unique within some documented context, are stable and don't change for a given object once assigned, and are never recycled cross entities.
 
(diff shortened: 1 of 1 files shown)


common-data-types/CommonDataTypes.md
```diff
@@ -47,15 +47,81 @@ Some JSON parsers (NodeJS’s, for example) convert numbers to floats by default
...
-    address:
+There exist a variety of field types that are required in multiple places. To achieve consistency across all API implementations, you must use common field names and semantics whenever applicable.
+
+### Generic Fields
+
+There are some data fields that come up again and again in API data:
+
+- `id`: the identity of the object. If used, IDs must opaque strings and not numbers. IDs are unique within some documented context, are stable and don't change for a given object once assigned, and are never recycled cross entities.
+
+- _`xyz_`_`id`: an attribute within one object holding the identifier of another object must use a name that corresponds to the type of the referenced object or the relationship to the referenced object followed by `_id` (e.g. `customer_id` not `customer_number`; `parent_node_id` for the reference to a parent node from a child node, even if both have the type `Node`)
```
ePaul: Editorial: The italics here doesn't really work. I would use just `xyz_id`.

common-data-types/CommonDataTypes.md
```diff
@@ -47,15 +47,81 @@ Some JSON parsers (NodeJS’s, for example) convert numbers to floats by default
...
+
+- `created`: when the object was created. If used this must be a date-time construct.
+
+- `modified`: when the object was updated. If used this must be a date-time construct.
```
ePaul: Editorial: add commas here after "If used".

common-data-types/CommonDataTypes.md
```diff
@@ -47,15 +47,81 @@ Some JSON parsers (NodeJS’s, for example) convert numbers to floats by default
...
+          description: when got this node created
+          type: string
+          format: 'date-time'
+        modified:
+          description: when got this node last updated
+          type: string
+          format: 'date-time'
+        type:
+          type: string
+          enum: [ 'LEAF', 'NODE' ]
```
tkrop: We should use `x-extensible-enum` and refer to this annotation above too. Types have the incompatible custom to be extended over time.

common-data-types/CommonDataTypes.md
```diff
@@ -67,41 +133,48 @@ cases. Use it in your APIs — and compatible extend it if necessary for your AP
         properties:
           salutation:
             type: string
-            description: |
```
tkrop: We should have proper descriptions in a `template`. They are meant for copy & past. Even if we define the common semantics in the guideline, an external user should be enabled to read an API without having the guideline in mind.

common-data-types/CommonDataTypes.md
```diff
@@ -47,15 +47,81 @@ Some JSON parsers (NodeJS’s, for example) convert numbers to floats by default
...
+
+### Generic Fields
+
+There are some data fields that come up again and again in API data:
+
+- `id`: the identity of the object. If used, IDs must opaque strings and not numbers. IDs are unique within some documented context, are stable and don't change for a given object once assigned, and are never recycled cross entities.
+
+- `xyz_id`: an attribute within one object holding the identifier of another object must use a name that corresponds to the type of the referenced object or the relationship to the referenced object followed by `_id` (e.g. `customer_id` not `customer_number`; `parent_node_id` for the reference to a parent node from a child node, even if both have the type `Node`)
+
+- `created`: when the object was created. If used, this must be a date-time construct.
```
whiskeysierra: `created` to me sounds more like a boolean property. What about `created_at`?

common-data-types/CommonDataTypes.md
```diff
@@ -47,15 +47,81 @@ Some JSON parsers (NodeJS’s, for example) convert numbers to floats by default
...
+These properties are not always strictly necessary, but making them idiomatic allows API client developers to build up a common understanding of Zalando's resources. There is very little utility for API consumers in having different names or value types for these fields across APIs.
+
+### Address Fields
+
+Address structures play a role in different functional and use-case contexts, including country
+variances. All attributes that relate to address information should follow the naming
+and semantics defined below.
+
+- `salutation`: a salutation and/or title used for personal contacts to some addressee.
+    __Hint__: not to be confused with the gender information that is stored per customer account
```
whiskeysierra: That is very Zalando specific.

common-data-types/CommonDataTypes.md
whiskeysierra: I see, the were defined like that already.

common-data-types/CommonDataTypes.md
meshcalero: I see your point. But then we have quite some redundancy: We said, we want to have both bullet points in the text and an example, so I moved most of the descriptions to the text. Having the descriptions in both text and schema examples is redundant and ugly to maintain. I would not want to go down this path. We could either remove the descriptions in the text (and by that make examples part of the official rule) or keep it as it is now.

common-data-types/CommonDataTypes.md
meshcalero: Agreed. I just did not want to change the descriptions that had already been there.  I'd suggest to just keep the hint that this is not the same as gender, but remove the reference to customer account

common-data-types/CommonDataTypes.md
meshcalero: Not sure this is important in the context of a naming example and believe in that specific example of a generic tree a final enum might actually be appropriate. But happy to follow any majority vote here.

tkrop: Yes, in this special case you are right. Tree is pretty final in its enum set.

common-data-types/CommonDataTypes.md
tkrop: Yes I see your point, but I still would go for the redundancy. I know the laziness of developers using copy and past. Having a good template hanging around reduces the pain point of defining address objects over and over again a lot.

tkrop:
Can you integrate #263 here?

tkrop:
👍

common-data-types/CommonDataTypes.md
```diff
@@ -47,15 +47,65 @@ Some JSON parsers (NodeJS’s, for example) convert numbers to floats by default
...
+
+There exist a variety of field types that are required in multiple places. To achieve consistency across all API implementations, you must use common field names and semantics whenever applicable.
+
+### Generic Fields
+
+There are some data fields that come up again and again in API data:
+
+- `id`: the identity of the object. If used, IDs must opaque strings and not numbers. IDs are unique within some documented context, are stable and don't change for a given object once assigned, and are never recycled cross entities.
+
+- `xyz_id`: an attribute within one object holding the identifier of another object must use a name that corresponds to the type of the referenced object or the relationship to the referenced object followed by `_id` (e.g. `customer_id` not `customer_number`; `parent_node_id` for the reference to a parent node from a child node, even if both have the type `Node`)
```
whiskeysierra: Could have a sentence describing the difference between IDs and links and when to use which one.

common-data-types/CommonDataTypes.md
```diff
@@ -47,15 +47,65 @@ Some JSON parsers (NodeJS’s, for example) convert numbers to floats by default
...
+### Address Fields
 
 Address structures play a role in different functional and use-case contexts, including country
-variances. The address structure below should be sufficient for most of our business-related use
-cases. Use it in your APIs — and compatible extend it if necessary for your API concerns:
+variances. All attributes that relate to address information should follow the naming
+and semantics defined below.
 
-    address:
+    addressee:
```
whiskeysierra: I strongly recommend this post: http://www.siliconglen.scot/usability/courtesytitles.html

It raises valid questions regarding:
 - splitting up first and last name
 - fixed set of salutations/titles (not an issue in this guideline)
 - deduce gender from title

common-data-types/CommonDataTypes.md
```diff
@@ -66,42 +116,82 @@ cases. Use it in your APIs — and compatible extend it if necessary for your AP
           - country_code
         properties:
           salutation:
-            type: string
             description: |
-              A salutation and/or title which may be used for personal contacts. Hint: not to be confused with the gender information that is stored per customer account
+              a salutation and/or title used for personal contacts to some addressee;
+              not to be confused with the gender information!
+            type: string
             example: Mr
```
whiskeysierra: Using a different, less common example, like e.g. *Lord Commander* would make people not assume it's a fixed enum containing *Mrs* and *Mr*.

common-data-types/CommonDataTypes.md
```diff
@@ -66,42 +116,82 @@ cases. Use it in your APIs — and compatible extend it if necessary for your AP
...
           first_name:
+            description: |
+              given name(s) or first name(s) of a person; may also include the middle names.
             type: string
-            description: given name(s) or first name(s) of a person; may also include the middle names
             example: Hans Dieter
           last_name:
```
whiskeysierra: > Having one field is also very important if you do not want to offend over one fifth of the world's population who write their surname first (including China and Japan). What would writing Smith John rather than John Smith do for your customer care? Having two fields does not allow you to concatenate them in the correct order, whereas asking someone to enter their full name uses their preferences rather than your assumptions. With one field you can easily accommodate all nationalities, naming preferences, name orders, suffixes, and titles. Amazon manage this, why is it such a problem for everyone else?
>
> http://www.siliconglen.scot/usability/courtesytitles.html

tkrop:
@whiskeysierra I agreed with everything you suggest, but at the moment we try to update the *as is state* to the next level without changing the semantic. The article raises valid questions, that we should tackle in a future issue with the help of marketing and other stakeholders. Could you be so nice to open raise the issue?

common-data-types/CommonDataTypes.md
```diff
@@ -66,42 +116,82 @@ cases. Use it in your APIs — and compatible extend it if necessary for your AP
...
+          - city
+          - zip
+          - country_code
+
+Grouping and cardinality of fields in specific data types may vary based on the specific use case
+(e.g. combining addressee and address fields into a single type when modeling an address label vs distinct addressee and address types when modeling users and their addresses).
+
+## {{ book.must }} Follow Hypertext Control Conventions
+
+API that provide hypertext controls (links) to interconnect API resources must follow
```
dehora: > API that 

"""
APIs that 
"""

whiskeysierra:
@tkrop I created #269

tkrop:
👍

whiskeysierra:
:+1:

ePaul:
:+1:

Issue #254, opened 2017-06-22
Clarify the usage of X-App-Domain and X-Sales-Channel Headers


maxim-tschumak:
and incorporate into the existing guidelines.

Related to / triggered by #228.

tkrop:
Only discovered one usage in our API repository. Same API is also responsible for other headers from #253.

ePaul:
**From the meeting**: We don't need to change anything here.

Issue #253, opened 2017-06-22
Remove the mention of the X-UID Header


maxim-tschumak:
and improve the guidelines.

Related to / triggered by #228.

ePaul:
Maybe this can be removed completely? Is there any reason to have it in an API definition?

tkrop:
Only discovered one usage in our API repository. Same API is also responsible for other headers from #254.

PR #255, opened 2017-06-22
Adding Proprietary Headers Description + 2 new Must Guidelines

### LINKED ISSUE #228: Clarify client-side requirements around X- headers
The [Proprietary Headers](http://zalando.github.io/restful-api-guidelines/headers/ProprietaryHeaders.html) indicate API use of well-known X- Headers is optional:

>  Whether services support these concerns or not is optional

X- headers (ignoring the history of the X- prefix for experimental development) are often discretionary for clients (eg MSFT's guidelines state they should not be required for an API to work).

As an example, last March, the Nakadi API, which is based on these guidelines was asked to remove the use of X- header:

> headers should be reserved for contextual information like language or flow-ids that are not part of the data / data model of the API.

That was in the context of querying, the full discussion is [here](https://github.com/zalando/nakadi/pull/593#discussion_r106128860). That guidance is reasonable but more generally whether API designers can demand clients must use X- headers  to work effectively with a service should be clarified.

maxim-tschumak:
Closes #228.

### DIFF
--- a/headers/ProprietaryHeaders.md
+++ b/headers/ProprietaryHeaders.md
@@ -5,6 +5,16 @@ they address overarching service-related concerns. Whether services support thes
 optional; therefore, the OpenAPI API specification is the right place to make this explicitly
 visible. Use the parameter definitions of the resource HTTP methods.
 
+HTTP/1.1 standard ([RFC-2068](https://www.ietf.org/rfc/rfc2068.txt)) defines two types of headers: end-to-end and hop-by-bop headers.
+End-to-end headers must be transmitted to the ultimate recipient of a request or response.
+Hop-by-hop headers, on the contrary, are meaningful for a single connection only.
+
+**All Zalando's proprietary headers are end-to-end headers**.
+
+In general, proprietary headers should be avoided.
+They can be used in cases when parameter represents a pass-through end-to-end header.
+A valid use-case of a proprietary header is providing context information which is not a part of the actual API, but is used by subsequent communication.
+
 Remember that HTTP header field names are not case-sensitive.
 
 | Header field name | Type    | Description                       | Header field value example                |
@@ -17,3 +27,30 @@ Remember that HTTP header field names are not case-sensitive.
 | X-Device-Type      | String | There are also use cases for steering customer experience (incl. features and content) depending on device type. Via this header info should be passed-through as generic aspect. Current range is smartphone, tablet, desktop, other | tablet |
 | X-Device-OS      | String | On top of device type above, we even want to differ between device platform, e.g. smartphone Android vs. iOS. Via this header info should be passed-through as generic aspect. Current range is iOS, Android, Windows, Linux, MacOS | Android |
 | X-App-Domain      | Integer | The app domain (i.e. shop channel context) of the request. Note, app-domain is a legacy concept that will be replaced in new platform by combinations of main CFA concerns like retailer, sales channel, country | 16 |
+
+
+## {{ book.must }} use only the Specified Proprietary Zalando Headers
+
+Avoid using proprietary headers other than specified in the overview above.
+
+`X-` headers were initially reserved for unstandardized parameters.
+Now, their usage is deprecated ([RFC-6648](https://tools.ietf.org/html/rfc6648)).
+This fact complicates the contract definition between consumer and producer of an API, since there is no aligned way of handling those headers.
+Additionally, there is a chance of having intermediate proxies in the cloud environment, which may drop or alter unstandardized headers.
+
+**Exception:** The only exception to this guideline are the conventional hop-by-hop `X-RateLimit-` headers which can be used as defined in [HTTP/Must-Use-429-with-Headers-For-Rate-Limits](../http/Http.md#must-use-429-with-headers-for-rate-limits).
+
+
+## {{ book.must }} pass-through End-to-end Proprietary Headers
+
+All specified above headers must be passed to the services down the call chain.
+The header names and values must remain unchanged.
+Passing unchanged header values must happen even if a subset of them is transmitted as a part of the message body in the successive communication.
+
+Some of the transitive services may require the meta information provided via proprietary headers and/or rely on it.
+Besides, the values of the custom headers can influence the results of the queries (e.g. the device type information influences the recommendation results).
+
+
+*Footnote:*
+The Internet Engineering Task Force's states in [RFC-6648](https://tools.ietf.org/html/rfc6648) that company specific header' names should incorporate the organization's name.
+We aim for backward compatibility, and therefore keep the `X-` prefix.
--- a/headers/ProprietaryHeaders.md
+++ b/headers/ProprietaryHeaders.md
@@ -11,15 +11,15 @@ In general, proprietary headers should be avoided.
 They can be used in cases when parameter represents a pass-through end-to-end header.
 A valid use-case of a proprietary header is providing context information which is not a part of the actual API, but is used by subsequent communication.
 
-From a conceptual point of view, the semantic of an operation should always be expressed by resource name and query parameters, i.e. what goes into the URL.
-Headers implement protocol-close functions like control, content negotiation and authentication.
-Thus, headers are reserved for general context information ([RFC-7231](https://tools.ietf.org/html/rfc7231)).
+From a conceptual point of view, the semantic of an operation should always be expressed by path and query parameters, i.e. what goes into the URL, as well as the content.
+Headers are used to implement functions close to the protocol layer, such as flow control, content negotiation, and authentication.
+Thus, headers are reserved for general context information ([RFC-7231](https://tools.ietf.org/html/rfc7231#section-5)).
 
 `X-` headers were initially reserved for unstandardized parameters.
 The usage of `X-` headers is deprecated ([RFC-6648](https://tools.ietf.org/html/rfc6648)).
 This fact complicates the contract definition between consumer and producer of an API, since there is no aligned way of handling those headers.
 
-You can use following proprietary headers specified in the overview.
+The following proprietary headers have been specified by this guideline for usage so far.
 Remember that HTTP header field names are not case-sensitive.
 
 | Header field name | Type    | Description                       | Header field value example                |
@@ -46,7 +46,7 @@ The header names and values must remain unchanged.
 Some of the transitive services may require the meta information provided via proprietary headers and/or rely on it.
 Besides, the values of the custom headers can influence the results of the queries (e.g. the device type information influences the recommendation results).
 
-The values of the proprietary headers may be used as a part of the request body in the following communication.
+The values of the proprietary headers may be used as a part of the request body in the subsequent communication.
 In such cases, the proprietary headers must also be propagated as headers to the successive calls, despite the duplication.
 
 
--- a/headers/ProprietaryHeaders.md
+++ b/headers/ProprietaryHeaders.md
@@ -8,8 +8,8 @@ visible. Use the parameter definitions of the resource HTTP methods.
 ## {{ book.must }} Use Only the Specified Proprietary Zalando Headers
 
 In general, proprietary headers should be avoided.
-They can be used in cases when parameter represents a pass-through end-to-end header.
-A valid use-case of a proprietary header is providing context information which is not a part of the actual API, but is used by subsequent communication.
+Still they can be used in cases where context parameters should be pass-through multiple services in an end-to-end fashion.
+A valid use-case of a proprietary header is providing context information which is not a part of the actual API, but is needed by subsequent communication.
 
 From a conceptual point of view, the semantic of an operation should always be expressed by path and query parameters, i.e. what goes into the URL, as well as the content.
 Headers are used to implement functions close to the protocol layer, such as flow control, content negotiation, and authentication.
--- a/headers/ProprietaryHeaders.md
+++ b/headers/ProprietaryHeaders.md
@@ -13,7 +13,7 @@ A valid use-case of a proprietary header is providing context information which 
 
 From a conceptual point of view, the semantic of an operation should always be expressed by resource name and query parameters, i.e. what goes into the URL.
 Headers implement protocol-close functions like control, content negotiation and authentication.
-Thus, headers are reserved for general context information [RFC-7231](https://tools.ietf.org/html/rfc7231).
+Thus, headers are reserved for general context information ([RFC-7231](https://tools.ietf.org/html/rfc7231)).
 
 `X-` headers were initially reserved for unstandardized parameters.
 The usage of `X-` headers is deprecated ([RFC-6648](https://tools.ietf.org/html/rfc6648)).
@@ -50,7 +50,9 @@ The values of the proprietary headers may be used as a part of the request body 
 In such cases, the proprietary headers must also be propagated as headers to the successive calls, despite the duplication.
 
 
+
 *Footnotes:*
+
 The Internet Engineering Task Force's states in [RFC-6648](https://tools.ietf.org/html/rfc6648) that company specific header' names should incorporate the organization's name.
 We aim for backward compatibility, and therefore keep the `X-` prefix.
 
--- a/headers/ProprietaryHeaders.md
+++ b/headers/ProprietaryHeaders.md
@@ -7,17 +7,19 @@ visible. Use the parameter definitions of the resource HTTP methods.
 
 ## {{ book.must }} Use Only the Specified Proprietary Zalando Headers
 
-Avoid using proprietary headers other than specified in the following overview.
-
 In general, proprietary headers should be avoided.
 They can be used in cases when parameter represents a pass-through end-to-end header.
 A valid use-case of a proprietary header is providing context information which is not a part of the actual API, but is used by subsequent communication.
 
-`X-` headers were initially reserved for unstandardized parameters.
-Now, their usage is deprecated ([RFC-6648](https://tools.ietf.org/html/rfc6648)).
-This fact complicates the contract definition between consumer and producer of an API, since there is no aligned way of handling those headers.
-Additionally, there is a chance of having intermediate proxies in the cloud environment, which may drop or alter unstandardized headers.
+From a conceptual point of view, the semantic of an operation should always be expressed by resource name and query parameters, i.e. what goes into the URL.
+Headers implement protocol-close functions like control, content negotiation and authentication.
+Thus, headers are reserved for general context information [RFC-7231](https://tools.ietf.org/html/rfc7231).
 
+`X-` headers were initially reserved for unstandardized parameters.
+The usage of `X-` headers is deprecated ([RFC-6648](https://tools.ietf.org/html/rfc6648)).
+This fact complicates the contract definition between consumer and producer of an API, since there is no aligned way of handling those headers.
+
+You can use following proprietary headers specified in the overview.
 Remember that HTTP header field names are not case-sensitive.
 
 | Header field name | Type    | Description                       | Header field value example                |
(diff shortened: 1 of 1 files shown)


headers/ProprietaryHeaders.md
```diff
@@ -5,6 +5,16 @@ they address overarching service-related concerns. Whether services support thes
 optional; therefore, the OpenAPI API specification is the right place to make this explicitly
 visible. Use the parameter definitions of the resource HTTP methods.
 
+HTTP/1.1 standard ([RFC-2068](https://www.ietf.org/rfc/rfc2068.txt)) defines two types of headers: end-to-end and hop-by-bop headers.
```
tkrop: I think we should place this explanation at begin of `## {{ book.must }} Pass-through End-to-end Proprietary Headers`.

headers/ProprietaryHeaders.md
```diff
@@ -5,6 +5,16 @@ they address overarching service-related concerns. Whether services support thes
...
+End-to-end headers must be transmitted to the ultimate recipient of a request or response.
+Hop-by-hop headers, on the contrary, are meaningful for a single connection only.
+
+**All Zalando's proprietary headers are end-to-end headers**.
```
tkrop: This is redundant to the rule. Please check how to integrate it to the rule.

maxim-tschumak:
@tkrop thank you for the feedback, I'll do the requested refactorings.

headers/ProprietaryHeaders.md
```diff
@@ -17,3 +30,26 @@ Remember that HTTP header field names are not case-sensitive.
...
+All Zalando's proprietary headers are end-to-end headers.
+
+All specified above headers must be passed to the services down the call chain.
+The header names and values must remain unchanged.
+Passing unchanged header values must happen even if a subset of them is transmitted as a part of the message body in the successive communication.
+
+Some of the transitive services may require the meta information provided via proprietary headers and/or rely on it.
+Besides, the values of the custom headers can influence the results of the queries (e.g. the device type information influences the recommendation results).
+
+HTTP/1.1 standard ([RFC-2068](https://www.ietf.org/rfc/rfc2068.txt)) defines two types of headers: end-to-end and hop-by-bop headers.
```
dehora: I'd prefer to not refer to an rfc that's obsoleted. This is covered in rfc2616, https://tools.ietf.org/html/rfc2616#section-13.5.1. Unfortunately that's obsolete as well, but I'm not sure if there's a definition of these terms in the recent RFCs. Also if you do keep a link reference, can you use `https://tools.ietf.org/html/` to keep the link consistent with the other two rfc links in the pr.

headers/ProprietaryHeaders.md
```diff
@@ -17,3 +30,26 @@ Remember that HTTP header field names are not case-sensitive.
...
+End-to-end headers must be transmitted to the ultimate recipient of a request or response.
+Hop-by-hop headers, on the contrary, are meaningful for a single connection only.
```
dehora: This classification of headers originally came about is to assist caches. I'm slightly concerned the usage here has different semantics. I see no mention of cache handling, but it seems they need to be propagated end to end for safe application processing purposes (eg tenant identification). If the latter is true, I would remove end to end as a concept and simply say they must be propagated.

headers/ProprietaryHeaders.md
dehora: > hop-by-bop

"hop-by-hop"

headers/ProprietaryHeaders.md
```diff
@@ -17,3 +30,26 @@ Remember that HTTP header field names are not case-sensitive.
...
+
+## {{ book.must }} Pass-through End-to-end Proprietary Headers
+
+All Zalando's proprietary headers are end-to-end headers.
+
+All specified above headers must be passed to the services down the call chain.
+The header names and values must remain unchanged.
+Passing unchanged header values must happen even if a subset of them is transmitted as a part of the message body in the successive communication.
+
+Some of the transitive services may require the meta information provided via proprietary headers and/or rely on it.
```
dehora: The text reads as if propagation happens on the request path. Is it required on the response path, ie are origin servers producing responses required to reflect X- headers sent to them?

I'm wondering if the propagation rules need to be stated per header.

headers/ProprietaryHeaders.md
```diff
@@ -5,6 +5,19 @@ they address overarching service-related concerns. Whether services support thes
...
+## {{ book.must }} Use Only the Specified Proprietary Zalando Headers
+
+Avoid using proprietary headers other than specified in the following overview.
+
+In general, proprietary headers should be avoided.
+They can be used in cases when parameter represents a pass-through end-to-end header.
+A valid use-case of a proprietary header is providing context information which is not a part of the actual API, but is used by subsequent communication.
+
+`X-` headers were initially reserved for unstandardized parameters.
+Now, their usage is deprecated ([RFC-6648](https://tools.ietf.org/html/rfc6648)).
```
dehora: >+Now, 

The opening now can be removed; rfc6648 is five years old and predates these guidelines.

headers/ProprietaryHeaders.md
```diff
@@ -5,6 +5,19 @@ they address overarching service-related concerns. Whether services support thes
...
+This fact complicates the contract definition between consumer and producer of an API, since there is no aligned way of handling those headers.
+Additionally, there is a chance of having intermediate proxies in the cloud environment, which may drop or alter unstandardized headers.
```
tkrop: Have found this: https://stackoverflow.com/questions/20820572/under-what-conditions-are-http-request-headers-removed-by-proxies

While the specs have good guidelines, a use case was described where an nginx dropped headers when reaching its limit of 4k (had big cookies).

headers/ProprietaryHeaders.md
```diff
@@ -17,3 +30,26 @@ Remember that HTTP header field names are not case-sensitive.
...
 | X-App-Domain      | Integer | The app domain (i.e. shop channel context) of the request. Note, app-domain is a legacy concept that will be replaced in new platform by combinations of main CFA concerns like retailer, sales channel, country | 16 |
+
+**Exception:** The only exception to this guideline are the conventional hop-by-hop `X-RateLimit-` headers which can be used as defined in [HTTP/Must-Use-429-with-Headers-For-Rate-Limits](../http/Http.md#must-use-429-with-headers-for-rate-limits).
+
+
+## {{ book.must }} Pass-through End-to-end Proprietary Headers
+
+All Zalando's proprietary headers are end-to-end headers.
+
+All specified above headers must be passed to the services down the call chain.
```
tkrop: I agree, this rule is, besides its good purpose to make headers in down stream services available, problematic from a security perspective. This rule requires services to forward information they are not fully aware of, which open side channels and attack vectors.

headers/ProprietaryHeaders.md
```diff
@@ -17,3 +30,26 @@ Remember that HTTP header field names are not case-sensitive.
 | X-Device-Type      | String | There are also use cases for steering customer experience (incl. features and content) depending on device type. Via this header info should be passed-through as generic aspect. Current range is smartphone, tablet, desktop, other | tablet |
 | X-Device-OS      | String | On top of device type above, we even want to differ between device platform, e.g. smartphone Android vs. iOS. Via this header info should be passed-through as generic aspect. Current range is iOS, Android, Windows, Linux, MacOS | Android |
 | X-App-Domain      | Integer | The app domain (i.e. shop channel context) of the request. Note, app-domain is a legacy concept that will be replaced in new platform by combinations of main CFA concerns like retailer, sales channel, country | 16 |
+
+**Exception:** The only exception to this guideline are the conventional hop-by-hop `X-RateLimit-` headers which can be used as defined in [HTTP/Must-Use-429-with-Headers-For-Rate-Limits](../http/Http.md#must-use-429-with-headers-for-rate-limits).
+
+
+## {{ book.must }} Pass-through End-to-end Proprietary Headers
```
dehora: > Pass Through Proprietary End-to-end Headers

lgtm if the end-to-end concept still applies (see comments elsewhere re caching) and pass through is less jargon than propagate. If e-2-e doesn't apply "Pass Through Proprietary Headers" might work

headers/ProprietaryHeaders.md
dehora: > problematic from a security perspective. This rule requires services to forward information they are not fully aware of

That's a good point, I hadn't considered it.

headers/ProprietaryHeaders.md
maxim-tschumak: You are right @dehora, this paragraph should not be part of the guideline.
How do you feel about moving this additional information to the footnotes? (My initial goal of this text was to have some explanation of the two types of header and the differences between them)

headers/ProprietaryHeaders.md
maxim-tschumak: Should I add the Stackoverflow answer (as a link)?

maxim-tschumak:
Many thanks for your review @dehora. I'm working on improvements.

headers/ProprietaryHeaders.md
```diff
@@ -17,3 +30,26 @@ Remember that HTTP header field names are not case-sensitive.
...
+
+All Zalando's proprietary headers are end-to-end headers.
+
+All specified above headers must be passed to the services down the call chain.
+The header names and values must remain unchanged.
+Passing unchanged header values must happen even if a subset of them is transmitted as a part of the message body in the successive communication.
```
maxim-tschumak: As @tfrauenstein mentioned, there are valid use cases where the following pattern applies:

1. Component A sends a request to component B with the header `X-header: header_value`
2. Component B sends a request to component C with body {property: header_value}
In this case, component B **must also send the initial header** (and not the value only as a body/parameter/etc.)

I'll try to rephrase this paragraph.

headers/ProprietaryHeaders.md
maxim-tschumak: We have some follow-up issues regarding the usage of the particular headers (see #253, #254).

headers/ProprietaryHeaders.md
dehora: > How do you feel about moving this additional information to the footnotes?

:+1: worth a try, and it's relevant for context, even if the x- headers themselves aren't strictly end-to-end in the exact http terms.

ePaul: Correct, End-to-End means from a HTTP perspective (being passed forward by proxies).
That a service is calling another service in order to fulfill its purpose is an implementation detail.

headers/ProprietaryHeaders.md
ePaul: I found [RCF 7230, section 3.2.1 Field Extensibility](https://tools.ietf.org/html/rfc7230#section-3.2.1), which says that a proxy should forward any unknown headers. [Section 6.1](https://tools.ietf.org/html/rfc7230#section-6.1) talks about hop-by-hop and end-to-end headers (which are distinguished by the former ones being listed in the Connection header).

maxim-tschumak: Good finding @ePaul, thanks!

maxim-tschumak:
@tkrop @dehora @ePaul thanks for the comments. I've incorporated the changes, please have another look.

headers/ProprietaryHeaders.md
```diff
@@ -5,6 +5,21 @@ they address overarching service-related concerns. Whether services support thes
 optional; therefore, the OpenAPI API specification is the right place to make this explicitly
 visible. Use the parameter definitions of the resource HTTP methods.
 
+## {{ book.must }} Use Only the Specified Proprietary Zalando Headers
+
+In general, proprietary headers should be avoided.
+They can be used in cases when parameter represents a pass-through end-to-end header.
+A valid use-case of a proprietary header is providing context information which is not a part of the actual API, but is used by subsequent communication.
+
+From a conceptual point of view, the semantic of an operation should always be expressed by resource name and query parameters, i.e. what goes into the URL.
```
tkrop: May be more precise `by resource name and query parameters, i.e. what goes into the URL.` -> `by path and query parameters, i.e. what goes into the URL, as well as the content.`

headers/ProprietaryHeaders.md
```diff
@@ -5,6 +5,21 @@ they address overarching service-related concerns. Whether services support thes
...
+Headers implement protocol-close functions like control, content negotiation and authentication.
```
tkrop: May be better `Headers are used to implement functions close to the protocol layer, such as flow control, content negotiation, and authentication.`

headers/ProprietaryHeaders.md
```diff
@@ -5,6 +5,21 @@ they address overarching service-related concerns. Whether services support thes
...
+Thus, headers are reserved for general context information ([RFC-7231](https://tools.ietf.org/html/rfc7231)).
+
+`X-` headers were initially reserved for unstandardized parameters.
+The usage of `X-` headers is deprecated ([RFC-6648](https://tools.ietf.org/html/rfc6648)).
+This fact complicates the contract definition between consumer and producer of an API, since there is no aligned way of handling those headers.
+
+You can use following proprietary headers specified in the overview.
```
tkrop: May be better `The following proprietary headers have been specified by this guideline for usage so far.`

headers/ProprietaryHeaders.md
```diff
@@ -17,3 +32,30 @@ Remember that HTTP header field names are not case-sensitive.
...
+
+All Zalando's proprietary headers are end-to-end headers.
+
+All headers specified above must be propagated to the services down the call chain.
+The header names and values must remain unchanged.
+
+Some of the transitive services may require the meta information provided via proprietary headers and/or rely on it.
+Besides, the values of the custom headers can influence the results of the queries (e.g. the device type information influences the recommendation results).
+
+The values of the proprietary headers may be used as a part of the request body in the following communication.
```
tkrop: `following` -> `subsequent`

headers/ProprietaryHeaders.md
```diff
@@ -5,6 +5,21 @@ they address overarching service-related concerns. Whether services support thes
...
 
+## {{ book.must }} Use Only the Specified Proprietary Zalando Headers
+
+In general, proprietary headers should be avoided.
+They can be used in cases when parameter represents a pass-through end-to-end header.
+A valid use-case of a proprietary header is providing context information which is not a part of the actual API, but is used by subsequent communication.
+
+From a conceptual point of view, the semantic of an operation should always be expressed by resource name and query parameters, i.e. what goes into the URL.
+Headers implement protocol-close functions like control, content negotiation and authentication.
+Thus, headers are reserved for general context information ([RFC-7231](https://tools.ietf.org/html/rfc7231)).
```
ePaul: Could you point to the relevant section of the RFC (It is quite big)?

A potential candidate would be the introductory paragraph of [Section 5](https://tools.ietf.org/html/rfc7231#section-5), but maybe there are better ones.

maxim-tschumak:
Thanks for the input @tkrop and @ePaul. Proposed improvements are incorporated.

headers/ProprietaryHeaders.md
```diff
@@ -5,6 +5,21 @@ they address overarching service-related concerns. Whether services support thes
 optional; therefore, the OpenAPI API specification is the right place to make this explicitly
 visible. Use the parameter definitions of the resource HTTP methods.
 
+## {{ book.must }} Use Only the Specified Proprietary Zalando Headers
+
+In general, proprietary headers should be avoided.
+They can be used in cases when parameter represents a pass-through end-to-end header.
```
tkrop: `parameter represents` -> `a parameter represents` but may be better
`Still they can be used in cases where context parameters should be pass-through multiple services in an end-to-end fashion.`

headers/ProprietaryHeaders.md
```diff
@@ -5,6 +5,21 @@ they address overarching service-related concerns. Whether services support thes
...
+A valid use-case of a proprietary header is providing context information which is not a part of the actual API, but is used by subsequent communication.
```
tkrop: `used` -> `needed`

headers/ProprietaryHeaders.md
```diff
@@ -5,6 +5,21 @@ they address overarching service-related concerns. Whether services support thes
 optional; therefore, the OpenAPI API specification is the right place to make this explicitly
 visible. Use the parameter definitions of the resource HTTP methods.
 
+## {{ book.must }} Use Only the Specified Proprietary Zalando Headers
+
+In general, proprietary headers should be avoided.
```
dehora: """

As a general rule, proprietary HTTP headers should be avoided. Still they can be
 useful in cases where context needs to be passed through multiple services in
 an end-to-end fashion. As such, a valid use-case for a proprietary header is
 providing context information, which is not a part of the actual API, but is
 needed by subsequent communication.
 
 From a conceptual point of view, the semantics and intent of an operation
 should always be expressed by URLs path and query parameters, the method, and
 the content. Headers are more often used to implement functions close to the
 protocol considerations, such as flow control, content negotiation, and
 authentication. Thus, headers are reserved for general context information
 ([RFC-7231](https://tools.ietf.org/html/rfc7231#section-5)).
 
 `X-` headers were initially reserved for unstandardized parameters, but the
 usage of `X-` headers is deprecated
 ([RFC-6648](https://tools.ietf.org/html/rfc6648)). This complicates the
 contract definition between consumer and producer of an API following these
 guidelines, since there is no aligned way of using those headers. Because of
 this, the guidelines restrict which `X-` headers can be used and how they are
 used.
"""

headers/ProprietaryHeaders.md
```diff
@@ -17,3 +32,30 @@ Remember that HTTP header field names are not case-sensitive.
...
+
+
+## {{ book.must }} Propagate Proprietary Headers
+
+All Zalando's proprietary headers are end-to-end headers.
+
+All headers specified above must be propagated to the services down the call chain.
+The header names and values must remain unchanged.
+
+Some of the transitive services may require the meta information provided via proprietary headers and/or rely on it.
```
dehora: > Some of the transitive services may require the meta information provided via proprietary headers and/or rely on it.

"""
Some downstream services might require or rely on the information provided via proprietary headers.
"""

headers/ProprietaryHeaders.md
```diff
@@ -17,3 +32,30 @@ Remember that HTTP header field names are not case-sensitive.
...
+Besides, the values of the custom headers can influence the results of the queries (e.g. the device type information influences the recommendation results).
```
dehora: > Besides, the values of the custom headers can influence the results of the queries (e.g. the device type information influences the recommendation results).

"""
For example, the values of the custom headers like `X-Device-Type` can affect the results of queries by using device type information to influence recommendation results.
"""

headers/ProprietaryHeaders.md
```diff
@@ -17,3 +32,30 @@ Remember that HTTP header field names are not case-sensitive.
...
+
+The values of the proprietary headers may be used as a part of the request body in the subsequent communication.
+In such cases, the proprietary headers must also be propagated as headers to the successive calls, despite the duplication.
```
dehora: > The values of the proprietary headers may be used as a part of the request body in the subsequent communication.
> In such cases, the proprietary headers must also be propagated as headers to the successive calls, despite the duplication.

"""
Sometimes the value of a proprietary header will be used as part of the entity
in a subsequent request. In such cases, the proprietary headers must still be
propagated as headers with the subsequent request, despite the duplication of
information.
"""

headers/ProprietaryHeaders.md
```diff
@@ -17,3 +32,30 @@ Remember that HTTP header field names are not case-sensitive.
...
+
+
+
+*Footnotes:*
+
+The Internet Engineering Task Force's states in [RFC-6648](https://tools.ietf.org/html/rfc6648) that company specific header' names should incorporate the organization's name.
+We aim for backward compatibility, and therefore keep the `X-` prefix.
```
dehora: Suggest moving this rationale up to the main section on `X-` headers as it makes it clear why they're being kept around.

headers/ProprietaryHeaders.md
```diff
@@ -17,3 +32,30 @@ Remember that HTTP header field names are not case-sensitive.
...
+
+HTTP/1.1 standard ([RFC-7230](https://tools.ietf.org/html/rfc7230#section-6.1)) defines two types of headers: end-to-end and hop-by-hop headers.
```
dehora: I'm not sure this footnote is needed. The propagation model for the `X-` headers here is neither one of these. If something was to be said it's that `X-` header handling in these guidelines is not hop-by-hop or end-to-end in the HTTP sense. Also, this level of explanatory detail didn't make into the recent HTTP RFCs, so the guidelines here might be fine without it.

maxim-tschumak: The idea here was to emphasize the difference between the two types of headers and to have a relation the HTTP standard - that's why it's stated (only) as a footnote.

maxim-tschumak:
@dehora Last time I've submitted by CS paper to IEEE, it was rejected because of one `+0` rating.
`+/-0` score is a great quality gate :)

Thank you for the improvements!

tkrop:
👍

ePaul:
:+1:

PR #256, opened 2017-06-22
Introduce simplified hypertext controls for pagination and self-reference

meshcalero:
- Adding section on simplified hypertext controls for pagination and self-references
- Updating HATEOAS section in Pagination to no longer use HAL
- Removing other references to HAL from Hypermedia.md

### DIFF
--- a/pagination/Pagination.md
+++ b/pagination/Pagination.md
@@ -64,56 +64,25 @@ Those collections should then have an `items` attribute holding the items of the
 
 You should avoid providing a total count in your API unless there's a clear need to do so. Very often, there are systems and performance implications to supporting full counts, especially as datasets grow and requests become complex queries or filters that drive full scans (e.g., your database might need to look at all candidate items to count them). While this is an implementation detail relative to the API, it's important to consider your ability to support serving counts over the life of a service.
 
-If the collection consists of links to other resources, the collection object name should use [IANA registered link relations](http://www.iana.org/assignments/link-relations/link-relations.xml) as names whenever appropriate, but use plural form.
+If the collection consists of links to other resources, the collection name should use [IANA registered link relations](http://www.iana.org/assignments/link-relations/link-relations.xml) as names whenever appropriate, but use plural form.
 
-E.g. a service for articles could represent an `Article` that embeds a collection `authors` with hyperlinks to the article's authors like that:
-
-```json
-{
-  "self": "https://...",
-  "id": "446f9876-e89b-12d3-a456-426655440000",
-  "title": "Manifesto for Agile Software Development ",
-  "authors": {
-    "index": 0,
-    "page_size": 5,
-    "items": [
-      {  
-        "href": "https://...",
-        "id": "123e4567-e89b-12d3-a456-426655440000",
-        "name": "Kent Beck"
-      },
-      {  
-        "href": "https://...",
-        "id": "987e2343-e89b-12d3-a456-426655440000",
-        "name": "Mike Beedle"
-      },
-      ...
-    ],
-    "first": "https://...",
-    "next": "https://...",
-    "prev": "https://...",
-    "last": "https://..."
-  }
-}
-
-```
-Following any of the pagination links would then return another page of authors
-
+E.g. a service for articles could represent the collection of hyperlinks to an article's `authors` like that:
 
 ```json
 {
-  "index": 5,
+  "self": "https://.../articles/xyz/authors/",
+  "index": 0,
   "page_size": 5,
   "items": [
     {  
       "href": "https://...",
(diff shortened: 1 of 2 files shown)


pagination/Pagination.md
```diff
@@ -58,46 +58,81 @@ Further reading:
...
 
-```http
-GET http://catalog-service.zalando.net/products?offset=10&limit=5 HTTP/1.1
-Accept: application/json
+You should avoid providing a total count in your API unless there's a clear need to do so. Very often, there are systems and performance implications to supporting full counts, especially as datasets grow and requests become complex queries or filters that drive full scans (e.g., your database might need to look at all candidate items to count them). While this is an implementation detail relative to the API, it's important to consider your ability to support serving counts over the life of a service.
+
+If the collection consists of links to other resources, the collection object name should use [IANA registered link relations](http://www.iana.org/assignments/link-relations/link-relations.xml) as names whenever appropriate, but use plural form.
+
+E.g. a service for articles could represent an `Article` that embeds a collection `authors` with hyperlinks to the article's authors like that:
```
tkrop: @meshcalero respect, that is a very tricky example. While it is nice to enable such a pattern, I would not like to see it adopted widely. I assume that most embedded resources are quite limited as in the below case e.g. I do not know many article that have more than 100 authors.

May be we should stick to a standard top level example here. Other opinions on that?

tkrop: Will also shorten the example 😸

pagination/Pagination.md
```diff
@@ -58,46 +58,81 @@ Further reading:
...
 }
+
 ```
 
-[Possible relations](http://www.iana.org/assignments/link-relations/link-relations.xml):
-`next`, `prev`, `last`, `first`.
+Previous editions of the guidelines required the use HAL compliant hypertext controls. This requirement has been relaxed to enable a more [concise representation and placement of links](../hyper-media/Hypermedia.html#must-use-common-hypertext-controls).
 
-Previous editions of the guidelines documented the `X-Total-Count` header to send back the total count of entities in conjunction with the `Link` header, which has since been deprecated. Instead when returning an object structure, the count can be added as a JSON property, and this is the preferred way to return count (or other result level) information. 
+Previous editions of the guidelines also documented the `X-Total-Count` header to send back the total count of entities in conjunction with the `Link` header, which has since been deprecated. Instead when returning an object structure, the count can be added as a JSON property, and this is the preferred way to return count (or other result level) information.
```
tkrop: As the use of the `Link` header has now been discouraged for a longer while, I think we should drop this explanation as well as the following sentences now.

pagination/Pagination.md
ePaul: Pagination in embedded resource list is tricky, yes.

In a recent case, my team decided to remove an embedded resource instead of trying to support pagination there (we made it instead a normal subresource with its own URL).

pagination/Pagination.md
ePaul: Agreed. We even have some discussion of total count added in this PR (line 65).

pagination/Pagination.md
```diff
@@ -58,46 +58,81 @@ Further reading:
...
+Previous editions of the guidelines also documented the `X-Total-Count` header to send back the total count of entities in conjunction with the `Link` header, which has since been deprecated. Instead when returning an object structure, the count can be added as a JSON property, and this is the preferred way to return count (or other result level) information. 
 
 The `X-Total-Count` is applicable in these cases:
 
 * The API design is still using the `Link` header.
 * The API is returning a non-JSON response media type, and isn't able to carry the information.
 * The JSON response is an array and not an object.
 
-You should avoid providing a total count in your API unless there's a clear need to do so. Very often, there are systems and performance implications to supporting full counts, especially as datasets grow and requests become complex queries or filters that drive full scans (e.g., your database might need to look at all candidate items to count them). While this is an implementation detail relative to the API, it's important to consider your ability to support serving counts over the life of a service.
```
tfrauenstein: Why this hint has been removed? 
It is orthogonal to the changed pagination "syntax".

meshcalero: It hasn't been removed. I just moved it below the new paragraph about collection meta-information.

pagination/Pagination.md
meshcalero: While I believe that being able to embed collections is an important pattern, esp. for edge services that feed mobile devices with all data they need for a specific use case in a single response, I agree that it does make sense, to keep the example for that specific rule clean and not mix it up with the topic of embedded collections. The primary reason why I used it was to indicate that the notation works for both cases (embedded and standalone).

tkrop:
👍

duergner:
👍

dehora:
:+1:

Issue #237, opened 2017-05-18
Reduce guidance dependency on Nakadi project


dehora:
The event models defined in the Nakadi project are required in the guidelines but the data models are currently managed in the Nakadi project. This is different to how common models like Money and Problems are handled.

The first step proposed is to reduce the number of references to the technology project without changing the basic guidance - this issue tracks that work. Follow on issues can look at inverting the dependency (eg having the data models embedded in the guidelines or a public web page) such that the Nakadi software project is informed by the guidelines rather than the other way around.

See #230

tfrauenstein:
Guild meeting: let's close this issue -- all changes have been done; there is only one open issue as work in progress #265

PR #257, opened 2017-06-23
Requires Event Type schema conform to Open API's Schema Object.

dehora:
This adds a guideline to ensure events use the subset of the JSON-Schema 04 keywords that are allowed by Open API 2.0. Open API uses a subset of JSON-Schema and adds further options. The aim is to keep events aligned with API resources in general.

I pulled this out of another, larger branch on adding standard event types from Nakadi (see #258), as it's probably worth discussing on its own, and can be merged as a standalone pr. 

For #237.

### DIFF
--- a/events/events.md
+++ b/events/events.md
@@ -10,6 +10,60 @@ Events are part of a service’s interface to the outside world equivalent in st
 
 Services publishing event data for use by others must make the event schema available for review. 
 
+## {{ book.must }} Ensure Event Type schemas conform to Open API's Schema Object
+
+Event type schema are defined in accordance with the Open API Schema Object
+specification, which uses a subset of
+[JSON Schema Draft 4](http://json-schema.org/), and also adds other
+features. This allows events to properly align with API resource representations
+(it's particulary useful for events that represent data changes about
+resources). The guideline exists since declaring event type schema using
+JSON-Schema syntax is common practice (because Open API doesn't yet allow
+standalone object definitions).
+
+In the rest of this section we'll call out some of the more notable differences
+between Open API and JSON-Schema. Please note this is not a complete list - in
+general it's recommended to familiarise yourself with Open API's Schema Object
+(the details are available from the
+["Schema Object" section of that specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schemaObject)).
+
+Open API _removes_ some JSON-Schema keywords. These must not be used in event
+type schemas. The list of Open API object keywords can be seen
+[here](https://github.com/OAI/OpenAPI-Specification/blob/master/schemas/v2.0/schema.json#L935-L1063),
+but for convenience the list of non-available keywords relative to JSON-Schema
+are:
+
+  - `additionalItems`
+  - `contains`
+  - `patternProperties`
+  - `dependencies`
+  - `propertyNames`
+  - `const`
+  - `not`
+  - `oneOf`
+
+Open API _redefines_ some keywords:
+  
+  - `additionalProperties`: For event types that declare compatibility
+    guarantees, there are recommended constraints around the use of this
+    field. See the guideline "Avoid `additionalProperties` in event type
+    definitions" for details.
+
+Open API _extends_ JSON-Schema with some keywords:
+
+  - `readOnly`: events are logically immutable, so `readOnly` can be considered
+    redundant, but harmless.
+
+  - `discriminator`: discriminators exist to support polymorphism and act as an 
+    alternative to `oneOf`. 
+    
+  - `^x-`: patterned objects in the form of
+    [vendor extensions](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#vendorExtensions)
+    can be used in event type schema, but it might be the case that general
+    purpose validators do not understand them to enforce a validation check, and
+    fall back to must-ignore processing. A future version of the guidelines may
+    define well known vendor extensions for events.
+
 ## {{ book.must }} Ensure that Events define useful business resources
  
  Events are intended to be used by other services including business process/data analytics and monitoring. They should be based around the resources and business processes you have defined for your service domain and adhere to its natural lifecycle (see also "Should: Define useful resources" in the [General Guidelines](../general-guidelines/GeneralGuidelines.md)).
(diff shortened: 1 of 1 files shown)


events/events.md
```diff
@@ -10,6 +10,80 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+## {{ book.should }} Avoid remote references in Event Type schema
+
+The `$ref` keyword should not use remote references in event type schema -
+schemas should be self-contained and use local references.
+
+This is a consideration for agents in a distributed publish/subcribe system that
+have to process events asynchronously, outside the typical request-response
+cycle of an API, and also for simplifying operations like long term storage of
+events, where schema access is needed long after the event was published. In
+both cases (among others) having a single, self-describing schema simplfies
```
ePaul: `simplfies` → `simplifies`.

events/events.md
```diff
@@ -10,6 +10,80 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+
+  - `readOnly`: events are logically immutable, so `readOnly` can be considered
+    redundant, but harmless.
+
+  - `discriminator`: discriminators if used, should be published with full
+    information in the schema. See the guideline "Avoid remote references in
+    Event Type schema" for details.
+
+  - `xml`: must not be used for events. Events are required to be JSON as per
+    the general API guidelines.
```
ePaul: I guess that we don't want XML for our REST-like APIs either. Not sure whether that needs to be mentioned here, though.

events/events.md
```diff
@@ -10,6 +10,80 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+    definitions" for details.
+
+Open API _extends_ JSON-Schema with some keywords:
+
+  - `readOnly`: events are logically immutable, so `readOnly` can be considered
+    redundant, but harmless.
+
+  - `discriminator`: discriminators if used, should be published with full
+    information in the schema. See the guideline "Avoid remote references in
+    Event Type schema" for details.
```
ePaul: Hmm, I think normally the discriminator in OpenAPI is kind of a replacement for `oneOf` (which is not available here).
I'm not sure how this would work with JSON-schema validation for events.

events/events.md
```diff
@@ -10,6 +10,80 @@ Events are part of a service’s interface to the outside world equivalent in st
 
 Services publishing event data for use by others must make the event schema available for review. 
 
+## {{ book.must }} Ensure Event Type schema conform to Open API's Schema Object
```
tkrop: `conform` -> `confirms`?

ePaul: No, that verb has a different meaning. (And it is correctly plural here.)

events/events.md
dehora: I'd like to see an example where it wouldn't work before dropping it. As a datapoint `oneOf` is considered workable for compatibility in the Nakadi project. It allows full JSON-Schema (and that variation between what's possible with an event and api representations is a driver for this pr). but some keywords had to be dropped to get sensible compatibility akin to Avro or Protocol Buffers - they are:  `not`, `patternProperties`, `additionalProperties` and `additionalItems`

rcillo:
👍 

Just would like to share my opinion the biggest issue with json-schema which is not addressed in OpenAPI either.

Json schema defines that when `additionalProperties` is absent, schema validation should behave as if `additionalProperties: {}`, e.g. an additionalProperties with an empty schema.

Empty schema is considered to be a wildcard that accepts everything. At least in the library used by Nakadi to validate events https://github.com/everit-org/json-schema/blob/4951bc18c2ad6d21e5f6611f765e8749e148fdfa/core/src/main/java/org/everit/json/schema/EmptySchema.java#L45-L48

So, the biggest problem is not with the restriction of attributes, but with the fact that json-schema is an "open" model.

The only way to make a schema definition "closed", e.g. it does not take any attributes that are not defined, is by explicitly requiring `additionalProperties: false` on each single nested object definition.

Once more, json-schema is open by design. It means that everyone doing `additionalProperties: true` don't actually know what they are doing because they could be doing just `{}`.

So, even if we are constraining the usage of certain keywords, the problem of json-schema/OpenAPI is not their keywords, but the fact that they are "open" by design.

ePaul:
@rcillo For the "open schema" thing, have a look at #248 (though that is for REST APIs).

I think there is a difference in intention between writing `additionalProperties: true` (or `additionalProperties:
 {}`) and not mentioning `additionalProperties` at all – in the first case it explicitly says "there are other properties (of unspecified form), you should probably look at them" (or, in the case of server input, "I'll look at them") and in the second case, you are just letting things open for future API extensions.

rcillo:
@ePaul I agree and understand the distinction. But from the point of view of someone who spends a lot of time doing support for Nakadi users, I can tell that the impact is the same:

- accidentally defined no-op schemas that accept all events
- very poorly defined schemas due to the fact that definitions are open and there is no need to fully describe data in order to publish it
- very hard to implement "closed" validators, since there are corner cases where introducing a "closed" definition breaks existent functionality, like "allOf" (one one of the objects in the allOf array is closed, no events are ever valid if these definitions are disjunct).

My point is that for API definition it helped a lot and showed to be extremely valuable. But for event validation, I think it has done a pretty bad job so far.

events/events.md
tkrop: sorry typo: `conform` -> `conforms`?

tkrop:
👍

events/events.md
```diff
@@ -10,6 +10,60 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+  - `const`
+  - `not`
+  - `oneOf`
+
+Open API _redefines_ some keywords:
+  
+  - `additionalProperties`: For event types that declare compatibility
+    guarantees, there are recommended constraints around the use of this
+    field. See the guideline "Avoid `additionalProperties` in event type
+    definitions" for details.
```
tfrauenstein: This seems to be in conflict with the referenced guideline which deprecated additionalProperties usage.
I think we should be clearer about additionalProperties non/usage; my understanding: 
additionalProperties=false does not mean, that next schema version evolution may not include new attributes -- what we want support and is the reason why consumers should be robust. 
additionalProperties=false means: payload may not include attributes not explicitly defined by the schema.
This is actually what we want to achieve, but it requires schema updates to be registered before producer writes new attributes (and mandatory new fields require 2-step schema updates).

dehora: I don't see the conflict, but the point is re-emphasised in 714b76bed483cf82be4dea70aebc836d7663c8f6 in the referenced guideline.

tkrop: I agree with @dehora . I don't see a need for change here. We do currently do not have a Open API specific explanation of keywords and parameters, that could include this explanation. My first address would be the compatibility section than.

tkrop:
@dehora can you check my suggestion for the heading? Than I would like to move on and merge, to reduce our backlog of open pull requests.

@tfrauenstein @ePaul any objections?

dehora:
> can you check my suggestion for the heading? 

@tkrop pluralised "schemas", it's good to go now.

tkrop:
👍

tkrop:
@epaul @tfrauenstein  anyone else to give his thumbs up?

ePaul:
:+1:

PR #258, opened 2017-06-23
Adds Nakadi event categories and event types to the guidelines.

dehora:
Adds Nakadi event categories and event types to the guidelines.

This extracts the event categories and event types defined in the Nakadi into
the guidelines with the intent of allowing the API guidelines to govern and

evolve these formats as de-facto standards for events. Nakadi itself becomes the
reference implementation of the event formats as well as their registration and
validation. Previously the guidelines referenced the Nakadi project's API
definition, which by having design guidelines depend on a technology project was
"the wrong way around". This change allows the API guidelines to inform
technology, evolve the event model over time, and have other services, libraries
and tools draw from guidelines for their model of events.

Some notes:

- Because the guidelines are now responsible for event data models, a brief
  summary of the main concepts is added to the preliminary section.

- Certain information in the event type is specific to Nakadi (such as
  metrics). This has not been brought into the guidelines, but the event type
  format is declared open for extension.

- The Nakadi Business Category has been renamed to the General
  Category. It turns out "business process" was a bit fuzzy as a concept, and
  also teams use its structure as a general purpose "event". Hence the category
  was renamed to reflect how it's being used. A note on the old name has been
  added. About 90% of Nakadi's registered event types within Zalando use the
  business process category.

- The undefined event category has not been carried forward into the guidelines
  since it's not strictly speaking, an event. There's no place to declare
  metadata for example, and an undefined event is really just an object
  conforming to a schema. The undefined event still exists in Nakadi and there
  are no plans to remove it, because it's useful for some usecases (as a
  datapoint, less than 5% of Nakadi's registered even types within Zalando are
  declared undefined).

- In all cases, the object definitions here are taken as-is from Nakadi with no
  changes in structure or semantics. Practically speaking this is "as you were"
  with respect to the impact guidelines have on day to day work with events and
  the aim also is not require major work in Nakadi (the single exception would
  be to support the category name "general" as well as the previous name
  "business").

The guidelines will evolve event structures in the future as there is a desire to
do so, but for now the goal is to have events as they are defined and used today,
as part of the guidelines.

For #237.

### DIFF
--- a/events/events.md
+++ b/events/events.md
@@ -2,6 +2,31 @@
 
 Zalando’s architecture centers around decoupled microservices and in that context we favour asynchronous event driven approaches. The guidelines in this section focus on how to design and publish events intended to be shared for others to consume. 
 
+**Events, Event Types and Categories.**
+
+Events are defined using an item called an _Event Type_. The Event Type allows
+events to have their structure declared  with a schema by producers and understood by
+consumers. An Event Type declares standard information, such as a name, an owning
+application (and by implication, an owning team), a schema defining the event's
+custom data, and a compatibility mode declaring how the schema will be
+evolved. Event Types also allow the declaration of validation and enrichment
+strategies for events, along with supplemental information such as how events
+can be partitioned in an event stream.
+
+Event Types belong to a well known _Event Category_ (such as a data change
+category), which provides extra information that is common to that kind of
+event.
+
+Event Types can be published and made available as API resources for teams to
+use, typically in an _Event Type Registry_. Each event published can then be
+validated against the overall structure of its event type and the schema for its
+custom data.
+
+The basic model described above was originally developed in the
+[Nakadi project](https://github.com/zalando/nakadi), which acts as a reference
+implementation of the event type registry, and as a validating publish/subscribe
+broker for event producers and consumers.
+
 ## {{ book.must }} Treat Events as part of the service interface
 
 Events are part of a service’s interface to the outside world equivalent in standing to a service’s REST API. Services publishing data for integration must treat their events as a first class design concern, just as they would an API. For example this means approaching events with the "API first" principle in mind [as described in the Introduction](../Introduction.md).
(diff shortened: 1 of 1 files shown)


events/events.md
```diff
@@ -10,28 +35,327 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+          of forward compatible event types can safely read events tagged with
+          the latest schema version as long as they follow the robustness
+          principle.
+
+          - 'none': Any schema modification is accepted, even if it might break
+          existing producers or consumers. When validating events, no additional
+          properties are accepted unless explicitly stated in the schema.
+          
+        type: string
+        default: forward
```
ePaul: Maybe add the `x-extensible-enum with the three values here too?

events/events.md
```diff
@@ -10,28 +35,327 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+          following json-schema attributes are not supported: `not`,
+          `patternProperties`, `additionalProperties` and
+          `additionalItems`. When validating events, additional properties is
+          `false`.
+          
+          - 'forward': Compatible schema changes are allowed. It's possible to
+          use the full json schema specification for defining schemas. Consumers
+          of forward compatible event types can safely read events tagged with
+          the latest schema version as long as they follow the robustness
+          principle.
```
ePaul: Should that link to our compatibility guideline section?

events/events.md
```diff
@@ -10,28 +35,327 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+<pre style="font-size: .85em">
+  {
+    <span style="font-weight: bold;">"metadata" : {...}</span>
+    <span style="color: blue;">"a": "a1",</span>
+    <span style="color: blue;">"b": "b1"</span>
+  }
+</pre>
+
+Note: the General Event in a previous version of the guidelines was called a
+_Business Process Event_.  Implementation experience has shown that the
```
ePaul: Actually it was just "Business Event", I think.

events/events.md
```diff
@@ -10,28 +35,327 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+      data:
+        description: |    
+          Contains custom payload for the event type. The payload must conform
+          to a schema associated with the event type declared in the metadata
+          object's `event_type` field.                
+        type: object
+      data_type:
+        description: |
+          This field is a duplicate of `event_type` defined in the `metadata` 
+          object. It is deprecated, but exists for legacy compatibility reasons.
```
ePaul: What exactly does "deprecated" mean for a required property?

events/events.md
```diff
@@ -10,28 +35,327 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+The General and Data Change categories share a common structure for
+_metadata_. The metadata structure is shown below as an Open API Schema Object:
+
+```yaml
+  EventMetadata:
+    type: object
+    description: | 
+      Carries metadata for an Event along with common fields. The required
+      fields are those expected to be sent by the producer, other fields may be
+      added by intermediaries such as publish/subcribe broker.
```
ePaul: `subcribe` → `subscribe`

events/events.md
```diff
@@ -10,28 +35,327 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+
+- Data Change Event: a category used for describing changes to data entities.
+
+The set of categories is expected to evolve in the future.
+
+A category describes a predefined structure that event publishers must conform
+to along with standard information about that kind of event (such as the
+operation for a data change event).
+
+**The General Event Category.**
```
ePaul: Just a thought: Instead of having the definitions of the events in this guideline, put them in the specific guidelines about general events or data change events below?

events/events.md
dehora: It's taken as is from Nakadi. I'd propose changing it in another pr, with the maintainer's consensus.

events/events.md
dehora: Probably not needed from within the open api `description`

events/events.md
dehora: Right, not meaningful, removed the comment.

events/events.md
dehora: That would mean the metadata block used by both of them needs to be repeated in each section, added as its own guideline, or left where it is standalone. I haven't figured out how best to do that.

events/events.md
dehora: Pulling out the modes to the main text solves this.

PetrGlad:
@ePaul Context information looks like a metadata. Maybe it belongs there?

BGehrels:
@PetrGlad Somehow yes, but then data_op would be metadata, too. I would like to think more in terms of "envelope" and "body", with `data: {username: 'BGehrels'}` being the body. The top level json structure consists of different Mixins, one for nakadi transmission metadata (`metadata: {received_at: 'some_date'}`), one for data change event metadata (`data_op: 'C'`), one for business event metadata (`what_happened: "user_registered_itself`) and maybe more for other use cases.

This would relieve us from having to conform to exactly one event category (inheritence is evil :-) )

rcillo:
We had some arguments about renaming "business" to "general" and the implications for Nakadi. If I'm not mistaken, we should undo this renaming and preserve the original name "business", right?

dehora:
> we should undo this renaming and preserve the original name "business", right?

@rcillo we said we'd look at an alternative that left the name as is. It's here: https://github.com/dehora/restful-api-guidelines/commit/345ec83152aaa94fcf0a0bb7fd82cb71b7e89b7e. @ePaul, for context if the category name is changed to general that requires Nakadi to accept a patch that allows that name to be submitted rather than business. If business is left as the name, people may ask how they can send "just" an event (it turns out in reality the name "business" gets used for "just" events, ie "business" isn't accurate or meaningful so it's a questionable name to use).

events/events.md
```diff
@@ -10,28 +35,341 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+          The compatibility modes are:
+
+          - 'compatible': This means consumers can reliably process events
+            produced under different versions, such that every event published
+            since the first version is remains valid with respect to the newest
+            schema. When in compatible mode, it's allowed to add new optional
+            properties and definitions to an existing schema, but no other
+            changes are allowed. Under this mode, the following json-schema
+            attributes are not supported: `not`, `patternProperties`,
+            additionalProperties` and `additionalItems`.
```
tkrop: missing ``` ` ``` at at begin of `additionalProperties`

events/events.md
```diff
@@ -10,28 +35,341 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+be flexible enough so that producers can evolve schemas while not inadvertently
+breaking existing consumers:
+
+- 'compatible': Consumers can reliably process events produced under different
+  versions, such that every event published since the first version is remains
+  valid with respect to the newest schema. When in compatible mode, it's allowed
+  to add new optional properties and definitions to an existing schema, but no
+  other changes are allowed. Under this mode, the following json-schema
+  attributes are not supported: `not`, `patternProperties`,
+  additionalProperties` and `additionalItems`.
```
tfrauenstein: event payload schema definition should always be compliant to OpenAPI Object Schema -- independently from the compatibility mode selected. With PR #257 we now have a specific guideline here.

events/events.md
```diff
@@ -10,28 +35,341 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+
+- 'forward': Compatible schema changes are allowed and it's possible to use the
+  full json schema specification for defining schemas. Consumers of forward
```
tfrauenstein: 1) see comment above. 

2) Additionally, forward compatibility definition could be improved. 
Its not very clear what changes are supported with what effects for either producers or consumers. 

I also propose to use more industry known terms, e.g. "fully compatible" vs. "forward compatible" vs. "(potentially) incompatible" -- in the description, if not even in the enum value names.

events/events.md
```diff
@@ -10,28 +35,341 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+        description: Defines the category of this EventType. 
+        type: string
+        x-extensible-enum:
+          - data
+          - general
+      compatibility_mode:
+        description: |
+          The compatibility modes are:
+
+          - 'compatible': This means consumers can reliably process events
```
tfrauenstein: editorial:  
- compatibility mode description is 100% redundant to description above -- pls. just refer here "...see above"
- readable layout could be improved by using ~80 chars/line for descriptions

events/events.md
```diff
@@ -10,28 +35,341 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+      - schema    
+    properties:
+      name:
+        description: |
+          Name of this EventType.  Note: the name can encode the
+          owner/responsible for this EventType and ideally should follow a
+          common pattern that makes it easy to read and understand.
+        type: string
+        pattern: '[a-zA-Z][-0-9a-zA-Z_]*(\.[a-zA-Z][-0-9a-zA-Z_]*)*'
+        example: order.order_cancelled, acme-platform.users
```
tfrauenstein: editorial:  acme-platform.users --> business_partner.contract

events/events.md
```diff
@@ -10,28 +35,341 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+      - name
+      - category
+      - owning_application
+      - schema    
+    properties:
+      name:
+        description: |
+          Name of this EventType.  Note: the name can encode the
+          owner/responsible for this EventType and ideally should follow a
+          common pattern that makes it easy to read and understand.
```
tfrauenstein: editorial:  update to:
Name should follow common pattern 'functional-component'.'business event or entity' that makes it easy to read and understand.

events/events.md
```diff
@@ -10,28 +35,341 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+        type: string
+        pattern: '[a-zA-Z][-0-9a-zA-Z_]*(\.[a-zA-Z][-0-9a-zA-Z_]*)*'
+        example: order.order_cancelled, acme-platform.users
+      owning_application:
+        description: Indicator of the application owning this `EventType`.
```
tfrauenstein: editorial:  Name of the application (as also used in infrastructure application registry, like Kio) owning this 'EventType'

events/events.md
```diff
@@ -10,28 +35,341 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+        $ref: '#/definitions/EventMetadata'
+      data:
+        description: |    
+          Contains custom payload for the event type. The payload must conform
+          to a schema associated with the event type declared in the metadata
+          object's `event_type` field.                
+        type: object
+      data_type:     
+        type: string
+        example: 'pennybags:order'
```
duergner: That should be a more general example, rather than referencing a specific Zalando team

events/events.md
```diff
@@ -10,28 +35,341 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+        description: The most recent payload schema for this EventType. 
+        type: object
+        properties:
+          version:
+            description: | 
+               Values are based on semantic versioning. Changes to `title` or
+              `description` are considered PATCH level changes. Adding new
+              optional fields is considered a MINOR level change. All other
+              changes are considered
+              MAJOR level.
```
tfrauenstein: Should be related to compatibility modes:
- "fully compatible" changes lead to PATCH or MiNOR level changes.
- "forward" or "none" changes may also lead to MAJOR level changes.

events/events.md
```diff
@@ -10,28 +35,341 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+          created_at:
+            description: Creation timestamp of the schema. 
+            type: string
+            readOnly: true
+            format: date-time
+            example: '1996-12-19T16:39:57-08:00'
+          type:
+            description: |            
+               The schema language of schema definition. Currently only
+               json_schema (JSON Schema v04) is defined, but in the future there
```
tfrauenstein: Should be changed to "OpenAPI 2.0 Object Schema", shouldn't it?

events/events.md
```diff
@@ -10,28 +35,341 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+how events are partitioned in the stream.
+
+## {{ book.must }} Ensure Events conform to a well-known Event Category
+
+An _event category_ describes a generic class of event. The guidelines define
+two such categories:
+
+- General Event: a general purpose category.
+
+- Data Change Event: a category used for describing changes to data entities.
```
tfrauenstein: editorial:   ...describing changes to data entities.  --> ...describing changes to data entities used for data replication based data integration.

events/events.md
```diff
@@ -10,28 +35,341 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+
+The structure of the _General Event Category_ is shown below as an Open API
+Schema Object definition:
+
+```yaml
+  GeneralEvent:
+    description: |
+      A general kind of event. Event kinds based on this event define their
+      custom schema payload as the top level of the document, with the
+      "metadata" field being required and reserved for standard metadata. An
```
tfrauenstein: as discussed, I hope we can rename it to "generic_event_data"

events/events.md
tfrauenstein: editorial: 
- pennybags:order --> sales_order.order
- description: name of the (business) data entity that has been mutated

non-editorial: 
- Can we get rid of this mandatory field -- it is actually redundant to the event type name -- which should be the data entity name in case of data change events.

events/events.md
```diff
@@ -10,28 +35,341 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+        type: string
+        example: 'example.important-business-event'
+      occurred_at:
+        description: When the event was created according to the producer.
+        type: string
+        format: date-time
+        example: '1996-12-19T16:39:57-08:00'
+      received_at:      
+        description: |           
+          When the event was seen by an intermediary such as a broker.
```
tfrauenstein: editorial: When the event was delivered to the consumer.

events/events.md
```diff
@@ -10,28 +35,341 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+    <span style="font-weight: bold;">"metadata": {...}</span>
+    <span style="font-weight: bold">"data_op": "C"</span>
+    <span style="font-weight: bold">"data_type": "example.order"</span>
+    <span style="font-weight: bold">"data": {</span>
+      <span style="color: blue;">"a": "a1",</span>
+      <span style="color: blue;">"b": "b1"</span>
+    <span style="font-weight: bold">}</span>
+  }
+</pre>
+
```
tfrauenstein: editorial:  we also should position hint/link here to the data change specific guidelines below

events/events.md
```diff
@@ -10,28 +35,341 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+    <span style="color: blue;">"b": "b1"</span>
+  }
+</pre>
+
+Note: the General Event in a previous version of the guidelines was called a
+_Business Event_.  Implementation experience has shown that the
+category's structure gets used for other kinds of events, hence the name has
+been generalized to reflect how teams are using it. The General Event is still
+useful and recommended for the purpose of defining events that drive a business
+process.
```
tfrauenstein: editorial:  we also should position hint/link here to the general event specific guideline below

events/events.md
duergner: That would not be the same timestamp.

dehora: > When the event was delivered to the consumer.

Delivery to a consumer happens later (also confirmed as the current nakadi behaviour, as per [here](https://github.com/dehora/nakadi/blob/master/src/main/java/org/zalando/nakadi/enrichment/MetadataEnrichmentStrategy.java#L18-L18), which happens on the producer side).

tfrauenstein: ok, it was an assumption form my side in order to provide concrete proposal how to improve the description which I not found helpful from event spec user point of view because "broker" is a specific implementation term. Also delta to occured_at is unclear.

events/events.md
dehora: > How about removing an enum value from a list?

Would need to check that against the Nakadi implementation of the compatibility check; it's not a keyword level action - @rcillo might know what the current behaviour is.

dehora: > With PR #257 we now have a specific guideline here.

:+1: when #257 is merged and there's a link this can be added. My suggestion is to merge #257 first, but if we require Nakadi to enforce the keyword checks (as opposed to the requirement being on the design event type definition side initially) then both prs are held up on a software upgrade.

dehora: updated in ac69d826722ad9bea9446558bd31bf8a1ddd4b3d

events/events.md
dehora: updated in ac69d826722ad9bea9446558bd31bf8a1ddd4b3d

events/events.md
dehora: updated in ac69d826722ad9bea9446558bd31bf8a1ddd4b3d

> if not even in the enum value names.

While that's not technically a breaking change in the definition, because it's a string not an enum;  in reality the current names can't be removed or changed without impacting internal end users and we probably don't want a set of duplicate names. I'd suggest not making name level changes in this pr.

events/events.md
dehora: updated in ac69d826722ad9bea9446558bd31bf8a1ddd4b3d

events/events.md
dehora: updated in ac69d826722ad9bea9446558bd31bf8a1ddd4b3d

events/events.md
dehora: updated in ac69d826722ad9bea9446558bd31bf8a1ddd4b3d. Left out Kio as Zalando specific

events/events.md
dehora: updated in ac69d826722ad9bea9446558bd31bf8a1ddd4b3d. Removed the descriptions but no "see above" to avoid the yaml object pointing at surrounding text.

events/events.md
dehora: updated in ac69d826722ad9bea9446558bd31bf8a1ddd4b3d.

events/events.md
dehora: updated ac69d826722ad9bea9446558bd31bf8a1ddd4b3d. Moved this content into the guideline text as per the compatibility level descriptions

events/events.md
dehora: updated in . It mentions Open API but does not redefine what "json_schema" denotes

Again, I'd say #257 needs to be merged first as that aligns people defining event type schema. How that gets enforced or declared in the `"type"` field (eg defining a new type field or changing the exact meaning of the `"json_schema"` field) I'd keep out of scope of this pr.

events/events.md
dehora: updated in ac69d826722ad9bea9446558bd31bf8a1ddd4b3d

events/events.md
dehora: That's a breaking change for both existing categories - "metadata" is a required field.

events/events.md
dehora: updated in ac69d826722ad9bea9446558bd31bf8a1ddd4b3d

events/events.md
dehora: > pennybags:order --> sales_order.order
> description: name of the (business) data entity that has been mutated

updated in ac69d826722ad9bea9446558bd31bf8a1ddd4b3d

> Can we get rid of this mandatory field 

That would be a breaking change to Nakadi's interface, it's required to be sent at the moment for both categories.

events/events.md
dehora: updated in ac69d826722ad9bea9446558bd31bf8a1ddd4b3d

events/events.md
dehora: The exact text from Nakadi is:

```
Timestamp of the reception of the Event by Nakadi. This is 
enriched upon reception of the Event.
```

> "broker" is a specific implementation term. 

Agreed. I'm unsure how to proceed here without some reference to middleware like Nakadi to avoid confusing it with the local consumer's received time. One option is to remove the field as it's optional data, but that changes the compilation target for a raft of clients being used internally (eg that information could be dropped for any consumer using swagger codegen which is fairly popular).

BGehrels: I guess the root cause of this is that we mix concerns here: we are mixing transport level metadata and application metadata into the same object spec.

I would propose to leave all attributes that are set by a broker out of this spec and just write: 
> brokers may add arbitrary additional fields that are not specified here and may set default values, if some of the specified fields are not given. Please look into your message brokers documentation for additional information.

Yes, this leads to broker specfic metadata fields not being generated by swagger codegen out of the box, but i think this is okay: They can always be added manually, either to the generated files, or to your applications swagger yaml. And most of the brokers on the market i know do not provide a field called `metadata` anyway.

dehora: > I would propose to leave all attributes that are set by a broker out of this spec 

So we're clear, that would include:

- `version`
- `flow_id`

Flow ID I personally can live with but it has a genuine operational impact if it becomes lossy. Version is more problematic for a data standpoint, that's needed by a class of consumers that perform event capture for things like storage, it should be a first class element in this design, but we can't require producers to send it.

As a side note, two things:

1. The goal in this pr, is not to make Nakadi invisible, eliminate the concept of enriching brokers, or "fix" existing structures - it's to adjust and partner around the governance of event types.

2. We need to be very careful about any kind of redefinition of meaning or real world processing behaviour for _individual events_, however sensible that may seem .


If the guild members can't live with those pr constraints, I think we need to reconsider the overall approach. Redefinition and long term improvement happens through evolution, which is almost certainly going to evolve a new common JSON structure and rethink on how we decalre categories.

dehora: > brokers may add arbitrary additional fields that are not specified here and may set default values, if some of the specified fields are not given. Please look into your message brokers documentation for additional information.

:+1: took a variation of this and added it in ce989f78c32e1ad796848c972c38c67c2a1b3de7.

BGehrels: Regarding flow_id: Since https://github.com/zalando/nakadi/pull/677 this is actually a field that can be specified by the sender and will only be enriched, if it is missing. so keeping it in here seems like a good idea

dehora:
> I would like to ask, whether we really like to add event categories at this stage into the guideline. 

I think we do (and we've discussed this :). The formats are very heavily used.

> As the concept is broken, 

Broken is a strong term; the problems are the dependency direction from guidelines to project (hence this pr), along with the ability to evolve, which is another round of work that will happen (ideally as a specification the guidelines can reference). Having these managed at the right level in the meantime will help with that imvho.

events/events.md
```diff
@@ -10,28 +35,366 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+_metadata_. The metadata structure is shown below as an Open API Schema Object:
+
+```yaml
+  EventMetadata:
+    type: object
+    description: | 
+      Carries metadata for an Event along with common fields. The required
+      fields are those expected to be sent by the producer, other fields may be
+      added by intermediaries such as publish/subscribe broker.      
+    required:
```
duergner: I would suggest to an additional required property `is_sandbox` here to indicate whether this event is production or not in order to allow for *testing in production*.

With Nakadi we should set this `is_sandbox` field by the producer based on the *business_partner* from the OAuth2 token (in case we go with the approach to distinguish between production and sandbox/testing on business partner level from IAM perspective.

events/events.md
tfrauenstein: 👍  as discussed, next step event format redesign issues are out of scope of this PR.

events/events.md
```diff
@@ -10,28 +35,366 @@ Events are part of a service’s interface to the outside world equivalent in st
...
+        type: string
+        example: 'example.important-business-event'
+      occurred_at:
+        description: When the event was created according to the producer.
+        type: string
+        format: date-time
+        example: '1996-12-19T16:39:57-08:00'
+      received_at:      
+        description: |           
+          When the event was seen by an intermediary such as a broker.
```
tfrauenstein: As discussed in the API Guild meeting, we should keep the original scope of this PR in mind:

1) in scope: definition of event format as it is currently applied and Nakadi supported.
Decouple it from Nakadi and improve specification but w/o breaking changes.
Allowing the API guidelines to govern and evolve these formats as de-facto standards for events. 
2) not in scope: definition of next event format generation with breaking changes.

The event format evolution will be a longer term story where we have to re-challenge and re-align on the "category" concept and the relation between event schemes and other metadata need for data governance & integration at Zalando. Even if we focus on getting rid of the most obvious event format design flaws, it will include a couple of incompatible changes, which requires careful discussion, migration plan and stakeholder alignment -- most important with Nakadi team and API Guild. This is outside the scope of a PR. 

Hence, let us make progress and proceed incrementally:

1) merge this PR -- don't throw it away and wait weeks for breaking change alignment

2) next generation low hanging fruits: get rid of current design flaws. Here Bill (and me) will provide solution concept next week -- will be the base of a second PR once aligned and in sync with Nakadi delivery. 

@dehora pls. confirm, when all improvements respecting status quo and non breaking constraint have been executed.

events/events.md
```diff
@@ -64,28 +89,370 @@ Open API _extends_ JSON-Schema with some keywords:
...
+In the example fragment below, the reserved metadata field is shown with fields
+"a" and "b" being defined as part of the custom schema:
+
+<pre style="font-size: .85em">
+  {
+    <span style="font-weight: bold;">"metadata" : {...}</span>
+    <span style="color: blue;">"a": "a1",</span>
+    <span style="color: blue;">"b": "b1"</span>
+  }
+</pre>
```
dehora: Looks like this:

screen shot 2017-08-11 at 13 10 46

events/events.md
```diff
@@ -64,28 +89,370 @@ Open API _extends_ JSON-Schema with some keywords:
...
+  {
+    <span style="font-weight: bold;">"metadata": {...}</span>
+    <span style="font-weight: bold">"data_op": "C"</span>
+    <span style="font-weight: bold">"data_type": "example.order"</span>
+    <span style="font-weight: bold">"data": {</span>
+      <span style="color: blue;">"a": "a1",</span>
+      <span style="color: blue;">"b": "b1"</span>
+    <span style="font-weight: bold">}</span>
+  }
+</pre>
```
dehora: Looks like this when published:

screen shot 2017-08-11 at 13 10 56

dehora:
Ok, some edits have been made and rebased in deed039. Again the purpose of the pr is not to fix or improve event structures, it's to capture event definitions as they are today, as part of the guidelines. We will evolve event structures in the future.

events/events.md
```diff
@@ -64,28 +89,369 @@ Open API _extends_ JSON-Schema with some keywords:
...
+        example: 'JAh6xH4OQhCJ9PutIV_RYw'
+      partition:
+        description: |
+          Indicates the partition assigned to this Event. Used for systems where
+          an event type's events can be sub-divided into partitions.          
+        type: string
+        example: '0'
+```
+
+Please note than intermediaries acting between the producer of an event and its ulimate consumers, may perform operations like validation of events and enrichment of an event's `"metadata"`. For example brokers such as Nakadi, can validate and enrich events with arbitrary additional fields that are not specified here and may set default or other values, if some of the specified fields are not supplied. How such systems work is outside the scope of these guidelines but producers and consumers working with such systems should be look into their documentation for additional information.
```
tkrop: `should be look` -> `should look`

events/events.md
```diff
@@ -64,28 +89,369 @@ Open API _extends_ JSON-Schema with some keywords:
...
+          Name of this EventType.  Note: the name can encode the
+          owner/responsible for this EventType and ideally should follow the
+          common pattern 'functional-component'.'business event or entity' that
+          makes it easy to read and understand.
+        type: string
+        pattern: '[a-zA-Z][-0-9a-zA-Z_]*(\.[a-zA-Z][-0-9a-zA-Z_]*)*'
+        example: order.order_cancelled, business_partner.contract
+      owning_application:
+        description: |
+          Name of the application (eg, as would be used in infrastructure
```
tkrop: Ignore this. I will update this in a followup PR.

events/events.md
```diff
@@ -64,28 +89,369 @@ Open API _extends_ JSON-Schema with some keywords:
...
+      - name
+      - category
+      - owning_application
+      - schema    
+    properties:
+      name:
+        description: |
+          Name of this EventType.  Note: the name can encode the
+          owner/responsible for this EventType and ideally should follow the
+          common pattern 'functional-component'.'business event or entity' that
```
tkrop: Ignore this. I will update this in a followup PR.

tkrop:
Bill is in vacation, so lets merge and make editorial changes afterward.

tkrop:
👍

maxim-tschumak:
:+1:

PR #261, opened 2017-06-27
Add node version hint

### LINKED ISSUE #231: Build doesn't work locally with recent node/npm versions

fmueller:
Closes #231 

As proposed, this adds hint about node version to use until we upgrade to current Gitbook version.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,5 +1,9 @@
 # Build
 
+## Requirements
+
+You have to use node version 5.1.1 until we upgrade to current Gitbook version.
+
 ## Install all dependencies
 
     npm install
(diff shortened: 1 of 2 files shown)


dehora:
👍

maxim-tschumak:
:+1:

PR #265, opened 2017-07-04
Avoids remote references in Event Type schema.

dehora:
This avoids remote references breaking the compatibility assurances available in event types, as the remote reference  is not under direct control of the system managing the event  types.

For #237.

### DIFF
--- a/events/events.md
+++ b/events/events.md
@@ -227,3 +227,20 @@ Event consumers must be able to process duplicate events.
 Most message brokers and data streaming systems offer “at-least-once” delivery. That is, one particular event is delivered to the consumers one or more times. Other circumstances can also cause duplicate events.
 
 For example, these situations occur if the publisher sends an event and doesn't receive the acknowledgment (e.g. due to a network issue). In this case, the publisher will try to send the same event again. This leads to two identical events in the event bus which have to be processed by the consumers. Similar conditions can appear on consumer side: an event has been processed successfully, but the consumer fails to confirm the processing.
+
+
+## {{ book.should }} Avoid remote references in Event Type schema
+
+The `$ref` keyword should not use remote references in event type schema -
+schemas should be self-contained and use local references.
+
+This is a consideration for agents in a distributed publish/subcribe system that
+are temporally and spatial decoupled. Consumers might have to process events 
+asynchronously, outside the typical request-response cycle of an API, and might 
+do so long after the event was published by a producer. One example is historial 
+processing of event data from long term storage, where schema access is needed  
+after the event was published. Having a single, self-contained schema simplifies 
+consumer designs and avoids transitive changes to remote schema that are not 
+directly managed as part of the event type, or schema that are no longer located 
+at a given url. Remote references allow the compatibility assurances available in
+event types to be broken.


whiskeysierra:
They should deference them and persist local references, in my opinion.

dehora:
> They should deference them and persist local references, in my opinion.

That doesn't work cleanly:

- There's no compatibility control - what if the deref'd  schema is not compatible?
- There's no assurance of availability  - what if the deref'd  schema is 4xx or 5xx?
- What is the version of the local definition if the remote definition changes?
- For events, what should the actual schema look like, inlined?

This is a general design problem with Open API - remotes refs allow runtime redefinition of schema, which is very close to or maybe isomorphic to transitive dependencies in libraries. The challenges here are not great for API descriptions when the API documents are not in a closed system, and are highly problematic for events, which are strongly coupled to definitions.

whiskeysierra:
It should be dereferenced on creation (or future updates) of the event type definition. It should dereference it synchronously in the scope of the original request. Failure to do so results in a failure for the client trying to create or modify the event type (no modification will be applied).

That is easy to understand for clients and solves all the issues while it still allows for remote references, which are quite handy.

whiskeysierra:
And yes, the result should be inlined.

dehora:
> And yes, the result should be inlined.

This is effectively saying event schema should be statically linked. It's not clear doing that mechanically is a better idea than the owner doing it manually since the latter gives the owner full control (eg the way modern library systems use lock files to stop transitive breaks). And at which point the owner should probably just inline it, as nothing bad can happen.

>  It should be dereferenced on creation (or future updates) 
>...
> And yes, the result should be inlined.

This suggests some system should inline a cached copy and maintain a link reference as well - I don't think that can be expressed in Open API. One issue with the just inline approach is that it can't be specified without making assumptions about implementations.

Please keep the objections coming, I think it's important we have a view on this. For data imvho, reproducibility is critical, and any construct that allows accidental redefinition of schema is a problematic, which is to say, `ref` is a problem.

whiskeysierra:
> And at which point the owner should probably just inline it themselves

That is a possibility of course, but that would mean that my 100 line event definition might be bloated by another 1k lines because I depend on something else, e.g. JSON Schema itself or JSON Patch.

Inlining them locally would also work, but doing that automatically requires tooling that is also not available.

tkrop:
@dehora, @whiskeysierra: The cases where we use remote references today are very stable resources. Actually, I do not think that any of them semantically changed since we defined them externally.

But I agree, it would be a bad thing to link to changing resources.

Instead of avoiding them, I would suggest to make clear, that referenced resources are not allowed to change at all - any change requires to create a new permanent resource with a new semantic version number. Systems, are allowed to treat the references as permanent resources that can be infinitely cached by the processing systems. Changing a referenced resource is a violation of the API guideline.

To support this, I think our API management system, in the long term should provide a proxy service for resolving references. It should detect changes of resources, send violation e-mails to the resource owners, and ensure the consistency of API definitions by forwarding the unchanged version.

tkrop:
After our discussion in the last guild meeting, @tfrauenstein @maxim-tschumak and @mfellner continued the discussion for some time, and we agreed, that we have conceptually the same problem for REST APIs as for Event APIs. The vision we have in mind, that we intend to validate compatibility of REST APIs and request in the future to uncover risk of unilateral API changes.

We concluded, that all systems registering APIs (i.e. API storage, Nakadi) should be supplied with de-referenced API specifications, to keep systems simple and understandable. An alternative approach would have been to introduce versioned references, but it was discarded because of the inherent complexity to check the referenced definition against a registration database.

I think we also agreed, that rules for both types of APIs, therefore, should be in sync.

I'm currently not sure, if everyone shares my opinion, that reuse of definitions is a good concept. I like it because, reuse via references shortens APIs and makes reading of APIs a lot easier. I although like the Idea of having a repository or database of common versioned definitions for events hanging around.

**My suggestion**: let us find a common place for a rule that is formulated as follows:

```
## {{ book.must }} Not Use Remote References in Released Schema Definitions

While the reuse of a definition via `$ref` keyword improves the readability of an API, 
remote references are a problem when definition and application of the API is getting
temporally and spatial decoupled. If an API definition is only reassembled at runtime,
transitive changes of remote schema definitions that are not directly managed as part
of the API can break the expected contract of the service or event producer/consumer.

Thus, it is inevitable to provide a single, self-contained API schema when releasing or
registering an API schema. I.e. whenever an API is released within a container image
or registered with an event bus, the API schema has to be de-referenced upfront.
```

whiskeysierra:
> the API schema has to be de-referenced upfront

Will you provide tooling for this?

whiskeysierra:
> We concluded, that all systems registering APIs (i.e. API storage, Nakadi) should be supplied with de-referenced API specifications, to keep systems simple [..].

To rephrase this, it will  *keep the API storage and Nakadi simple*, but force **everybody else** to deal with inlining referenced schemas which introduce a complexity to all APIs that use them, e.g. makes it harder to read and update/maintain them.

tkrop:
@whiskeysierra While my suggestion is formulated a bit more restrictive than the original request, I think it is even the better approach, as the weak formulation would not solve the problem. Registration system have to reject APIs anyway, if they have to protect the consistency of results.

As far as I see, there is at least one tool available, but I assume other tools have even build in support:

https://www.npmjs.com/package/json-schema-ref-parser

ePaul:
*From the meeting*:

- for REST-APIs, it should be enough to use immutable URLs (like ones referring to Github commits or tags).
- for event definitions, the event type registry should do the inlining (and the state at the registration time counts).

ePaul:
This needs to be converted to AsciiDoc.

ePaul:
**Discussed again in the meeting.**

* We want to not have it in the guidelines.
* If Nakadi wants to prevent this, it is free to do so. If people complain about this, maybe that would be the time to discuss implementing the resolution in Nakadi.
* For Restful APIs, we don't write anything into the guidelines now. When the API repository wants to implement version comparisons, it needs to resolve the links, and store the results.

PR #267, opened 2017-07-06
Add guidance on precision of numbers

### LINKED ISSUE #235: Add guidance on large numbers in JSON and interoperability
JSON the specification doesn't provide a limit on the number production in its grammar but very long numbers or doubles with a long mantissa may lose precision when processed by common implementations (such as JavaScript). 

[rfc7159](https://tools.ietf.org/html/rfc7159) allow arbitrary long number productions in its grammar but says implementations may impose limitations without saying what they are. This tends to result in interoperability problems and disagreements on who's responsible for the issue.

tkrop:
Fixes #235

### DIFF
--- a/data-formats/DataFormats.md
+++ b/data-formats/DataFormats.md
@@ -36,7 +36,7 @@ Use the following standard formats for country, language and currency codes:
 
 ## {{ book.must }} Define Format for Type Number and Integer
 
-Whenever an API defines a property of type `number` or `integer`, the precision has to be defined by the format as follows to prevent clients from guessing the precision incorrect, and thereby changing the value unintentionally:
+Whenever an API defines a property of type `number` or `integer`, the precision must be defined by the format as follows to prevent clients from guessing the precision incorrectly, and thereby changing the value unintentionally:
 
 | type    | format  | specified value range                                 |
 |---------|---------|-------------------------------------------------------|
@@ -47,7 +47,7 @@ Whenever an API defines a property of type `number` or `integer`, the precision 
 | number  | double  | IEEE 754-2008/ISO 60559:2011 binary128 decimal number |
 | number  | decimal | arbitrarily precise signed decimal number             |
 
-The precision must be translated by clients and servers into the most specific language types. An Example of the correct usage can be found below:
+The precision must be translated by clients and servers into the most specific language types. E.g. for the following definitions the most specific language types in Java will translate to `BigDecimal` for `Money.amount` and `int` or `Integer` for the `OrderList.count`:
 
 ```yaml
 Money:
--- a/data-formats/DataFormats.md
+++ b/data-formats/DataFormats.md
@@ -34,7 +34,7 @@ Use the following standard formats for country, language and currency codes:
 
 * [ISO 4217 currency codes](https://en.wikipedia.org/wiki/ISO_4217)
 
-## {{ book.must }} Use Define Format for Type Number and Integer
+## {{ book.must }} Define Format for Type Number and Integer
 
 Whenever an API defines a property of type `number` or `integer`, the precision has to be defined by the format as follows to prevent clients from guessing the precision incorrect, and thereby changing the value unintentionally:
 
--- a/data-formats/DataFormats.md
+++ b/data-formats/DataFormats.md
@@ -47,7 +47,7 @@ Whenever an API defines a property of type `number` or `integer`, the precision 
 | number  | double  | IEEE 754-2008/ISO 60559:2011 binary128 decimal number |
 | number  | decimal | arbitrarily precise signed decimal number             |
 
-The precision must be translated by clients and servers into the most specific language types. E.g. for the following definitions the most specific language types in Java will translate to `BigDecimal` for `Money.amount` and `int` or `Integer` for the `OrderList.count`:
+The precision must be translated by clients and servers into the most specific language types. E.g. for the following definitions the most specific language types in Java will translate to `BigDecimal` for `Money.amount` and `int` or `Integer` for the `OrderList.page_size`:
 
 ```yaml
 Money:
@@ -60,10 +60,10 @@ Money:
       example: 99.95
    ...
 
-OrdersList:
+OrderList:
   type: object
   properties:
-    count:
+    page_size:
       type: integer
       description: Number of orders in list
       format: int32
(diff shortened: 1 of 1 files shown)


data-formats/DataFormats.md
```diff
@@ -34,6 +34,42 @@ Use the following standard formats for country, language and currency codes:
 
 * [ISO 4217 currency codes](https://en.wikipedia.org/wiki/ISO_4217)
 
+## {{ book.must }} Use Define Format for Type Number and Integer
```
ePaul: grammar: "Use define Format" → "Use Defined Format"?  Or just "Define Format"?

data-formats/DataFormats.md
```diff
@@ -34,6 +34,42 @@ Use the following standard formats for country, language and currency codes:
...
+| type    | format  | language types |
+|---------|---------|----------------|
+| integer | int32   | int, Integer   |
+| integer | int34   | long, Long     |
+| integer | bigint  | BigInteger     |
+| number  | float   | float, Float   |
+| number  | double  | double, Double |
+| number  | decimal | BigDecimal     |
+
+The precision must be translated by clients and servers into the most specific language types. An Example of the correct usage can be found below:
```
ePaul: Do we actually want to prescribe how a server/client is implementing this?
What do we want to reach here?

The format has two goals:
* For integer, `int32` and `int64` guarantee/require that the value is in some specific range (e.g. between -2<sup>31</sup> and 2<sup>31</sup>-1 for int32), while `bigint` tells that the value can be arbitrarily large.
* For number, `float` and `double` tell us that the value originally is a binary floating point number fitting into the corresponding IEEE 754 representation, then converted to a decimal representation for transfer. `decimal` just tells us to take the (decimal formatted) number as it is.

A client or server could just handle it as a string without violating that goal, so requiring them to translate it to a specific numeric type is wrong. (Even though most ones will be doing it anyways.)

data-formats/DataFormats.md
```diff
@@ -34,6 +34,42 @@ Use the following standard formats for country, language and currency codes:
 
 * [ISO 4217 currency codes](https://en.wikipedia.org/wiki/ISO_4217)
 
+## {{ book.must }} Use Define Format for Type Number and Integer
+
+Whenever an API defines a property of type `number` or `integer`, the precision has to be defined by the format as follows to prevent clients from guessing the precision incorrect:
+
+| type    | format  | language types |
```
ePaul: The "language types" listed here are basically Java types? Why not call it this way?
Maybe in addition we could have a column with an language-independent definition (or reference to such one)?

ePaul:
Hmm, I meant my bigger comment to have the "needs changes" mark, not the small one.

data-formats/DataFormats.md
tkrop: I give it a try, but I disagree with

> A client or server could just handle it as a string without violating that goal, so requiring them to translate it to a specific numeric type is wrong. (Even though most ones will be doing it anyways.)

If defined, it is wrong for servers and clients to create or accept decimal numbers that are not compliant to the IEEE 754-2008/ISO 60559:2011. Using strings as represent this value, is a violation of the specification in my opinion, because the implementation will not behave according to the specification.

data-formats/DataFormats.md
tkrop: Yes, I know. I thought about adding the keyword Java to them. I found it difficult to do language independent definition, but I give it a try.

ePaul:
:+1:

data-formats/DataFormats.md
```diff
@@ -34,6 +34,42 @@ Use the following standard formats for country, language and currency codes:
...
+| type    | format  | specified value range                                 |
+|---------|---------|-------------------------------------------------------|
+| integer | int32   | integer between -2<sup>31</sup> and 2<sup>31</sup>-1  |
+| integer | int64   | integer between -2<sup>63</sup> and 2<sup>63</sup>-1  |
+| integer | bigint  | arbitrarily large signed integer number               |
+| number  | float   | IEEE 754-2008/ISO 60559:2011 binary64 decimal number  |
+| number  | double  | IEEE 754-2008/ISO 60559:2011 binary128 decimal number |
+| number  | decimal | arbitrarily precise signed decimal number             |
+
+The precision must be translated by clients and servers into the most specific language types. An Example of the correct usage can be found below:
```
dehora: > An Example

"""
An example
"""

data-formats/DataFormats.md
```diff
@@ -34,6 +34,42 @@ Use the following standard formats for country, language and currency codes:
 
 * [ISO 4217 currency codes](https://en.wikipedia.org/wiki/ISO_4217)
 
+## {{ book.must }} Define Format for Type Number and Integer
+
+Whenever an API defines a property of type `number` or `integer`, the precision has to be defined by the format as follows to prevent clients from guessing the precision incorrect, and thereby changing the value unintentionally:
```
dehora: > the precision has to be defined

"""
must be defined
"""

> to prevent clients from guessing the precision incorrect

"""
to prevent clients from guessing the precision incorrectly
"""

data-formats/DataFormats.md
dehora: > The precision must be translated by clients and servers into the most specific language types.

What does most "specific language type" mean? The example in the follow-on sentence is an example for something else.

tkrop: Thanks, I improve this.

data-formats/DataFormats.md
```diff
@@ -34,6 +34,42 @@ Use the following standard formats for country, language and currency codes:
...
+  type: object
+  properties:
+    amount:
+      type: number
+      description: Amount expressed as a decimal number of major currency units
+      format: decimal
+      example: 99.95
+   ...
+
+OrdersList:
```
whiskeysierra: The text above uses *Order**s**List*.

data-formats/DataFormats.md
```diff
@@ -34,6 +34,42 @@ Use the following standard formats for country, language and currency codes:
...
+  type: object
+  properties:
+    count:
```
whiskeysierra: The pagination section calls this property `page_size`. Would be nice to be consistent here.

whiskeysierra:
:+1:

dehora:
:+1:

ePaul:
:+1:

PR #268, opened 2017-07-12
Enforce snake_case even for IANA link rel types #264

### LINKED ISSUE #264: Property naming convention is incompatible with IANA link relations
Since the latest changes to hypertext controls there is an incompatible corner case with the following two guidelines: 
- [Property names must be snake_case (and never camelCase).](https://github.com/zalando/restful-api-guidelines/blob/master/json-guidelines/JsonGuidelines.md#-bookmust--property-names-must-be-snake_case-and-never-camelcase)
- [Use common hypertext controls](https://github.com/zalando/restful-api-guidelines/blob/master/hyper-media/Hypermedia.md#-bookmust--use-common-hypertext-controls)

> The name of an attribute holding such a HttpLink object specifies the relation between the object that contains the link and the linked resource. Implementations should use names from the IANA Link Relation Registry whenever appropriate.

All standard link relations use *hyphen-case*, e.g. [`version-history`](https://tools.ietf.org/html/rfc5829#page-4), which would yield:

```json
{
  "version-history": "http://localhost/revisions"
}
```

meshcalero:
Fixes #264

### DIFF
--- a/hyper-media/Hypermedia.md
+++ b/hyper-media/Hypermedia.md
@@ -51,7 +51,7 @@ HttpLink:
 The name of an attribute holding such a `HttpLink` object specifies the relation between the object that 
 contains the link and the linked resource. Implementations should use names from the 
 [IANA Link Relation Registry](http://www.iana.org/assignments/link-relations/link-relations.xhtml) 
-whenever appropriate.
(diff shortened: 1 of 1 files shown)


whiskeysierra:
:+1:

tkrop:
👍

PR #271, opened 2017-07-18
Clarify JSON payload requirement

### LINKED ISSUE #262: Section "13. Data Formats" contains contradictory requirements
If this document strives to conform RFC 2119 as stated in "1. Introduction" section then these two requirements in "13. Data Formats" are mutually exclusive
* Must: Use JSON as the Body Payload
* May: Use other Media Types than JSON

I suggest merging it into single section "Should: Use JSON as the Body Payload" which would list considerations that migh allow to chose alternative formats.

Also "May: Use other Media Types than JSON" says "But only do this if you can not transfer the information in JSON." - which also can be improved. E.g. I can imagine that _any_ binary data can be Base64 encoded and put into JSON string.

PetrGlad:
Hopefully resolves #262

I included a note about integration with external systems. Maybe that can be removed as interaction with such systems seems is out of scope of these guidelines.

Also I see not guidelines whether  RFC 2119 keywords should be upper-cased (normally they are, but I do not see this convention in in other documents here).

I am not sure whether this warrants an entry in changelog.

### DIFF
--- a/data-formats/DataFormats.md
+++ b/data-formats/DataFormats.md
@@ -1,16 +1,21 @@
 # Data Formats
 
-## {{ book.must }} Use JSON as the Body Payload
+## {{ book.should }} Use JSON as the Body Payload
 
-JSON-encode the body payload. The JSON payload must follow [RFC-7159](https://tools.ietf.org/html/rfc7159) by having
+Use JSON-encoded body payload for transferring structured data. 
+The JSON payload must follow [RFC-7159](https://tools.ietf.org/html/rfc7159) by having
 (if possible) a serialized object as the top-level structure, since it would allow for future extension.
 This also applies for collection resources where one naturally would assume an array. See the
 [pagination](../pagination/Pagination.md#could-use-pagination-links-where-applicable) section for an example.
 
-## {{ book.may }} Use other Media Types than JSON
-
-If for given use case JSON does not make sense, for instance when providing attachments in form of PDFs, you should
-use another, more sufficient media type. But only do this if you can not transfer the information in JSON.
+Other media types may be used in following cases:
+* Transferring binary data or data whose structure is not relevant. This is the case if payload structure 
+is not interpreted and consumed by clients as is. Examples:
+  * Downloading images (JPG, PNG, GIF)
+  * Exporting data in external document formats (PDF, DOC, XML)
+* Returning structured data in other media formats if this is required by external system that cannot 
+be made JSON-compatible. In this case data should also be available JSON-encoded if a client requests
+it with content negotiation.
 
 ## {{ book.must }} Use Standard Date and Time Formats
 
--- a/data-formats/DataFormats.md
+++ b/data-formats/DataFormats.md
@@ -1,6 +1,6 @@
 # Data Formats
 
-## {{ book.should }} Use JSON as the Body Payload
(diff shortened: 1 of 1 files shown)


data-formats/DataFormats.md
```diff
@@ -76,4 +81,4 @@ Previously, this guideline allowed the use of custom media types like `applicati
 This usage is not recommended anymore and should be avoided, except where it is necessary for cases
 of [media type versioning](../compatibility/Compatibility.md#must-use-media-type-versioning). Instead, the standard media type name `application/json` (or [`application/problem+json` for HTTP error details](http://zalando.github.io/restful-api-guidelines/common-data-objects/CommonDataObjects.html#must-use-problem-json)) should be used for JSON-formatted data.
 
-Custom media types with subtypes beginning with `x` bring no advantage compared to the standard media type for JSON, and make automated processing more difficult. They are also [discouraged by RFC 6838](https://tools.ietf.org/html/rfc6838#section-3.4).
\ No newline at end of file
+Custom media types with subtypes beginning with `x` bring no advantage compared to the standard media type for JSON, and make automated processing more difficult. They are also [discouraged by RFC 6838](https://tools.ietf.org/html/rfc6838#section-3.4).
```
PetrGlad: Adds EOL.

data-formats/DataFormats.md
```diff
@@ -1,16 +1,21 @@
 # Data Formats
 
-## {{ book.must }} Use JSON as the Body Payload
+## {{ book.should }} Use JSON as the Body Payload
```
duergner: I would still vote to have it as a **MUST** rule but highlight the possible exceptions as we want to enforce JSON unless there is a strong reason not to use JSON.

It might make sense to differentiate already here between structured data and binary data as it should be obvious that JSON is neither required nor desired when it comes to transmission of e.g. image data.

Providing data in e.g. *XML*, *DOC* or *PDF* format is, if implemented correctly based on content negotiation, simply a convenience implementation for the clients (simply put). But as already mentioned by your change that we urge to have a JSON version there as well.

tkrop: Agreed, if we use

```
## {{ book.should }} Use JSON as the Body Payload for Structured Data
```

we could use a second heading

```
## {{ book.may }} Use other Media Types than JSON for Binary Data or Alternative  Content Representations
```

to highlight the second rule sufficiently in the in table of contents.

tkrop:
@PetrGlad thanks for your contribution. It looks very good, can you integrate the comments on @duergner review?

data-formats/DataFormats.md
PetrGlad: OK

PetrGlad: @duergner Presence MUST in that section was the reason of this change since definition of MUST does not allow any exceptions which makes specification logically inconsistent. According to RFC 2119 recommendation to use something unless you have reason not to is SHOULD. Alternatively one could not claim that the documentation follows RFC 2119 and then MUST may have other meanings.
 
I hope suggestion @tkrop would be a good resolution for that.

PetrGlad:
See latest changes.

ePaul:
:+1:

tkrop:
👍

duergner:
👍

tkrop:
@NextVisitor please merge if you approve 😺

Issue #275, opened 2017-07-25
API Discovery section should be clarified


PetrGlad:
Since api-discovery projects were moved from github.com to our internal GHE
the link to https://github.com/zalando-incubator/api-discovery is now broken.
A side effect of this is that recommendation to use API Discovery endpoint `https://example.com/.well-known/schema-discovery` is now incomplete and missing any hints on required format and rules for `schema-discovery` resource.

I could not find recommendations for schema discovery in [OpenAPI spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md). Discovery does seem to be out of it's scope. Also RFC [Defining Well-Known Uniform Resource Identifiers (URIs)](https://tools.ietf.org/html/rfc5785) requires that all names below "/.well-known/" must be registered.
But [IANA registry](https://www.iana.org/assignments/well-known-uris/well-known-uris.xml) of such names does not list anything related to OpenAPI or Swagger.

The resolutions could be (not necessarily mutually exclusive) 
* Copy-paste the format details from internal project team-architecture/api-infrastructure (but still it should note that it is not a standard one).
* If this format is not an open specification but is used only at Zalando then we should remove this section or limit it to most generic recommendations.

maxim-tschumak:
Thanks for your input @PetrGlad!

We are going to roll out a new API discovery process soon. The current solution relies on the `.well-known` schema discovery endpoint. That's why we don't want to touch this guideline so far. After the change in the discovery workflow, we will adjust the respected guideline.

I'm going to adjust the link to the `api-infrastructure` repository since it's broken.

PetrGlad:
Also https://stackoverflow.com/questions/41660658/openapi-or-swagger-json-auto-discovery

PetrGlad:
@maxim-tschumak It is a bad idea to refer internal project in open document. Therefore I did  not list that as possible resolution.

A good solution would be to register schema discovery name in IANA since all other users of OpenAPI spec would benefit from it.

maxim-tschumak:
You are right @PetrGlad - I'm removing the internal link from the guideline and rephrasing the description.

I like your idea of registering the schema discovery name in IANA. Our future discovery solution will not rely on an additional endpoint though.

PetrGlad:
A less intrusive recommendation could be to put declaration into `/openapi.json`

dehora:
>Since api-discovery projects were moved from github.com to our internal GHE
[...]

PR #277, opened 2017-07-25
Removing api discovery link (#275)

maxim-tschumak:
Related to #275.

### DIFF
--- a/api-operation/ApiOperation.md
+++ b/api-operation/ApiOperation.md
@@ -22,7 +22,7 @@ An internal API is only used within the application and only by the owning team,
 for operational or implementation internal purposes.
 
 Background: In our dynamic and complex service infrastructure, it is important to provide API client developers a central place with online access to the OpenAPI reference definitions of all running applications.
-As a part of Zalando infrastructire, API Discovery solution is used as an API definition crawler.
+As a part of the Zalando infrastructure, the `.well-known/schema-discovery` endpoint is used by the API Discovery to detect all API definitions.
 It checks all running applications via the endpoint above and stores the discovered API definitions.
 API Discovery itself provides a RESTful API as well as an API Viewer (Swagger-UI) for central access to all discovered API definitions.
 
--- a/api-operation/ApiOperation.md
+++ b/api-operation/ApiOperation.md
@@ -5,10 +5,17 @@
 All service applications must support access to the OpenAPI Reference Definitions of their external APIs — it
 is optional for internal APIs — via the following two API endpoints:  
 
-* endpoint(s) for GET access on its OpenAPI definition(s), for instance
-  `https://example.com/swagger.json` or `https://example.com/swagger.yaml`.
-* API Discovery endpoint `https://example.com/.well-known/schema-discovery` that delivers
-  the OpenAPI definition endpoint(s) above (see the link below for a description of its format).
+* endpoint(s) for GET access on its OpenAPI definition(s), for instance `https://example.com/swagger.json` or `https://example.com/swagger.yaml`.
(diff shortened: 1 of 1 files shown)


api-operation/ApiOperation.md
```diff
@@ -21,12 +21,10 @@ another application owned by a different team or even an external business partn
...
 for operational or implementation internal purposes.
 
-Background: In our dynamic and complex service infrastructure, it is important to provide API client
-developers a central place with online access to the OpenAPI reference definitions of all running applications.
-[API Discovery](https://github.com/zalando-incubator/api-discovery) is an API definition
-crawler of the Zalando platform infrastructure; it checks all running applications via the endpoint above and
-stores the discovered API definitions. API Discovery itself provides a RESTful API as well as an
-API Viewer (Swagger-UI) for central access to all discovered API definitions.
+Background: In our dynamic and complex service infrastructure, it is important to provide API client developers a central place with online access to the OpenAPI reference definitions of all running applications.
+As a part of Zalando infrastructire, API Discovery solution is used as an API definition crawler.
```
tkrop:
```
As a part of the Zalando infrastructure, the `.well-known/schema-discovery` endpoint is used by the API Discovery to detect all API definitions.
```

maxim-tschumak: done

dehora:
-1. This doesn't fix the problem on what to put at the well-known link.

duergner:
Would it make sense to simply add the expected structure of that `schema-discovery` document directly to the guidelines?

tkrop:
@dehora Oh, I see. You are right. Here is the current information http://stups.readthedocs.io/en/latest/components/twintip.html, but what I do not know: will it be deleted in short too?

dehora:
>  but what I do not know: will it be deleted in short too?

I don't know, but I can tell you elements of Stups are being moved away from. Unless Zalando's not running twintip internally it seems to make sense to inline the JSON structure or link to that page directly so people know what to do.

dehora:
@duergner that option was mentioned here: https://github.com/zalando/restful-api-guidelines/issues/275#issuecomment-317723639. I don't really mind how it gets fixed.

duergner:
If we can inline I would suggest to inline. Similar to what we did with JSON Fields.

dehora:
> If we can inline I would suggest to inline.

I'm fine with this: @maxim-tschumak @tkrop,  thoughts?

maxim-tschumak:
I agree with @duergner and @dehora

api-operation/ApiOperation.md
```diff
@@ -5,10 +5,17 @@
 All service applications must support access to the OpenAPI Reference Definitions of their external APIs — it
 is optional for internal APIs — via the following two API endpoints:  
 
-* endpoint(s) for GET access on its OpenAPI definition(s), for instance
-  `https://example.com/swagger.json` or `https://example.com/swagger.yaml`.
-* API Discovery endpoint `https://example.com/.well-known/schema-discovery` that delivers
-  the OpenAPI definition endpoint(s) above (see the link below for a description of its format).
+* endpoint(s) for GET access on its OpenAPI definition(s), for instance `https://example.com/swagger.json` or `https://example.com/swagger.yaml`.
+* API Discovery endpoint `https://example.com/.well-known/schema-discovery` that delivers the following JSON discovery document:
```
ePaul: Maybe better "... that delivers a JSON discovery document of the following form (adapt the `schema_url` to point to the endpoint mentioned above):"

dehora:
:+1:

duergner:
👍

PR #279, opened 2017-08-02
No Caching Support by Default

### LINKED ISSUE #252: New Guideline: APIs should not support HTTP caching (because they are authenticated)
Related to / triggered by #228.

Background: We have several X-headers which potentially influence the results of API calls. A naive cache could get confused by this if this is not taken into account.

maxim-tschumak:
Closes #252.

### DIFF
--- a/performance/Performance.md
+++ b/performance/Performance.md
@@ -141,7 +141,7 @@ GET /order/123?embed=(items) HTTP/1.1
 
 While RESTful design suggests to support caching, our guidelines requires APIs to protect endpoints by SSL and [OAuth authorization](security/Security.html#must-secure-endpoints-with-oauth-2.0).
 As consequence, caching is non-trivial and has to take many aspects into account, e.g. general cacheability of response information, invalidation parameters, authentication, multiple instances of consumers.
-Thus, using transport layer caching is difficult, in best case inefficient, and in worst case impossible.
+Thus, using transparent HTTP caches is difficult, in best case inefficient, and in worst case impossible.
 
 As result API providers should always set the `Cache-Control: no-cache` header.
 
--- a/performance/Performance.md
+++ b/performance/Performance.md
@@ -145,6 +145,6 @@ Thus, using transport layer caching is difficult, in best case inefficient, and 
 
 As result API providers should always set the `Cache-Control: no-cache` header.
 
-**Note:** as this is a technical information that is attached to each response by service frameworks, so there is no need to document this header in the API specification.
+**Note:** as this is a technical information that is attached to each response by service frameworks, there is no need to document this header in the API specification.
 
 **Warning:** if an API is intended to support caching, it should take care to explicitly specify this ability by defining the caching boundaries, e.g. by giving dynamically hints using the `Vary` and `Cache-Control:` headers ([RFC-7234](https://tools.ietf.org/html/rfc7234#section-4.1)).
--- a/performance/Performance.md
+++ b/performance/Performance.md
@@ -137,14 +137,14 @@ GET /order/123?embed=(items) HTTP/1.1
 }
 ```
 
-## {{ book.must }} Not Implement Client-Side Caching If Not Supported by the API Provider
+## {{ book.should }} Not Implement Caching
 
-API consumers must not implement client-side caching.
-By default, the API producers don't support this.
-Client-side caching is only allowed if the support is explicitly specified by the API producer.
+While RESTful design suggests to support caching, our guidelines requires APIs to protect endpoints by SSL and [OAuth authorization](security/Security.html#must-secure-endpoints-with-oauth-2.0).
+As consequence, caching is non-trivial and has to take many aspects into account, e.g. general cacheability of response information, invalidation parameters, authentication, multiple instances of consumers.
+Thus, using transport layer caching is difficult, in best case inefficient, and in worst case impossible.
 
-If an API producer wants to allow client-side caching, he must specify the support in the API definition explicitly.
-In this case, the producer must also set the `Vary` headers ([RFC-7234](https://tools.ietf.org/html/rfc7234#section-4.1)).
+As result API providers should always set the `Cache-Control: no-cache` header.
 
-Zalando's [proprietary headers](http://zalando.github.io/restful-api-guidelines/headers/ProprietaryHeaders.html) can affect the query results.
-Existing caching solutions and libraries are not aware of the semantics and expected implications to the caching.
+**Note:** as this is a technical information that is attached to each response by service frameworks, so there is no need to document this header in the API specification.
+
+**Warning:** if an API is intended to support caching, it should take care to explicitly specify this ability by defining the caching boundaries, e.g. by giving dynamically hints using the `Vary` and `Cache-Control:` headers ([RFC-7234](https://tools.ietf.org/html/rfc7234#section-4.1)).
--- a/performance/Performance.md
+++ b/performance/Performance.md
@@ -136,3 +136,15 @@ GET /order/123?embed=(items) HTTP/1.1
   }
 }
 ```
+
+## {{ book.must }} Not Implement Client-Side Caching If Not Supported by the API Provider
+
+API consumers must not implement client-side caching.
+By default, the API producers don't support this.
+Client-side caching is only allowed if the support is explicitly specified by the API producer.
+
+If an API producer wants to allow client-side caching, he must specify the support in the API definition explicitly.
+In this case, the producer must also set the `Vary` headers ([RFC-7234](https://tools.ietf.org/html/rfc7234#section-4.1)).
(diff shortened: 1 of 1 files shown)


netme:
👍

performance/Performance.md
```diff
@@ -136,3 +136,15 @@ GET /order/123?embed=(items) HTTP/1.1
...
 ```
+
+## {{ book.must }} Not Implement Client-Side Caching If Not Supported by the API Provider
+
+API consumers must not implement client-side caching.
+By default, the API producers don't support this.
+Client-side caching is only allowed if the support is explicitly specified by the API producer.
+
+If an API producer wants to allow client-side caching, he must specify the support in the API definition explicitly.
+In this case, the producer must also set the `Vary` headers ([RFC-7234](https://tools.ietf.org/html/rfc7234#section-4.1)).
```
duergner: Do we really want to enforce this? `Vary` headers are especially useful if we want to allow (semi-)automatic client side caching configuration. But as all our APIs are authenticated and should therefore have *no-cache* headers I do not really see benefits of this header.

I see client side caching only happening on the client side business logic and being written by the developer of the client. I'd therefore focus more on a clear specification / explanation on the description of such cacheable calls rather than forcing to set a HTTP header

whiskeysierra:
Caching is one of six major constraints that define REST. Shouldn't we try to find a way to support it rather than just discouraging it by default?

We have JWT tokens now, that means, since they are self-contained, they can be validated locally without remote calls. This sounds promising to me: https://blog.feryn.eu/validating-json-web-tokens-in-varnish/

duergner:
Well if we want to have a generic definition of how caching can be done this would involve defining on a spec level what a cache needs to to when checking the authentication.

This would mean not just validating the token, but also checking whether the token has the required claims / scopes that are necessary for accessing the resource.

It get's even more complicated if the representation of the object returned varies on the privileges assigned to the callee token. You could e.g. have one endpoint which you can call with employee tokens and service tokens (for our IAM setup) where the service tokens can read *public* information whereas the employee token might be eligible to get more details in certain cases.

whiskeysierra:
> Well if we want to have a generic definition of how caching can be done this would involve defining on a spec level what a cache needs to to when checking the authentication.

Agreed.

> This would mean not just validating the token, but also checking whether the token has the required claims / scopes that are necessary for accessing the resource.

Also true. That has the negative effect that service providers need to touch two places everytime they want to enforce a specific scope for some endpoints. Especially in varnish this may not be as handy as e.g. in Spring.

> It get's even more complicated if the representation of the object returned varies on the privileges assigned to the callee token. You could e.g. have one endpoint which you can call with employee tokens and service tokens (for our IAM setup) where the service tokens can read public information whereas the employee token might be eligible to get more details in certain cases.

I'd assume that those cases are rare (even non-existent right now?), so if we can find something that works for 95% of our current use cases we're already in a better place. For anything more complicated disallowing caching is always an option.

meshcalero:
> Well if we want to have a generic definition of how caching can be done this would involve defining on a spec level what a cache needs to to when checking the authentication.

Maybe I miss the complexity here, but wouldn't the simplest solution just be to treat the Authorization header on a client side cache like any other header. As JWT token are self-contained, adding the Authorization header to the vary list would be sufficient for a service to indicate that the response depends on the Authorization header and for a client side cache to add the Authorization header value to the cache key. Of course this would bind the TTL of the cache entry to the TTL of the Auth token, but this in combination with ETAG should provide a more than good-enough caching solution for most cases.

duergner:
That should be sufficient yes. But will require non trivial computation on the server side as cache TTL need to be calculated depending on the TTL of the JWT token instead of being a fixed and pre-defind value.

I'm still not convinced whether this additional complexity is really worth the effort required. Especially compared to handling caching on business logic level on the client side based on informal specification.

performance/Performance.md
```diff
@@ -136,3 +136,15 @@ GET /order/123?embed=(items) HTTP/1.1
...
 }
 ```
+
+## {{ book.must }} Not Implement Client-Side Caching If Not Supported by the API Provider
+
+API consumers must not implement client-side caching.
+By default, the API producers don't support this.
+Client-side caching is only allowed if the support is explicitly specified by the API producer.
+
+If an API producer wants to allow client-side caching, he must specify the support in the API definition explicitly.
```
tkrop: I think we must be more specific how the API provider has to specify or declare this.

performance/Performance.md
tkrop: I think @duergner is right. We should downgrade this to a hint, that the API provider should consider using the `Vary:`-Header if necessary. It is a hint that the API designer has to consider headers special for caching.

performance/Performance.md
```diff
@@ -136,3 +136,15 @@ GET /order/123?embed=(items) HTTP/1.1
   }
 }
 ```
+
+## {{ book.must }} Not Implement Client-Side Caching If Not Supported by the API Provider
+
+API consumers must not implement client-side caching.
+By default, the API producers don't support this.
```
tkrop: Why? We should give more a reason here (from @duergner):

> But as all our APIs are authenticated and should therefore have no-cache headers ...

tkrop:
@duergner @dehora @meshcalero I would even argue that a transparent Web cache is not efficiently applicable in our application environment. So the only viable options are a client side caching based on business logic or a pure server side caching, which may even be more resource efficient.

maxim-tschumak:
@tkrop and I have rewritten the guideline. Please have another look.

ePaul:
> Maybe I miss the complexity here, but wouldn't the simplest solution just be to treat the Authorization header on a client side cache like any other header. As JWT token are self-contained, adding the Authorization header to the vary list would be sufficient for a service to indicate that the response depends on the Authorization header and for a client side cache to add the Authorization header value to the cache key.

Do you really want the cache to store all auth tokens? We try to educate everyone not to publish their tokens anywhere.

performance/Performance.md
```diff
@@ -136,3 +136,15 @@ GET /order/123?embed=(items) HTTP/1.1
   }
 }
 ```
+
+## {{ book.should }} Not Implement Caching
+
+While RESTful design suggests to support caching, our guidelines requires APIs to protect endpoints by SSL and [OAuth authorization](security/Security.html#must-secure-endpoints-with-oauth-2.0).
+As consequence, caching is non-trivial and has to take many aspects into account, e.g. general cacheability of response information, invalidation parameters, authentication, multiple instances of consumers.
+Thus, using transport layer caching is difficult, in best case inefficient, and in worst case impossible.
```
ePaul: I'm not sure that "transport layer" is the right term – that would be TCP (or maybe TCP+TLS) in this case, which is not what we want here. Maybe "using transparent HTTP caches" is better?

tkrop: Agreed :)

whiskeysierra:
> Do you really want the cache to store all auth tokens? We try to educate everyone not to publish their tokens anywhere.

They would certainly store tokens as part of the key (if necessary) but they won't be accessible by anyone, at least not by design. What's the harm here?

tkrop:
👍

tfrauenstein:
This guideline proposal wants to address potential security problems with generic HTTP caching?
This is a theoretical problem since AFAIK we currently do not use generic HTTP caches.
And is it really a good idea to require that all GET responses should be armed with headers for "no cache" or even "TTL" dynamically computed from request token TTL?

So my proposal: let's drop this rule candidate. In the security section we just give the general hint that we discourage usage of generic HTTP caches in front of our REST API (GET endpoints) since they are secured by tokens and usually not supported in efficient and secure way by generic HTTP caches.

tkrop:
@tfrauenstein No, we do not want to address security problems with caches, we just want to state that caches are not of much help with respect to our other requirements.

> And is it really a good idea to require that all GET responses should be armed with headers for "no cache" or even "TTL" dynamically computed from request token TTL?

@tfrauenstein Yes, this is exactly what this suggestion is about:

1. As consumers should not use caching unless he knows what he is doing, and
2. The service is explicitly allowing consumers to do so.

> So my proposal: let's drop this rule candidate. In the security section we just give the general hint that we discourage usage of generic HTTP caches in front of our REST API (GET endpoints) since they are secured by tokens and usually not supported in efficient and secure way by generic HTTP caches.

I would expect to find this topic in `performance`, as it is majorly performance related and only influenced by `security`. I suggest to leave it where it is now.

ePaul:
*From the meeting*:

* we want to turn the rule around to say "if you (in exceptional case) support caching, you should document it. In the other case, send "no-cache". → @tkrop tries to reword this.

whiskeysierra:
Why is this closed already?

tkrop:
@whiskeysierra Replaced by suggestion #315 implementing the last comments.

Issue #132, opened 2016-08-30
Recommend functional names


bocytko:
I will touch a very sensitive topic: service names. IMHO we should advise naming services that implement APIs to be with their name as close as possible to the API name itself.

I'm observing a high number of services with names that are more or less random (hero names, chemical elements, ...). Such names make it very difficult to understand the architecture of a system or even parts of it. It's also impossible to add such names to any kind of documentation or diagram.

A rule of thumb is to name the service with the root aggregate, entity or bounded context that's being managed by it. Examples:
- `Stock Location` for managing _Stock Locations_,
- `Sales Order` for managing _Sales Orders_,
- `Retail Catalog` manages _Retail Products_ (aggregate of product, price and stock data).
- `Fulfillment Order` for managing _Fulfillment Orders_, _Shipment Orders_, _Shipments_ (bounded context).

One can of course argue, that in different domains, an order means sth. different. However, prefixing the entities with the domain (bounded context) name is widely used: _Purchase Order_, _Sales Order_, _Fulfillment Order_, _Shipment Order_.
### Technical services & abbreviations

Code names are widely used for technical services, appliances and libraries (_Nakadi_, _buku_, _bubuku_, _spilo_), which is perfectly fine in my point of view. Other examples exist though (_Exception monitor_, _Deploy Control_).

Smart abbreviations like _Cuxy_ for _CUstomer proXY_ should be allowed for technical components like caches, tools.
### Suggestion for the guidelines

**Service names (and thus their public URLs)**
- Must: Service name is closely related to the entities, aggregate or bounded context that they manage (i.e. _Fulfillment Order_, _Sales Order_, _Stock Location_).
- Must: Use code names internally, don't expose them to clients.
- Could: Use code names for technical infrastructure components (like _Nakadi_).

/cc: @codekultur @tfrauenstein

ePaul:
I guess this could go into the "Naming" section.

codekultur:
I think we should discuss this together with the naming of permissions-topic, as they might relate - both should be modelled around bounded context names IMHO.

LappleApple:
@ePaul @codekultur @bocytko Hey, is this question still on the table ("open") or did you resolve it?

codekultur:
[...]

PR #281, opened 2017-08-15
Changed guidelines to support functional naming schema

### LINKED ISSUE #129: Prefer functional names in oauth scopes
Not infrequently, oauth scopes are defined using a leading team name or a project name:
- `"myteam.content.read"`

We should prefer to have scopes organised around a bounded context eg "sales", or a domain function, eg "cart". A product/service name if it exists, is preferable to a project name (eg "grpc" the product name is better than "stubby" the internal project name). This doesn't affect auth mechanics  and the documentation doesn't encourage using team names, but it's quite common. 

There are at least two reasons to do this:
- Avoid complecting team names and project names with long lived API functions.  In Zalando at least, teams are to an extent encouraged to have function free names. Teams and projects come and go, APIs are forever. On the other hand a scope called "mighty-ducks.data.read" doesn't signal intent well. 
- For public facing systems we want a well organised set of bounded contexts and product domains that make sense to end users and not just Zalandos.

tkrop:
Fixes #129, #132


ePaul:
This needs to be redone in AsciiDoc format.

ePaul:
Not for now, postponed.

PR #282, opened 2017-08-22
Migrating to Asciidoc

maxim-tschumak:
This PR implements the migration from Markdown/GitBook to Asciidoc(tor) and includes:

* Translation from Markdown to Asciidoc markup
* New build and deployment scripts
* Introduces RuleId concept and concrete rule ids (and tooling around them)
* Support for the existing links (redirects to the current rule locations)
* Adjusted documentation
* Updated Maintainers list

Please also have a look at the [generated HTML](https://maxim-tschumak.github.io/restful-api-guidelines/).

@tfrauenstein Feel free to award the rule ids [1,99] to the most prominent rules.

Related to #231, #69, #10, #238.
Closes #260.

PR #284, opened 2017-08-23
Fix typo in Compatibility.md

jonicious:
"JSON objects" -> "JSON object" without the "s" at the end.


tkrop:
Thanks for the hint. Will be fixed with #282.

Issue #278, opened 2017-07-28
Are additionalProperties subject to property naming rules?


whiskeysierra:
The [compatibility guidelines](http://zalando.github.io/restful-api-guidelines/compatibility/Compatibility.html) contain a small reference to `additionalProperties` with a schema as a value:

> Note that this guideline concentrates on default extensibility and does not exclude the use of `additionalProperties` with a schema as a value, which might be appropriate in some circumstances.

Those are used to represent dynamic key-based data structures, usually referred to as maps, objects or hashes. Do the [property naming rules](http://zalando.github.io/restful-api-guidelines/json-guidelines/JsonGuidelines.html) apply to the keys in those data structures as well?

whiskeysierra:
Our JSON guidelines carefully (?) use the word `property` instead of key. I'd therefore treat an object that was specified using `additionalProperties` as a complex value, but a value nonetheless. In other words clients should not assume the same rules for keys in complex values like that.

What are other opinions on this?

ePaul:
I would say that for maps (i.e. an object with just `additionalProperties` and not directly defined properties), the naming rules don't apply – the keys are not defined in the API definition anyways.

We might need to clarify this in the guidelines.

ePaul:
**From the meeting**:

The JSON guideline about snake_case could get an additional paragraph "this does not apply to maps".

wssbck:
This seems like a very easy way to circumvent any data quality requirements when it comes to JSON data. From the perspective of an API consumer what matters is the data it receives, not whether it was covered with additionalProperties in the schema definition.

tkrop:
Unfortunately, the issue up to now does not reflect the discussion behind this issue, which in my opinion is: **what are the rules for defining property keys/names?**

The reason behind this question is: (1) can generic processing tools can rely on naming conventions of properties, and (2) are special characters are excluded from property names, i.e. is the character restricted to [a-zA-Z0-9_]?

If there is such a rule or understanding implied by the term *snake case*, it is currently very hidden. We should either uncover it, or make clear, that it was never intended to exist.

ePaul:
@wssbck could you elaborate on what kind of "data quality" you are thinking about?

If we have a map with country-codes as keys, a consumer is not served by having them lower-case instead of the natural upper-case.

dehora:
> (1) can generic processing tools can rely on naming conventions of properties,

No, imo. It's too fragile for tools to rely on conventions. Conventions can change.

[...]

PR #288, opened 2017-08-30
#278: add regular expression for property names (rule 119)

ePaul:
This makes it clearer already in the table of contents what is actually wanted here, by condensing the ASCII subset explained in the rule into a regular expression.

This fixes an issue which came up in the discussion for #278.

Do we actually want the `$` in property names? Should we say something about when to use it, or what its special meaning is?

### DIFF
--- a/chapters/json-guidelines.adoc
+++ b/chapters/json-guidelines.adoc
@@ -22,7 +22,7 @@ essential to establish a consistent look and feel such that JSON looks
 as if it came from the same hand.
 
 [#119]
-== {MUST} Property names must match `^[a-zA-Z_$][a-zA-Z_$0-9]$`
+== {MUST} Property names must match `^[a-zA-Z_$][a-zA-Z_$0-9]*$`
 
 Property names are restricted to ASCII encoded strings. The first
 character must be a letter, an underscore or a dollar sign, and
--- a/chapters/json-guidelines.adoc
+++ b/chapters/json-guidelines.adoc
@@ -22,12 +22,13 @@ essential to establish a consistent look and feel such that JSON looks
 as if it came from the same hand.
 
 [#119]
-== {MUST} Property names must match `^[a-zA-Z_$][a-zA-Z_$0-9]*$`
+== {MUST} Property names must match `^[a-z_][a-z_0-9]*$`
 
-Property names are restricted to ASCII encoded strings. The first
-character must be a letter, an underscore or a dollar sign, and
-subsequent characters can be a letter, an underscore, a dollar sign, or
-a number.
+Property names are restricted to ASCII strings. The first
+character must be a letter, or an underscore, and subsequent
+characters can be a letter, an underscore, or a number.
+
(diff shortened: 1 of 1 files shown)


ePaul:
Actually, the regular expression is a lot more permissive than our "camel case" requirement of the next section. Should it possibly be `^[a-z_][a-z_0-9]*` instead? Is `$` part of camel case?

chapters/json-guidelines.adoc
```diff
@@ -22,7 +22,7 @@ essential to establish a consistent look and feel such that JSON looks
 as if it came from the same hand.
 
 [#119]
-== {MUST} Property names must be an ASCII subset
+== {MUST} Property names must match `^[a-zA-Z_$][a-zA-Z_$0-9]$`
```
ePaul: There is a `*` missing here.

ePaul: Fixed.

tkrop:
👍

tkrop:
@ePaul With respect to your question, I think it is included because of some libraries that map "Java" classes. I would like to throw it out. In this case we could refer to the [:word:] (\w) character class in the title. Besides. do we need to restrict the first character to a letter?

Note: I would like to use text in the header instead of a regular expression, but with the current set I do not have an idea how to name it.

ePaul:
I guess having the first character a digit creates problems when implementing in several languages.

I used the regexp in the title (and let the full explanation stay in the body) because it is shorter.
I'm also in favor of removing `$`. I'll change it to this if we get positive feedback from someone else.

tkrop:
I have checked all public available APIs. `$` is only used by a single team to mark meta data in resources. They have used this as an alternative concept instead to the recommended HAL. They claim that their APIs are *very old* and has actually moved over to a different team.

I will check with the other team, but I think we can safely remove this character from our list.

tkrop:
@ePaul I can check for first characters in property names, but it will take a few hours, cause I have other tasks to do first.

tkrop:
I did some more investigation on property names in our APIs:

Beside names starting `$`, we have a limited number of properties starting with

- `[0-9]`: 2 APIs using numbers as property names for representing kinds of histograms,
- `_links`: 18 APIs using our HAL adaptation,
- `_id`: 1 API using this strange duplication method to transport the identifier numeric, and
- `_tags`: 1 API using the underscore without reason (maybe because `tags` keyword in OpenAPI, but has no impact on parameter or property names).

duergner:
I think there is no library specific reason to allow for `$`. It might be the default configuration (most probably done in order to avoid clashing with any **real** property names) but that should be configurable for mapping libraries.

I'd prefer if we could come up with something that would not allow `_` as a first character per se but rather only allow specific keywords like `_links` to be exempt from the Regex.

ePaul:
@tkrop @duergner thanks for your comments. I reworded my proposal to use the simpler Regexp, and also mention that `_` at the start of the name is not recommended.

tfrauenstein:
👍

tkrop:
👍

PR #291, opened 2017-08-30
fix deployment of non-master branches

### LINKED ISSUE #290: Travis CI is pushing the result of feature branch builds
The result of the feature branch builds in Travis CI seems to be pushed automatically to the GH-pages branch. We had this problem before (#65), and it seems to be back with the migration to ASCIIdoctor (#282).

For example, right now my change from #288 is already visible in the generated HTML/PDF, although it is not yet merged.

ePaul:
This replaces the `-eq` arithmetic comparison with a `=` string comparison.

This is supposed to fix #290.


duergner:
👍

ePaul:
:+1:

PR #307, opened 2017-10-26
Added meta information requirements

tkrop:
For API management we have additional meta information requirements, that are also in general helpful for API users. Therfore, I suggest to put them in the general OpenAPI specification section.


whiskeysierra:
:+1:

tfrauenstein:
👍

Issue #283, opened 2017-08-22
Usability Improvements


maxim-tschumak:
After migrating to Asciidoc (#260), we are able to improve the usability of the guidelines for users and maintainers. Here are some proposals:

- [ ] "disable/enable sidebar" button
- [ ] Place footnotes right after rule (instead of at the end of the whole document)
- [ ] Collapse chapters in the navigation sidebar
- [ ] Highlighting of the current position in the navigation sidebar
- [ ] Buttons (and/or keyboard) navigation between the chapters: next/prev, as in the old version of the guidelines
- [ ] Further typography improvements
  - [ ] Fonts
  - [ ] Font size in ToC (chapters should be stressed)

These feature requests are nice to have. Not all of them have to be implemented.
Feel free to add your ideas and suggestions!

ePaul:
The ToC could also show the rule IDs (nice to have, by no means necessary).

ePaul:
**From the meeting**:

Whoever has time, feel free to grab parts of this and fix them, opening a PR.

PR #308, opened 2017-10-26
Fix chapter numbering and improve table of contents.

tkrop:
May be because of the last asciidoctor update the behavior of chapter enumeration has changed - maybe it was always wrong.

Is also related to #283.


maxim-tschumak:
:+1:

tkrop:
👍

PR #306, opened 2017-10-26
Add rule for api identifiers.

tkrop:
For API management we want to introduce API identifiers, that allow to track histories of APIs independent of volatile API titles. For more information see the internal proposal of [API identifiers for API histories](https://docs.google.com/document/d/1lRXcTZbZMnFeTVCaazitSWxSdKXGWkOUn99Gr-huNXg/edit#).

PR #311, opened 2017-11-07
New guideline for Maps – #278

### LINKED ISSUE #278: Are additionalProperties subject to property naming rules?

ePaul:
This fixes #278, by adding a new guideline (216, because 215 is already used by #306), indicating how maps should be represented.

On the way, I also merged rules 118 and 119, and removed the pseudo-rules (which were just headings) 117 and 121.

For reviewing, it might be clearer looking at each commit individually.

### DIFF
--- a/chapters/json-guidelines.adoc
+++ b/chapters/json-guidelines.adoc
@@ -27,6 +27,57 @@ Others, like Google and Amazon, use both - but not only camelCase. It’s
 essential to establish a consistent look and feel such that JSON looks
 as if it came from the same hand.
 
+[#216]
+==  Define maps using `additionalProperties`, without restriction on key format.
+
+A "map" here is a mapping from string keys to some other type.
+In JSON this is represented as an object, the key-value pairs being represented
+by property names and property values.
+In OpenAPI schema (as well as in JSON schema) they should be represented using
+additionalProperties with a schema for the value type. Such an object should normally
+have no other defined properties.
+
+The map keys don't count as property names in the sense of #118[rule #118], and can
+follow whatever format is natural for their domain. Please document this in the description
+of the map object's schema.
+
+Here is an example for such a map definition (the `translations` property):
+
+```yaml
+definitions:
+  Message:
+    description:
+      A message together with translations in several languages.
+    type: object
+    properties:
+      message_key:
+        type: string
+        description: The message key.
+      translations:
+        description:
+          The translations of this message into several languages.
+          The keys are usual locale identifiers (ISO-639- language codes,
+          together with an optional ISO-3166 country-code, separated by `_`).
+        type: object
+        additionalProperties:
+          type: string
+          description:
+            the translation of this message into the language identified by the key.
+```
+
+An actual JSON object described by this might then look like this:
+```json
+{ "message_key": "hello_world",
+  "translations": {
+    "de": "Hallo Welt!",
+    "en": "Hello World!",
+    "eo": "Saluton mondo!",
+    "cz": "Ahoj Světe!",
+    "el": "Γεια σου κόσμε!"
+  }
+}
+```
+
 [#120]
 == {SHOULD} Array names should be pluralized
 
--- a/chapters/json-guidelines.adoc
+++ b/chapters/json-guidelines.adoc
@@ -28,7 +28,7 @@ essential to establish a consistent look and feel such that JSON looks
 as if it came from the same hand.
 
 [#216]
-==  Define maps using `additionalProperties`, without restriction on key format.
+== {SHOULD} Define maps using `additionalProperties`, without restriction on key format.
 
 A "map" here is a mapping from string keys to some other type.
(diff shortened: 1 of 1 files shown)


chapters/json-guidelines.adoc
```diff
@@ -9,36 +9,81 @@ media type and custom JSON media types defined for APIs. The guidelines
...
+          type: string
+          description:
+            the translation of this message into the language identified by the key.
+```
+
+An actual JSON object described by this might then look like this:
+```json
+{ "message_key": "hello_world",
+  "translations": {
+    "de": "Hallo Welt!",
```
jhorstmann: A better example would be one that actually violates the naming conventions mentioned above, with [IETF / BCP47 language tags](https://en.wikipedia.org/wiki/IETF_language_tag) as keys:

```json
{
    "en-US": "color",
    "en-GB": "colour"
}
```

whiskeysierra: As a side note, we even have a [rule](http://zalando.github.io/restful-api-guidelines/#128) that promots BCP47.

ePaul: Fixed.

ePaul:
:+1:

chapters/json-guidelines.adoc
```diff
@@ -9,36 +9,80 @@ media type and custom JSON media types defined for APIs. The guidelines
...
 (It is recommended to use `_` at the start of property names only for keywords like `_links`.)
 
+Rationale: No established industry standard exists, but many popular Internet
+companies prefer snake_case: e.g. GitHub, Stack Exchange, Twitter.
+Others, like Google and Amazon, use both - but not only camelCase. It’s
+essential to establish a consistent look and feel such that JSON looks
+as if it came from the same hand.
+
+[#216]
+== {SHOULD} Define maps using `additionalProperties`, without restriction on key format.
```
tkrop: how about ```== {SHOULD} represent dynamic maps by `additionalProperties` ```?
In my opinion `without restriction on key format.` is ambiguous an may lead to misconceptions and incorrect expectations. Therefore, I would like users to read the section instead of interpreting the title.

chapters/json-guidelines.adoc
```diff
@@ -9,36 +9,80 @@ media type and custom JSON media types defined for APIs. The guidelines
...
-
-No established industry standard exists, but many popular Internet
-companies prefer snake_case: e.g. GitHub, Stack Exchange, Twitter.
-Others, like Google and Amazon, use both - but not only camelCase. It’s
-essential to establish a consistent look and feel such that JSON looks
-as if it came from the same hand.
-
-[#119]
-== {MUST} Property names must match `^[a-z_][a-z_0-9]*$`
+== {MUST} Property names must be ASCII snake_case (and never camelCase): `^[a-z_][a-z_0-9]*$`
```
tkrop: may be better ```+== {MUST} Property names must be snake_case (and never camelCase) matching `^[a-z_][a-z_0-9]*$` ```

whiskeysierra:
:+1:

chapters/json-guidelines.adoc
```diff
@@ -9,36 +9,80 @@ media type and custom JSON media types defined for APIs. The guidelines
 clarifies some specific cases to allow Zalando JSON data to have an
 idiomatic form across teams and services.
 
-[#117]
-== {MUST} Use Consistent Property Names
+The first some of the following guidelines are about property names,
```
tkrop: What is the intention of this sentence? Please improve or delete.

May be you wanted to reformulate the consistency rule for properties and values here?

ePaul: @tkrop The "consistency rules" (117 and 121) were originally not rules, but just headings between the chapter and the rule level, grouping a bunch of rules.

As this heading level disappeared with the migration to AsciiDoc, they somehow got converted to rules with numbers (117 and 121) (but without content).

The idea of this sentence is to transport the content of those pseudo-rules in the introductory section, but I guess I could live without it. (Or if you have a better idea of how to word it better, I'm open for it.)

tkrop: Thanks for answering. I think the sentence does not work out well, grammar, etc. As it is already merged, how do we proceed with this issue?

ePaul: If you know what to do, just open a new PR (editorial). If not, open a new issue.

PR #313, opened 2017-11-08
fix: updating information about the tooling around the rule id (#312)

### LINKED ISSUE #312: Contributing Manual is not Consistent with the Current Set Up of Rule Ids
`CONTRIBUTING.adoc` states that there is a file in the repo in which the last used Rule Id is stored. This is not the case. We switched to another approach of proofing of uniqueness of the Rule Id:
1. There is a script which checks the uniqueness of all assigned Rule Ids (from the sources).
2. There is a script which generates a new unique Rule Id.

`CONTRIBUTING.adoc` has to be adjusted.

Thanks to @ePaul for pointing out!

maxim-tschumak:
Fixes #312.

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -13,14 +13,12 @@ concluded and at least 2 reviewers gave their approvals.
 When your pull request does major changes, please also add an entry to
 the changelog.
 
-== Rule ID
+== Rule Id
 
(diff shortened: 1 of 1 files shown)


tkrop:
👍

maxim-tschumak:
:+1:

PR #314, opened 2017-11-09
Change guideline for API versioning

### LINKED ISSUE #310: Guideline on API Version Number is Impractical for Reviews
The current guideline for API version numbers is impractical for the review process. It requires API writers to update the minor or major version after the review has been done, while changing the "review" number on each request (see [rule 116](https://zalando.github.io/restful-api-guidelines/#116)).

This results in the problem that a reviewer cannot review the version number as it will be only changed afterwards. This results in more errors of releasing API with incorrect versions. Today most reviews are using GH/GHE which provide a commit hash for distinguishing reviewed API versions. In addition, this rule is hardly unknown or in consequences is unexpected, so that it is mostly not applied correctly.

I suggest to change the rule to standard semantic versioning, which is the most expected and formally already used process for API versioning.

tkrop:
Fixes #310.

### DIFF
--- a/chapters/compatibility.adoc
+++ b/chapters/compatibility.adoc
@@ -291,7 +291,7 @@ negotiation (see above).
 [#116]
 == {MUST} Use Semantic Versioning
 
-OpenAPI allows to specify the an API specification version in
+OpenAPI allows to specify the API specification version in
 `#/info/version`. To share a common semantic of version information we
 expect API designers to comply to http://semver.org/spec/v2.0.0.html[
 Semantic Versioning 2.0] rules `1` to `8` and `11` restricted to the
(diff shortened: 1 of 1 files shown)


chapters/compatibility.adoc
```diff
@@ -289,9 +289,32 @@ versioning, and go instead with media type versioning and content
 negotiation (see above).
 
 [#116]
-== {SHOULD} Provide Version Information in OpenAPI Documentation
+== {MUST} Use Semantic Versioning
 
-Only the documentation, not the API itself, needs version information.
+OpenAPI allows to specify the an API specification version in
```
maxim-tschumak: the an API version ->
the ~an~ API version

maxim-tschumak:
:+1:

whiskeysierra:
:+1:

whiskeysierra:
@tkrop Can you resolve the conflict?

tkrop:
👍

whiskeysierra:
:+1:

PR #315, opened 2017-11-09
Add Rule for Caching Support

### LINKED ISSUE #252: New Guideline: APIs should not support HTTP caching (because they are authenticated)

tkrop:
Replaces #279. 
Fixes #252.

### DIFF
--- a/chapters/performance.adoc
+++ b/chapters/performance.adoc
@@ -182,7 +182,7 @@ and <<104,OAuth authorization>>, resource update and invalidation rules,
 existence of multiple consumer instances. As a consequence, caching is
 in best case complex, in worst case inefficient. Thus, client side and
 transparent HTTP caching should be avoided for RESTful APIs unless the
-API designer has proven to known better.
+API designer has proven to know better.
 
 As default, API providers should set the `Cache-Control: no-cache`
 header.
--- a/chapters/performance.adoc
+++ b/chapters/performance.adoc
@@ -167,3 +167,26 @@ GET /order/123?embed=(items) HTTP/1.1
   }
 }
 ----
+
+[217]
+== {MUST} Document Caching, if Supported
+
+If an API is intended to support caching, it must take care to specify
+this ability by defining the caching boundaries i.e. time-to-live and
+cache constraints, by providing the `Cache-Control` and `Vary` headers
+(Please read https://tools.ietf.org/html/rfc7234[RFC-7234] carefully).
+
+Caching has to take many aspects into account, e.g. general cacheability
+of response information, our guideline to protect endpoints using SSL
+and #104[OAuth authorization], resource update and invalidation rules,
+existence of multiple consumer instances. As a consequence, caching is
+in best case complex, in worst case inefficient. Thus, client side and
+transparent HTTP caching should be avoided for RESTful APIs unless
+proven to known better.
+
+As default, API providers should set the `Cache-Control: no-cache`
+header.
+
+**Note:** as this is a technical standard information, that is attached
+to each response by most service frameworks automatically, there is no
+need to document this header in the API specification.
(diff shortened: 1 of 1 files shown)


chapters/performance.adoc
```diff
@@ -167,3 +167,26 @@ GET /order/123?embed=(items) HTTP/1.1
...
+cache constraints, by providing the `Cache-Control` and `Vary` headers
+(Please read https://tools.ietf.org/html/rfc7234[RFC-7234] carefully).
+
+Caching has to take many aspects into account, e.g. general cacheability
+of response information, our guideline to protect endpoints using SSL
+and #104[OAuth authorization], resource update and invalidation rules,
+existence of multiple consumer instances. As a consequence, caching is
+in best case complex, in worst case inefficient. Thus, client side and
+transparent HTTP caching should be avoided for RESTful APIs unless
+proven to known better.
```
whiskeysierra: > have been proven to work better.

chapters/performance.adoc
```diff
@@ -167,3 +167,26 @@ GET /order/123?embed=(items) HTTP/1.1
...
+== {MUST} Document Caching, if Supported
+
+If an API is intended to support caching, it must take care to specify
+this ability by defining the caching boundaries i.e. time-to-live and
+cache constraints, by providing the `Cache-Control` and `Vary` headers
+(Please read https://tools.ietf.org/html/rfc7234[RFC-7234] carefully).
+
+Caching has to take many aspects into account, e.g. general cacheability
+of response information, our guideline to protect endpoints using SSL
+and #104[OAuth authorization], resource update and invalidation rules,
```
maxim-tschumak: should be <<104,OAuth Authorizatoin>>

chapters/performance.adoc
tkrop: Tried another version, please have a look.

chapters/performance.adoc
```diff
@@ -167,3 +167,26 @@ GET /order/123?embed=(items) HTTP/1.1
...
+cache constraints, by providing the `Cache-Control` and `Vary` headers
+(Please read https://tools.ietf.org/html/rfc7234[RFC-7234] carefully).
+
+Caching has to take many aspects into account, e.g. general cacheability
+of response information, our guideline to protect endpoints using SSL
+and <<104,OAuth authorization>>, resource update and invalidation rules,
+existence of multiple consumer instances. As a consequence, caching is
+in best case complex, in worst case inefficient. Thus, client side and
+transparent HTTP caching should be avoided for RESTful APIs unless the
+API designer has proven to known better.
```
whiskeysierra: > [..] has proven to know better.

chapters/performance.adoc
```diff
@@ -167,3 +167,26 @@ GET /order/123?embed=(items) HTTP/1.1
...
+in best case complex, in worst case inefficient. Thus, client side and
+transparent HTTP caching should be avoided for RESTful APIs unless the
+API designer has proven to know better.
+
+As default, API providers should set the `Cache-Control: no-cache`
+header.
+
+**Note:** as this is a technical standard information, that is attached
+to each response by most service frameworks automatically, there is no
+need to document this header in the API specification.
```
ePaul: Hmm, this could be read as "caching headers don't need to be documented", which seems contrary to the actual goal of this rule.

We should make it more clear that just the default case of "Cache-Control: no-cache" doesn't need to be documented, while the case of caching being supported still needs to be documented.

tkrop: You are absolutely right. I didn't considered this interpretation.

ePaul:
:+1:

maxim-tschumak:
:+1:

tkrop:
👍

PR #316, opened 2017-11-09
switching to docker-based api discovery process (#275)

### LINKED ISSUE #275: API Discovery section should be clarified

maxim-tschumak:
Closes #275.

### DIFF
--- a/chapters/api-operation.adoc
+++ b/chapters/api-operation.adoc
@@ -2,37 +2,15 @@
 = API Operation
 
 [#192]
-== {MUST} Publish OpenAPI Specification in Docker Artifacts
+== {MUST} Publish OpenAPI Specification
 
 All service applications must publish OpenAPI specifications of their
 external APIs. This is optional for internal APIs.
 
-Publishing of APIs is done via Docker images as follows:
-
-If the service provides one or more APIs::
-Copy all API specifications into the `/zalando-apis` directory of the
-Docker image of the service:
-
-[source]
-----
-COPY my-api-1.yaml ... my-api-n.yaml /zalando-apis/
-----
-
-If the service does not provide any APIs::
-Put an empty `no-apis` file into the `/zalando-apis` directory of the
-Docker image of the service:
-
-[source]
-----
-RUN mkdir /zalando-apis && touch /zalando-apis/no-apis
-----
-
-Please note that in the later case a shell has to be installed in the docker
-container. Alternatively, you can create empty `no-apis` file locally and make
-use of Docker `COPY` command in the Dockerfile.
-
-We prefer the Docker image publishing method over the past (now legacy)
-`.well-known/schema-discovery` service endpoint-based discovery process. However,
+OpenAPI specifications are published via deployment artifacts by the API owner,
+as described in the internal https://docs.google.com/a/zalando.de/document/d/1WBpRHp1DAtz7Pfjt0QwPuZNR1e70APRqUlCCby3JiOM/edit?usp=sharing[narrative].
+We prefer the deployment artifact-based method over the past (now legacy)
+`.well-known/schema-discovery` service endpoint-based publishing process. However,
 we currently support this way for backward compatibility reasons.
 
 We distinguish between internal and external APIs of an application which is
@@ -45,10 +23,8 @@ instance, for operational or implementation internal purposes.
 Background: In our dynamic and complex service infrastructure, it is important
 to provide API client developers a central place with online access to the API
 specifications of all running applications. As a part of the infrastructure,
-the API discovery process is used to detect API specifications. It scans all
-published Docker images and correlates API-related information with running
-applications on AWS and Kubernetes clusters. The findings are published in the
-API Portal - the universal hub for all Zalando APIs.
+the API publishing process is used to detect API specifications.
+The findings are published in the API Portal - the universal hub for all Zalando APIs.
 
 [#193]
 == {SHOULD} Monitor API Usage
--- a/chapters/api-operation.adoc
+++ b/chapters/api-operation.adoc
@@ -4,8 +4,9 @@
 [#192]
 == {MUST} Publish OpenAPI Specification
 
-All service applications must publish OpenAPI specifications of their
-external APIs. This is optional for internal APIs.
+All service applications must publish OpenAPI specifications of their external
+APIs. While this is optional for internal APIs, we still recommend to do so as
+well to profit from the API management infrastructure.
 
 OpenAPI specifications are published via deployment artifacts by the API owner,
 as described in the internal https://docs.google.com/a/zalando.de/document/d/1WBpRHp1DAtz7Pfjt0QwPuZNR1e70APRqUlCCby3JiOM/edit?usp=sharing[narrative].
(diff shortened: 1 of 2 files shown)


maxim-tschumak:
:-1:

ePaul:
Hmm, should we maybe leave both ways in the guidelines for some time?

maxim-tschumak:
@ePaul We (team api-management) decided to replace the rule in order to force the switch to the new process. (we still will support the old discovery method in the near future)

maxim-tschumak:
Please review the changes and let's make it merge-ready. We can merge this change as soon as the new process is announced.

whiskeysierra:
> We (team api-management) decided to replace the rule in order to force the switch to the new process.

Good luck with that. Since you're not in the position to force anyone, how are you planning to do this?

maxim-tschumak:
> Good luck with that. Since you're not in the position to force anyone, how are you planning to do this?

Ok, "to force" is maybe the wrong word. Let's name it "motivate"? As I said, we still support the old way.
This rule cannot be enforced anyway since we cannot check which docker images have to have a published API definition.

chapters/api-operation.adoc
```diff
@@ -2,36 +2,32 @@
...
-  "ui_url": "/ui/"
-}
+COPY my-api.yaml /zalando-apis/my-api.yaml
 ----
 
-* whether these endpoints have to be secured by OAuth depends on context
-of your API and product management
-* if you secure these endpoint, only use `uid` scope
+If docker image doesn't provide any APIs::
+Put empty `no-apis` file into the `/zalando-apis` directory:
```
hjacobs: This still feels weird to me: why do we leak "API information" to arbitrary applications / Docker images? Why would my background job running in Kubernetes need a `no-apis` file?

maxim-tschumak: Eventually, this should help to recognize three possibilities of API declaration by the docker image provider/owner:
1. Docker image provides API(s) (public API)
2. Docker image doesn't provide any API (e.g. background job)
3. Docker image owner forgot/hasn't declared any API

The last one would indicate a non-compliant behavior.

hjacobs: While I understand your intentions, I still think it's bad practice to leak the "API domain" into unrelated applications/images. Stupid analogy to explain what I mean: that's like putting a `no-web-app` file into every git repo which does not have a web application just to be able to detect web applications.

tkrop:
> Hmm, should we maybe leave both ways in the guidelines for some time?

@ePaul I do not see any value in explaining the old method when the new (preferred) method is doing the same.

chapters/api-operation.adoc
mfellner: @hjacobs So you are saying that there is a subset of all docker images (images for background-services etc.) that are naturally outside the "API domain" and it wouldn't make sense to "pollute"  them with unrelated domains.

If there is a way to identify this "domain" that a docker image naturally belongs to, we can exclude them from the compliance process (i.e. don't generate violations if `no-apis` does not exist).

whiskeysierra: > If there is a way to identify this "domain" that a docker image naturally belongs to, we can exclude them from the compliance process

If you would keep the current way of exposing API schemas using a well defined endpoint, then there would be no need for any of this, since background services, naturally wouldn't have an HTTP API.

mfellner: @whiskeysierra we'll keep the old way for backwards compatibility reasons. However, as we've explained in our [narrative](https://docs.google.com/document/d/1kmK_ieU1FqU6vQvgSonmBNu1dAPwD3TvJ39jMnk2hms) this way is not sufficient to discover potentially all APIs and has other drawbacks as well. We concluded that placing API definitions inside the docker image is the best compromise.

whiskeysierra: In your narrative, you state that

> many services/developers do not comply to the rules-of-play and forget to implement this endpoint

Since that behaviour won't change, I assume you're planning to act as a gatekeeper, preventing everyone from deploying if they're not complying? In contrast to the current approach that only allows to react, rather than prevent.

maxim-tschumak: >  In contrast to the current approach that only allows to react, rather than prevent.

Yes, we can only react to docker images which don't provide API-related information. By having this information in place, we can potentially prevent non-compliant deployments. (no concrete plans so far)

maxim-tschumak:
@ePaul @whiskeysierra I've added information about the legacy discovery process (and that we still support it)

tkrop:
Changed my mind. I think we should mention that the old /.well-known/schema-discovery is still supported but deprecated because of its inprecision to reflect the actually deployed API.

maxim-tschumak:
:+1:

ePaul:
The issue with the new way is that it only works for docker-based deployments, not for old datacenter deployctl deployments.

maxim-tschumak:
Good point @ePaul! We'll investigate how can we cope with data center deployments.

mfellner:
Although I support our solution of shipping API specifications at a certain location in Docker images I think the language of the proposed rule change is too specific.

Right now our (open source) guidelines don't make assumptions about application deployments (and deployment artifacts) but the new rule is written like Docker images are the obvious and only way. First of all I agree with @tkrop that we should still mention the old /.well-known way (as deprecated) but imho we must also think about generalizing our internal solution of specifications in Docker images for the updated rule.

Since this is a MUST rule,  we shouldn't introduce non-API related constraints like Docker images. Adopters of our guidelines may not be using Docker and even our own tooling around this process isn't even open source (yet).

tkrop:
We could talk in general about `deployment unit` instead of `Docker` image, but this would just make the guidance more obscure as long as we give a detailed example. So I see the following options:

1. We keep it as more or less specific as it is now (reintroducing the details about the `.well-known/schema-discovery` endpoint), or
2. We strip the rule to the general must rule about `publishing APIs` and refer to an external *Narrativ* explaining the details how API publishing and discovery is handled within Zalando.

I would prefer the second.

tkrop:
Please have a final look. I would like to have it merged on Tuesday.

whiskeysierra:
:+1:

mfellner:
👍

maxim-tschumak:
:+1:

PR #326, opened 2017-11-22
improve idempotent behavior description

ISO50:
Try to make clearer the point that idempotent behavior is not about the response of a http operation.


ePaul:
:+1:

whiskeysierra:
:+1:

PR #332, opened 2017-11-27
Fix #331: Correct a few typos

### LINKED ISSUE #331: Typos in the Guidelines
There are multiple typos in the guidelines, for example:

* `IDs must opaque strings` should be `IDs must be opaque strings`
* `different versionsof a resource` should be `different versions of a resource`
* `make sure that all constraints a clearly defined` should be `make sure that all constraints are clearly defined`

etc.

It would be nice to proof-read and correct these typos as they may sometimes distract a bit from the main points made in the document.

antivanov:
* Correct a few typos
 * Interchange the terms `page` and `offset` in *"137. Use Conventional Query Strings"*

### DIFF
--- a/chapters/naming.adoc
+++ b/chapters/naming.adoc
@@ -86,7 +86,7 @@ below. Hint: You can use size as an alternate query string.
 Hint: In combination with limit, you can use page as an alternative to
 offset.
 * `sort` — comma-separated list of fields to sort. To indicate sorting
-direction, fields my prefixed with + (ascending) or - (descending,
+direction, fields may be prefixed with + (ascending) or - (descending,
 default), e.g. /sales-orders?sort=+id
 * `fields` — to retrieve a subset of fields. See
 <<156,_Support Filtering of Resource Fields_>> below.
--- a/chapters/api-operation.adoc
+++ b/chapters/api-operation.adoc
@@ -34,7 +34,7 @@ be specified in the OpenAPI definition as they are generic and provide
 no API specific information.
 
 We distinguish between internal and external APIs of an application
-which is owned by a specific team and often implemented via small set of
+which is owned by a specific team and often implemented via a small set of
 services. An external API is used by clients outside the team - usually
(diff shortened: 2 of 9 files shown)


hjacobs:
Thank you! :smile:

antivanov:
Please, check that the small corrections make sense and let me know if PR needs to be updated. Would be nice to have a couple of reviewers looking over the corrections 😄

chapters/naming.adoc
```diff
@@ -83,10 +83,10 @@ should have an entity specific alias, like sku
 below. Hint: You can use size as an alternate query string.
 * `cursor` — key-based page start. See <<pagination>> section below.
 * `offset` — numeric offset page start. See <<pagination>> section below.
-Hint: In combination with limit, you can use page as an alternative to
-offset.
+Hint: In combination with limit, you can use offset as an alternative to
```
tkrop: I think, this change is not what was intended by the hint. Our standard for pagination is `offset`/`cursor` and `limit`. The hint was supposed to add `page` as an alternative to `offset`.

As far as I can see pagination is not suggesting this replacement, thus I would suggest to drop this hint here, too.

antivanov: Thanks, looks like I misunderstood how the pagination is defined. So `page` in the sentence 

>Hint: In combination with `limit`, you can use `page` as an alternative to
offset.

has the same meaning as `cursor` when slightly rephrased:

>Hint:  As pagination is defined by parameters `offset|cursor` and `limit`, in combination with `limit`, you can use `cursor` as an alternative to `offset`.

Will revert this change to use the original text.

antivanov: Reverted the change in the hint.

antivanov:
Thanks for reviewing, reverted the hint for *"Use Conventional Query Strings [137]"* back to the original.

tkrop:
👍

maxim-tschumak:
:+1:

tkrop:
👍

PR #333, opened 2017-11-28
Default to client credentials flow

bzums:
I did notice the note about the `flow` property: 

> Unfortunately the flow field is mandatory and cannot be omitted. API endpoints should always set flow: password and ignore this information.

However I would prefer not to encourage to use a "[discouraged](https://pages.github.bus.zalan.do/foundation/docs/oauth2-guide/oauth2-grants.html?highlight=flow#resource-owner-password-grant)" flow (even if not intended actually).

### DIFF
--- a/chapters/security.adoc
+++ b/chapters/security.adoc
@@ -15,14 +15,14 @@ specification or take a look at the following example.
 securityDefinitions:
   oauth2:
     type: oauth2
-    flow: password
+    flow: clientCredentials
     tokenUrl: https://token.services.auth.zalando.com/oauth2/access_token
     scopes:
       fulfillment-order-service.read: Access right needed to read from the fulfillment order service.
       fulfillment-order-service.write: Access right needed to write to the fulfillment order service.      
 ----
 
-The example defines OAuth2 with password flow as security standard used
+The example defines OAuth2 with client credentials flow as security standard used
 for authentication when accessing endpoints; additionally, there are two
 API access rights defined via the scopes section for later endpoint
 authorization usage - please see next section.
@@ -30,7 +30,7 @@ authorization usage - please see next section.
 It makes little sense specifying the flow to retrieve OAuth tokens in
 the `securityDefinitions` section, as API endpoints should not care, how
 OAuth tokens were created. Unfortunately the `flow` field is mandatory
-and cannot be ommited. API endpoints should always set `flow: password`
+and cannot be ommited. API endpoints should always set `flow: clientCredentials`
 and ignore this information.
 
 [#105]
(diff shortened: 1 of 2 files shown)


chapters/security.adoc
```diff
@@ -15,22 +15,22 @@ specification or take a look at the following example.
 securityDefinitions:
   oauth2:
     type: oauth2
-    flow: password
+    flow: clientCredentials
```
codekultur: according to https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#securityDefinitionsObject IMHO the flow name must be named 'application' for client credential flows.

codekultur:
Good spot. Thanks! :) (see my comment on the flow name for details)

chapters/security.adoc
```diff
@@ -15,22 +15,22 @@ specification or take a look at the following example.
...
     tokenUrl: https://token.services.auth.zalando.com/oauth2/access_token
```
codekultur: I'm not sure, but I think the tokenUrl needs to be changed as well then, as this IMHO is the URL of the adaptor used to translate the previously used password flow used for s2s communication to the now supported client credentials grant with Platform IAM. Please ask team foundation for details.

chapters/security.adoc
bzums: Ah ok, I used `clientCredentials` because of the spec from swagger: https://swagger.io/docs/specification/authentication/oauth2/

chapters/security.adoc
bzums: You are probably right, I will check that and update the PR

bzums:
@codekultur I updated the PR. I am still a bit confused about `application` flow from https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#securityDefinitionsObject (cause it is never named like that anywhere else, in the RFC for example or the swagger spec)

codekultur:
@bzums great! Thanks... I'm confused about the name  `application` as well, as the term is totally different from the OAuth spec, but IMHO https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md is the definitive source of truth when it comes to writing  OpenAPI 2.0 specs.

Are you really sure about the tokenUrl? For me this looks like the URL for the Platform IAM system in a sandbox (non productive) environment. Is this also the URL for the production IAM system?

bzums:
@codekultur you are absolutely right, I updated the url to the production endpoint

codekultur:
@bzums Perfect! Thanks for your contribution :)

codekultur:
👍

bzums:
👍

PR #334, opened 2017-11-28
clarify rule 196, fixes #302

### LINKED ISSUE #302: rule #196 is hard to understand
> Must: Ensure Event Type schemas conform to Open API’s Schema Object [196]

maybe an explanation of "Open API's Schema Object" (as opposed to Open API itself) would help.

also, i stumbled over "the list of non-available keywords relative to JSON-Schema". to a casual reader, this sounds like `oneOf` is forbidden, while i suppose it simply means that `oneOf` is a reserved keyword.

tlossen:
* make rationale for using object schema clearer
* clarify implications
* changed wording

### DIFF
--- a/chapters/events.adoc
+++ b/chapters/events.adoc
@@ -45,22 +45,25 @@ with the "API first" principle in mind as described in the <<introduction>>.
 == {MUST} Make Event schema available for review
 
 Services publishing event data for use by others must make the event
-schema available for review.
+schema as well as the event type definition available for review.
 
 [#196]
-== {MUST} Ensure Event schema conforms to Open API's Schema Object
+== {MUST} Ensure Event schema conforms to Open API Schema Object
 
-To align event definitions with API resource definitions, we use
-Schema Object from Open API to define events. This is particularly
-useful for events that represent data changes about resources.
+To align the event schema specifications to API specifications, we use
+the Schema Object as defined by the Open API Specifications to define event
+schemas. This is particularly useful for events that represent data changes
+about resources also used in other APIs.
 
-Schema Object is an "extended subset" of
-http://json-schema.org/[JSON Schema Draft 4]. For your convenience,
-we highlight some important differences below; please refer to the
-https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schemaObject[Schema Object specification] for details.
+The https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schemaObject[Open
+API Schema Object] is an **extended subset** of
+http://json-schema.org/[JSON Schema Draft 4]. For convenience,
+we highlight some important differences below. Please refer to the
+https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schemaObject[Open
+API Schema Object specification] for details.
 
-Schema Object _removes_ some JSON Schema keywords; the following properties
-*must not* be used in event schemas:
+As the Open API Schema Object specification _removes_ some JSON Schema
+keywords, the following properties *must not* be used in event schemas:
 
 * `additionalItems`
 * `contains`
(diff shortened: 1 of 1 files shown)


maxim-tschumak:
:+1:

chapters/events.adoc
```diff
@@ -50,27 +50,17 @@ schema available for review.
...
-https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schemaObject["Schema
-Object" section of that specification]).
-
-Open API _removes_ some JSON-Schema keywords. These must not be used in
-event type schemas. The list of Open API object keywords can be seen
-https://github.com/OAI/OpenAPI-Specification/blob/master/schemas/v2.0/schema.json#L935-L1063[here],
-but for convenience the list of non-available keywords relative to
-JSON-Schema are:
+To align event definitions with API resource definitions, we use
+Schema Object from Open API to define events. This is particularly
```
tkrop: May be better `Schema Object from Open API to define events.` -> `Schema Objects from the Open API specification to define events.`

While the title refers to *Schema Object* in the API specification, I would still use the plural form int the description here and in subsequent sentences for grammatical reason.

chapters/events.adoc
```diff
@@ -50,27 +50,17 @@ schema available for review.
...
+useful for events that represent data changes about resources.
+
+Schema Object is an "extended subset" of
+http://json-schema.org/[JSON Schema Draft 4]. For your convenience,
+we highlight some important differences below; please refer to the
+https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schemaObject[Schema Object specification] for details.
+
+Open API _removes_ some JSON Schema keywords; the following properties
```
tkrop: I dislike the aspect that these two sentences feel so unconnected - without explicit chain of reasoning. May be better `As Open API does not support all JSON Schema keywords, these must not be used in event schemas. Forbidden keywords are:`

chapters/events.adoc
tlossen: i prefer singular here, because i see `schema object` on the same level as `json schema`. but i replaced `open api` with `schema object` in more cases now, please have another look.

chapters/events.adoc
tlossen: that's what i was trying to express with `extended subset`.

chapters/events.adoc
tkrop: If you like, but the sentence really violates the language grammar in this form 😞.

chapters/events.adoc
tkrop: Sorry, I just made a suggestion to make the sentences easier to read and understand. Actually I would like to suggest replace the term `extended subset` by the above sentence and a second sentence, that explains the additional keywords allowed to improve reading even more.

chapters/events.adoc
```diff
@@ -42,35 +42,25 @@ just as they would an API. For example this means approaching events
...
-
-In the rest of this section we'll call out some of the more notable
-differences between Open API and JSON-Schema. Please note this is not a
-complete list - in general it's recommended to familiarise yourself with
-Open API's Schema Object (the details are available from the
-https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schemaObject["Schema
-Object" section of that specification]).
-
-Open API _removes_ some JSON-Schema keywords. These must not be used in
-event type schemas. The list of Open API object keywords can be seen
```
dehora:
```diff
-Open API _removes_ some JSON-Schema keywords. These must not be used in
-event type schemas. 
```

I'm not sure this should be removed as a editorial change.

ePaul: The sentence is still there (slightly reworded).

chapters/events.adoc
```diff
@@ -81,19 +74,18 @@ JSON-Schema are:
 * `not`
 * `oneOf`
 
-Open API _redefines_ some keywords:
+On the other side Schema Object _redefines_ some JSON Schema keywords:
 
 * `additionalProperties`: For event types that declare compatibility
 guarantees, there are recommended constraints around the use of this
 field. See the guideline "Avoid `additionalProperties` in event type
 definitions" for details.
```
ePaul: Can we get a link to  rule 210 here?

ePaul:
:+1:

maxim-tschumak:
:+1:

chapters/events.adoc
tkrop: I like this suggestion: see #372.

PR #342, opened 2017-12-20
Rule 150 allows PATCH to respond 204 No Content

### LINKED ISSUE #341: Use Specific HTTP Status Codes [150] Denies PATCH -> 204
Rule 150 to use specific HTTP status codes seems to suggest that the only successful PATCH responses are:
* 202 Accepted (implies asynchronous, change may or may not happen in the future)
* 200 OK (implies synchronous completion, suggests some content will follow)

But it that leaves a seemingly obvious cases unsupported: 
* 204 No Content (implies synchronous completion, but nothing further to say)

Any reason not to allow 204 responses to a PATCH?

roxspring:
Fixes #341

### DIFF
--- a/chapters/http.adoc
+++ b/chapters/http.adoc
@@ -270,7 +270,7 @@ with the Location header. (More details found in the
 |202 |Accepted - The request was successful and will be processed
 asynchronously. |POST, PUT, DELETE, PATCH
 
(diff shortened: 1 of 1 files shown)


tkrop:
👍

maxim-tschumak:
:+1:

PR #344, opened 2017-12-27
fix: document-internal references (#323)

### LINKED ISSUE #323: Rule #150 incorrectly references discussion of 207 for rate limiting
The [rule 150](http://zalando.github.io/restful-api-guidelines/#150), in particular, when discussing the `429` response code, contains the following:

> Too many requests - the client does not consider rate limiting and sent too many requests. See Must: [Use 207 for Batch or Bulk Requests](http://zalando.github.io/restful-api-guidelines/#152).

This incorrectly links to http://zalando.github.io/restful-api-guidelines/#152 where it should actually link to http://zalando.github.io/restful-api-guidelines/#153, which is "Must: Use 429 with Headers for Rate Limits [153]"

maxim-tschumak:
Fixes #323.


tkrop:
👍

maxim-tschumak:
:+1:

PR #351, opened 2018-01-05
Editorial: clarify difference between "incompatible" and "breaking" AP…

tfrauenstein:
…I changes.


chapters/compatibility.adoc
```diff
@@ -16,11 +16,20 @@ There are two techniques to change APIs without breaking them:
...
-consumers enable us (having Postel’s Law in mind) to make compatible
-changes without versioning.
-
-*Hint:* Please note that the compatibility guarantees are for the "on
+versioning (see rules <<113,113>>,<<114,114>> below). 
+The following guidelines for service providers (<<107,rule 107>>) and
+consumers (<<108,rule 108>>) enable us (having Postel’s Law in mind) 
+to make compatible changes without versioning.
+
+*Hint:* Note, there is a difference between incompatible and braking changes:
```
tkrop: May I suggest the following - in my humble opinion - easier to read and understand version:

`*Note:* There is a difference between incompatible and braking changes. Incompatible changes are changes that are not covered by the compatibility rules below. Breaking changes are incompatible changes deployed into operation, and thereby breaking running API consumers. However, in specific controlled situations it is possible to deploy incompatible changes in a non-breaking way, if no API consumer is using the affected API aspects (see also <<deprecation, Deprecation>> guidelines).`

chapters/compatibility.adoc
```diff
@@ -16,11 +16,20 @@ There are two techniques to change APIs without breaking them:
...
+Incompatible changes are changes that are not covered by the compatibility 
+rules below; they potentially break API consumers. 
+Braking changes are changes that really brake running API consumers when deployed into operation.
+Typically, incompatible changes are braking changes when deployed into (uncontrolled) operation environment. 
+However, in specific controlled situations with no API consumers using the affected API aspect, 
+incompatible changes are not braking changes (see also <<deprecation, Deprecation>> guidelines).
```
ePaul: I think you want to use "break/breaking", not "brake/braking" here. Several occasions (at least 5) in this paragraph.

chapters/compatibility.adoc
tfrauenstein: thx -- basically adopted in update.

tkrop:
👍

meshcalero:
+1

Issue #356, opened 2018-01-16
Should collections consume 2 resources in "Limit number of Resources [146]"?


roxspring:
Currently the Zally implementation literally counts the number of path entries in the swagger file and ensures that this number is below some configured paths_count_limit (currently 8). This has the effect that any collection of `things` instantly burns up two of the allowed paths `/things` and `/things/{id}` - is that the intention of this guideline?

In my team we were considering a variation where we just put a limit on the count of top-level resources i.e. `/things`, `/things/{id}`, `/things/{id}/related-thing/` all count as a single top-level resource, where `/extra/`, `/extra/stuff/` would count as a second top-level resource. It occurred to me that this might have been the original intention of the rule?

maxim-tschumak:
Good catch! The rule implementation is incorrect. It's not about the number of paths.

The current guideline rule is imprecise and doesn't tell us if we should check top-level resources only or all resources. (It's probably meant to be not too precise)

I believe it makes sense to count top-level resources only. Just in order to be not too strict.

maxim-tschumak:
@ePaul @tkrop @mfellner please have a look

ePaul:
For the reference, those numbers came in in issue #28 (PR #71) (previously it said 16, with a harder wording, which was already in there before we started version control on the guidelines). (@tfrauenstein and @whiskeysierra were involved back then, too.)

There was no mention of "only count top-level" in the discussion, and I would read it as counting all resources (including "sub-resources", i.e. resources whose names contain one or more slashes). However, I can understand that others read it differently.

However we decide, we should certainly clarify the wording in the rules here.

(Actually, I think the name "resource" here is wrong, as e.g. `/orders/{id}` is not one resource, but an unlimited family of resources, one for each value of the parameter: `/orders/1`, `/orders/2`, `/orders/3`, ... We don't want to count each of this.)

roxspring:
Just to make sure I understand your suggestion @tkrop, the following would all relate to a single type of resource:
```
GET /customers
POST /customers
GET /customers/{id}
POST /customers/{id}/some-customer-focused-action
```
Whereas the following would be a second resource:
```
GET /addresses
POST /addresses
GET /addresses/{id}
```
Sub-resources are a little unclear but you'd be happy with a Zally rule interpreting these as a third resource:
```
GET /customers/{id}/addresses
GET /customers/{id}/addresses/{addr}
```

[...]

PR #361, opened 2018-02-02
#356 Limit Resource Families

roxspring:
Suggested improvements for "SHOULD Limit number of Resources [146]" rule as discussed in #356.

Also some initial changes to update Zally accordingly:
https://github.com/zalando/zally/compare/master...roxspring:feature/limit-resource-families

### DIFF
--- a/chapters/resources.adoc
+++ b/chapters/resources.adoc
@@ -176,12 +176,12 @@ To keep maintenance and service evolution manageable, we should follow
 and do not mix different business functionalities in same API
 definition. In practice this means that the number of resource families
 exposed via an API should be limited, where a resource family is defined
-as a set of highly related resources such as a collection, it's members
+as a set of highly related resources such as a collection, its members
 and their direct subresources. Collections of related sub-resources 
 constitute a separate family for the purpose of limiting.
 
 For example, the resources here would be counted as three resource
-families, one for customers, one for the addressess, and one for the
+families, one for customers, one for the addresses, and one for the
 customers' related addresses:
 * /resources
 
--- a/chapters/resources.adoc
+++ b/chapters/resources.adoc
@@ -180,10 +180,13 @@ as a set of highly related resources such as a collection, its members
 and their direct subresources. Collections of related sub-resources 
 constitute a separate family for the purpose of limiting.
 
-For example, the resources here would be counted as three resource
+For example, the resources below would be counted as three resource
 families, one for customers, one for the addresses, and one for the
-customers' related addresses:
-* /resources
+customers' related addresses. We consider `/customers/{id}/preferences`
+part of the `/customers` family because it has a one-to-one relation to
+the customer without an additional identifier. We consider `/addresses`
+and `/customers/{id}/addresses` as separate resource families because 
+there's no reliable way to be sure they are the same.
 
 [source,http]
 ----
--- a/chapters/resources.adoc
+++ b/chapters/resources.adoc
@@ -169,24 +169,24 @@ directly, for instance:
 ----
 
 [#146]
-== {SHOULD} Limit number of Resource families
+== {SHOULD} Limit number of Resource types
 
 To keep maintenance and service evolution manageable, we should follow
 "functional segmentation" and "separation of concern" design principles
 and do not mix different business functionalities in same API
-definition. In practice this means that the number of resource families
-exposed via an API should be limited, where a resource family is defined
-as a set of highly related resources such as a collection, its members
-and their direct subresources. Collections of related sub-resources 
-constitute a separate family for the purpose of limiting.
+definition. In practice this means that the number of resource types
+exposed via an API should be limited. In this context a resource type is
+defined as a set of highly related resources such as a collection, its
+members and any direct sub-resources.
 
 For example, the resources below would be counted as three resource
-families, one for customers, one for the addresses, and one for the
-customers' related addresses. We consider `/customers/{id}/preferences`
-part of the `/customers` family because it has a one-to-one relation to
-the customer without an additional identifier. We consider `/addresses`
-and `/customers/{id}/addresses` as separate resource families because 
-there's no reliable way to be sure they are the same.
+types, one for customers, one for the addresses, and one for the
+customers' related addresses. We consider `/addresses` and
+`/customers/{id}/addresses` as separate resource families because
+there's no reliable way to be sure they are the same. We consider
+`/customers/{id}/preferences` part of the `/customers` resource type
+because it has a one-to-one relation to the customer without an
+additional identifier.
 
 [source,http]
 ----
--- a/chapters/resources.adoc
+++ b/chapters/resources.adoc
@@ -189,7 +189,7 @@ customers' related addresses:
 ----
 /customers
 /customers/{id}
-/customers/{id}/some-customer-action
+/customers/{id}/preferences
 /customers/{id}/addresses
 /customers/{id}/addresses/{addr}
 /addresses
--- a/chapters/resources.adoc
+++ b/chapters/resources.adoc
@@ -169,14 +169,35 @@ directly, for instance:
 ----
 
 [#146]
-== {SHOULD} Limit number of Resources
+== {SHOULD} Limit number of Resource families
 
 To keep maintenance and service evolution manageable, we should follow
 "functional segmentation" and "separation of concern" design principles
 and do not mix different business functionalities in same API
-definition. In this sense the number of resources exposed via API should
-be limited - our experience is that a typical range of resources for a
-well-designed API is between 4 and 8. There may be exceptions with more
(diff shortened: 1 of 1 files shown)


ePaul:
:+1:

chapters/resources.adoc
```diff
@@ -169,14 +169,35 @@ directly, for instance:
...
+For example, the resources here would be counted as three resource
+families, one for customers, one for the addressess, and one for the
+customers' related addresses:
+* /resources
+
+[source,http]
+----
+/customers
+/customers/{id}
+/customers/{id}/some-customer-action
```
maxim-tschumak: `action` sounds like RPC, we try to avoid this

roxspring: `/customers/{id}/preferences` seem more reasonable?

maxim-tschumak: Yes

chapters/resources.adoc
```diff
@@ -169,14 +169,35 @@ directly, for instance:
...
+/customers/{id}/addresses
+/customers/{id}/addresses/{addr}
+/addresses
```
maxim-tschumak: This is maybe a bit misleading in this example: I see `addresses` as **one** resource family with multiple endpoints/locations since the (sub)collection names are the same.

chapters/resources.adoc
roxspring: Done.

chapters/resources.adoc
```diff
@@ -169,14 +169,35 @@ directly, for instance:
...
-be limited - our experience is that a typical range of resources for a
-well-designed API is between 4 and 8. There may be exceptions with more
+definition. In practice this means that the number of resource families
+exposed via an API should be limited, where a resource family is defined
+as a set of highly related resources such as a collection, it's members
+and their direct subresources. Collections of related sub-resources 
+constitute a separate family for the purpose of limiting.
+
+For example, the resources here would be counted as three resource
+families, one for customers, one for the addressess, and one for the
```
maxim-tschumak: I see `addresses` and `customer's addresses` as the same resource family.

chapters/resources.adoc
roxspring: Over in #356 @tfrauenstein said:
> For definition and checking simplicity reasons, I suggest to count a resource type exposed as top-level and as sub-level resource twice. This design is not often applied and quite exotic.

I was just trying to make this explicit by showing the addresses twice and it being counted twice.

If we're going to conclude that that `/addressses` and `/customers/{id}/addresses` are a single family then I'd like to explicitly define the rules that drive the conclusion so that the Zally implementation can implement it unambiguously. (Not that "ease of Zally implementation" should be a driving consideration in writing guidelines, but I think it's a useful influence)

To me it seems we need to pick one of the following to be true:
1. `/addressses` and `/customers/{id}/addresses` are 2 separate resource families (easiest to implement)
2. All resource families ending `/addressses` are combined to count as 1 family (pretty easy, suspect could have false positives)
3. All resource families using the matching schema type are combined to count as 1 family (not sure about checking schema equality)
4. Some other heuristic I haven't considered.

chapters/resources.adoc
maxim-tschumak: typo: `addressess` -> `addresses`

chapters/resources.adoc
```diff
@@ -169,14 +169,35 @@ directly, for instance:
...
 
 To keep maintenance and service evolution manageable, we should follow
 "functional segmentation" and "separation of concern" design principles
 and do not mix different business functionalities in same API
-definition. In this sense the number of resources exposed via API should
-be limited - our experience is that a typical range of resources for a
-well-designed API is between 4 and 8. There may be exceptions with more
+definition. In practice this means that the number of resource families
+exposed via an API should be limited, where a resource family is defined
+as a set of highly related resources such as a collection, it's members
```
dehora: "it's members" -> "its members"

chapters/resources.adoc
```diff
@@ -169,14 +169,35 @@ directly, for instance:
 ----
 
 [#146]
-== {SHOULD} Limit number of Resources
+== {SHOULD} Limit number of Resource families
```
tkrop: While I support the choice of `resource family` over `resource type`, I think it is not intuitively clear how it is defined for readers. Therefore, I think, we have to provide somewhere a clear definition, what a `resource family`.

Something like:

A `resource family` is group of highly related `resource types` used to describe the same set of resources.

May be it will be good to have an example:

The `customer resource family` is defined by a general customer resource type describing a resource in response to a `GET/PUT /customers/{id}` request, a list customer resource type describing the resources in response to a `GET /customers?type=new` request, a create customer resource type describing the resource in the context of a `POST /customers` request, and a customer preferences resource type describing the separated aspect of customer preferences in response to a `GET /customers/{id}/preferences`.

chapters/resources.adoc
```diff
@@ -169,14 +169,35 @@ directly, for instance:
...
+as a set of highly related resources such as a collection, its members
+and their direct subresources. Collections of related sub-resources 
+constitute a separate family for the purpose of limiting.
+
+For example, the resources here would be counted as three resource
+families, one for customers, one for the addresses, and one for the
+customers' related addresses:
+* /resources
+
+[source,http]
```
tkrop: I have the feeling the example is missing concepts of http methods (and media-types).

chapters/resources.adoc
```diff
@@ -169,14 +169,35 @@ directly, for instance:
...
+----
+/customers
+/customers/{id}
+/customers/{id}/preferences
```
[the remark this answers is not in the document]
roxspring: How's this:
> For example, the resources below would be counted as three resource
families, one for customers, one for the addresses, and one for the
customers' related addresses. We consider `/customers/{id}/preferences`
part of the `/customers` family because it has a one-to-one relation to
the customer without an additional identifier.

chapters/resources.adoc
tfrauenstein: Hi Rob, thx. for the PR!
I'm struggling with the term 'resource family' -- it is too general.
Let's call it 'resource type' -- it is more specific and commonly used in this context.

chapters/resources.adoc
tfrauenstein: 1) I would not mention media types here, but add hint that a resource type typically is exposed via different http methods (endpoints). 

2) I would explicitly say, that we count top-level and sub resources. 

3) In the example we have 4 resources, 2 top level resources 'customer' and 'addresses', and 2 sub resources 'preferences' and 'addresses'. Does not match current wording.

chapters/resources.adoc
roxspring: Well obviously we disagree :)

But there's no point dragging it out further so I've switched the terminology back in the hope of closing down the conversation and PR quicker!

chapters/resources.adoc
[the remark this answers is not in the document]
tkrop: Fine for me.

chapters/resources.adoc
tkrop: @tfrauenstein I agree with rob. I initial although had my doubts, but when I considered `resource type` as an alternative, I realized that intuitively this better covers the distinct shape of the resources in a resource family, i.e. the *list* resource, the *input* resource, the *output* resource, the extracted resource *aspects* etc.

chapters/resources.adoc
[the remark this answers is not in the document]
tkrop: But, I would put this sentence below the example.

chapters/resources.adoc
tkrop: 1. I agree with (1), still I would write GET/PUT/POST in front.
2. Seems not to be the consent as we (Rob, me, and others) would not count `preferences` as a resource.
3. Reflects the different understanding of 2.

roxspring:
@tkrop, @tfrauenstein, @maxim-tschumak - trying to pull the unresolved conversations out hidden sections...

* I've had another edit and now the example is introduced, presented and only then explained - hopefully clearly showing why the numbers and text match.
* I very nearly added some example POST|GET|PUT|DELETE prefixes but noticed there's only one example within this chapter that uses those where most don't. For now I've left them off to stick with the majority.
* I got a bit frustrated with the circular discussions but am determined to get back to a constructive frame of mind ;)

I think the not-entirely-resolved topics are:
1. `/customers/{id}/preferences` is part of `/customers`
2. `/customers/{id}/addresses` and `/addresses` are separate resources
3. "Resource Types" vs "Resource Families"
4. Example has no METHOD prefixes.

Please take another look and let's see if we can agree 1st on the interpretation, 2nd on the terminology, and 3rd work through the explanation to make sure it's clear enough.

tkrop:
👍

maxim-tschumak:
:+1:

PR #373, opened 2018-04-10
Change x-api-id to be required

### LINKED ISSUE #371: API-id must be provided
We should change rule #215 from "should" to "must" ( see http://zalando.github.io/restful-api-guidelines/index.html#215 ).

The API id is crucial to identify APIs throughout their life cycle; it is required, for instance, to support API Review tracking, API Linter tracking, API History via Portal consistently over the API change evolution, and is a crucial metadata for our API management procedures.

And the API-title is not an equivalent substitute for API-id since it eventually gets updated.

tkrop:
Fixes #371.

I have changed the order of extension rules because I think to have the ID extension as first extension feels more natural. Any opposing opinion?

### DIFF
--- a/chapters/meta-information.adoc
+++ b/chapters/meta-information.adoc
@@ -14,11 +14,8 @@ to allow for API management:
 
 Following OpenAPI extension properties *must* be provided in addition:
 
-- `#/info/x-audience` intended target audience of the API (<<219, see rule 219>>)
-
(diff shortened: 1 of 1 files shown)


maxim-tschumak:
:+1:

ePaul:
:+1:

ePaul:
:+1:

duergner:
:+1:

Issue #245, opened 2017-06-15
Highlighting of sort key(s) in data change events


SebastianNelk:
The actual event schema definition does not explicitly state the sort key of data change events.

Please create a guideline to highlight the sort key in the event schema definition taking into concern that this could be distributed across multiple properties.

This is important to enable consumers to ensure the right event order during consumption.

Thanks!

ePaul:
This is related to rule [203](http://zalando.github.io/restful-api-guidelines/#203). I guess this rule should adapted here to say how to indicate the ordering key in the event schema definition.

codekultur:
We already have a field named "parent_eids" in the event metadata section, which can be used for partial and absolute ordering (depending on the interpretation of the sender). IMHO it makes sense to use this field for ordering. Maybe we just need to extend the metadata by a specification if the order defined via this parent id, is either a partial or absolute order.

What do you think?

SebastianNelk:
Following the guildlines:
http://zalando.github.io/restful-api-guidelines/index.html#198

This is what parent_eids is ment for:
    parent_eids:
      description: |
        Event identifiers of the Event that caused the generation of
        this Event. Set by the producer.
      type: array
      items:
        type: string
        format: uuid
      example: '105a76d8-db49-4144-ace7-e683e8f4ba46'

So using it as order filed doesn't match for my understanding. Metadata should be extended by something like:
    order_attributes:
      description: |
        List of attributes to be used for ordering in event consumption. Set by the producer.
      type: array
      readOnly: true
      items:
        type: string
       example: ["cSKU", "timestamp"]

codekultur:
Well, it depends on what "Event identifiers of the Event that caused the generation of
this Event." means or how it is interpreted. I would read it as the event that is the predecessor of the event in question, which IMHO defines an order.

Or is the parent_eids field used for something else and I misinterpreted the description (maybe we should then think about sharpening the description).

SebastianNelk:
It's clearly highlighting it have to be eids like in business event of "order placed" created afterwards "billing provider contacted". What we are looking for is a order of consumption of the actual event it is defined for.

ePaul:
If the eid is an UUID created by one of the usual processes, ordering by it does not work (because creation time is not conserved). If the eid is a counter just formatted in UUID format, we can order by it.

The `parent_eids` field works for grouping events which are related to each other (like the example given by Sebastian), but not so much for ordering unrelated events for the same entity.

[...]

PR #374, opened 2018-04-10
Indicate ordering key in event type definition (#245)

ePaul:
This adds an `ordering_key_fields` property to the event type definition (and mentions it in Rule 203), solving #245 (hopefully).


SebastianNelk:
:+1:

tfrauenstein:
👍

rcillo:
@ePaul should we already define how to behave in case the ordering fields need to be modified?

chapters/events.adoc
```diff
@@ -229,6 +229,20 @@ EventType:
...
+          description:
+            indicates by which of the fields in the schema the events can be ordered.
+            In most cases, this would have just a single item (the path of the field
+            by which this is to be ordered), but can have multiple items, in which case
+            those are considered as a compound key, with lexicographic ordering (first
+            item is most significant).
+          items:
+            type: string
+            description:
+              Indicates a single ordering field. This is a JsonPointer, which is applied
```
ePaul: TODO: pointer to field of primitive type (String or number)

whiskeysierra: Maybe worth to hint that ISO 8601 are not comparable if the timezone offset changes. Shouldn't be a problem if the event producer sticks to UTC.

ePaul:
@rcillo I think we don't need to care about special handling for modifying this (or any compatibility rules). It should still be checked the same way it was checked originally, if at all.

dehora:
@rcillo @SebastianNelk Can you confirm 

1. this allows event designs where the order is a single field in the event and doesn't always require looking into the payload? We will need to be able to to support opaque content in the future and I'd like to be sure this feature isn't over optimising for the lake/analytics part of the system.

1. This is getting implemented in Nakadi? Documenting an event option that isn't supported in real world systems is going to confuse people

@tfrauenstein I'd like to delay merging until the above are clarified.

tfrauenstein:
@dehora 
1) yes, merge will be done not before Nakadi live
2) sorry, but I'm not sure whether I have understood your question:

The idea was, that Nakadi also provides check functionality as to make sure that the order-key json path value is consistent with event scheme (i.e. is linking to valid attribute defined in scheme). And yes, also order-key path for meta attributes (outside the scheme) should be also supported. 

During event instance creation, Nakadi only checks consistency of payload with scheme.

dehora:
> sorry, but I'm not sure whether I have understood your question:

It's to avoid a situation where there is a capability in the API guidelines but not in the Nakadi project software or production deployment. @SebastianNelk's indicated this won't merge until the feature is available for use, so I think this ok; other than that would suggest @ePaul block it with -1 review so it doesn't accidentally get merged.

rcillo:
@tfrauenstein sure. I'll make sure it will be possible to change. Regarding implementation, I started coding it. I stopped to discuss within the team if this attribute should have some dependency on `partition_key_fields`.

We are afraid that users who are used to rely on the fact that order is defined by the position of the event in the partition could misuse this new attribute. So we are considering to limit the usage of this field to random partition strategy, so that it becomes clear that there is no order based on position, but rather the user is expected to implement some reordering after consumption.

We were also afraid that users could assume that Nakadi would perform some kind of ordering for them. Even with properly defined API, the truth is that many developers rely heavily on the client code, rather the API definition. So we believe that this field could cause some usability issues.

In order to mitigate this risk, I have been thinking about a name that will clarify that this field is merely informational.

SebastianNelk:
@rcillo This field is only informational but required to properly consume the event as this is the truth in case of data change events to really get the latest state of the represented entity of the event.

rcillo:
I understand the use case and the meaning of the field. What I'm sharing with you is the point of view of someone who supports Nakadi users. This field needs to fit within the already existent Nakadi attributes and must not cause confusion or ambiguities. In Nakadi, order is defined by position. If an event comes after another, it means that they were published one after the other, which means that's their natural order.

This order is only guaranteed for the same partition. This leads to a situation of partial order. Consumers cannot compare the order across partitions.

With this new field it becomes possible for consumers to compare the oder of any 2 events, regardless of their partition or position. So this new field conflicts with existent semantics of ordering and for that reason I'm saying that we should be careful not to cause confusion among users.

After introducing this field, users need to be aware that the position no longer has anything to do with ordering, in case this `ordering_key_field` is used. They must now figure out the order based on the value of a field in the events. Given this field can change over time, producer would now be able to constantly break the contract regarding order.

In Nakadi, we design everything to work as a contract so that consumers can be implemented in a stable way, that keeps operating normally as the event type evolves. But this new field breaks with this design rule that we put so much effort to follow.

ePaul:
@rcillo I'm open for other name proposals, as well as proposals about how to document it clearer that this ordering is not to be done by Nakadi, but an information from the producer to the consumers.

I don't think that this should only be allowed for "random" partitioning (if you shard your consumers, you still want to have everything for one entity in the same partition, even if the ordering is wrong).

dehora:
> This field needs to fit within the already existent Nakadi attributes and must not cause confusion or ambiguities.

The underlying challenge seems to be that the category structures, from the point of view of the Nakadi project are non-extensible. We've discussed this in the past, and the Nakadi project stewards have indicated Nakadi is closed for extension by default. The consequence is no real evolution or learning is possible by producers and consumers without involving the event middleware.

So for me, the problem to solve, is that end to end application semantics, which imvho should be owned by the producer and consumer are couple with the transport (or if you prefer because it's an API, transfer) semantics of the broker project.  My concern is that each innovation in the events space is hobbled by requiring multiple actors and multiple to agree, where that agreement isn't strictly necessary with a different design approach that separates concerns. This goes both ways, the Nakadi project needs flexibility to innovate on underlying systems design without having to cater for application semantics. See also [rfc3724](https://www.ietf.org/rfc/rfc3724.txt).  


@rcillo @tfrauenstein @ePaul @tkrop @whiskeysierra @SebastianNelk   I'd love to hear how you think we can make progress on this. (On the other hand having the event design in the API guidelines is super useful because it forces us to think this through :)

SebastianNelk:
Correct me if I'm wrong, but this is a producer -, not a nakadi, attribute and very important missing piece of information which has to be part of the contract between producer and consumer to ensure to have the latest state per entity as this can't be ensured by the order events are made available in Nakadi.

ePaul:
I think we have two things here:

1. Nakadi guarantees that events (in a partition) are delivered to consumers in the same order as they were submitted to this partition. This is a useful thing to have, and I don't want to abolish it.

2. Nakadi can't guarantee that events will be consumed (i.e. taken action upon) in the same order as they originally happened, because there might be things happening outside of Nakadi's sphere of influence which change that order – caused by bugs or other misfunctions.

   As an example, if there is a bug which makes the producer create events which are not acceptable by Nakadi, the producer can't submit this event. One possible strategy is to just put it in some local storage, and continue submitting other events, until some engineer looks at the event, fixes it (and hopefully the bug) and it will be submitted again. (The alternative, just stopping to submit any events at all, is often not appropriate.)

   The same can happen at the consumer side – if the consumer can't handle an event properly due to a bug, it can put is just into some local storage, and continue consuming other events, until the bug is fixed and it can try again. (The alternative would be to just stop committing anything to Nakadi, with the effect that it will get the same event again and again, and no later events can be consumed. This is often not appropriate, too.)

So given that there are reasons outside of the sphere of Nakadi's influence which might reorder events, we (as consumers) still want to be able to restore the original order after (or during) consumption. The API guidelines (in [rule 203](https://opensource.zalando.com/restful-api-guidelines/#203)) already stated that producers should provide a means for this by providing some ordering attribute, and this PR just makes it possible to document this ordering in the event type definition. (So this is for the *event type registry* part of Nakadi, not the transport part.)

Of course, we should consider what happens if a producer changes the ordering attribute documentation, and whether this can be considered a "compatible change". I'm open for proposals regarding this.

rcillo:
@ePaul  thank you for the detailed explanation. We are adding it to Nakadi at this moment https://github.com/zalando/nakadi/pull/892. My team's contribution to this discussion (and the PR I mentioned) is to bring the point of view of a team that spends at least 20% of its time supporting more than 50 teams who uses this service. Please, take a look at the PR and let us know if it's clear enough.

chapters/events.adoc
```diff
@@ -229,6 +229,20 @@ EventType:
           description: |
               The schema as string in the syntax defined in the field type.
           type: string
+        ordering_key_fields:
```
v-stepanov: I think you put this field on a wrong level. You made it a property of "schema" field. As I understand it is intended to be a root-level property of event-type instead.

tfrauenstein:
Thx. for the contributions. We now have clarified, that correct providing of ordering-key is a producer obligation without impact to ordering on event transport layer as provided by Nakadi and orthogonal to Nakadi partitioning. 

But sorry, some further basic aspects need clarification: 

a) There still seems to be different views on the "scope of the ordering-key": Is it for all events of the event type and/or is it for all events related to a entity instance? There might be use cases for both. Definitely a major use case is "data entity change replication", i.e. ordering-key only strictly increasing per data instance / entity. In this case, we have to additionally provide information to identify the entity instance, i.e. to introduce "ordering-instance-key" (just a name proposal) optional and additionally to "ordering-key".

b) Event type definition in Nakadi API and API Guideline are currently redundant.
I see no technical feasible short term path to avoid this, do you?
If not, we have to make sure that both PRs are consistent and in sync before merged. 
This is easier, if we avoid parallel discussion and early PR in Nakadi.
Hence, I propose to merge proposed changes into this PR and continue discussion here. 

c) Also related to Bills comment: We have here the situation that ordering on application layer, different (and transparent) than on transport infrastructure Nakadi layer is a quite generic problem and having standards how to provide this information is useful for other generic services like Data Platform / BI data sync. 
However, E2E application semantics is usually defined via description elements of the event schema. Shouldn't we go this path here and define a standard mark-up in scope of the event attribute description avoiding Nakadi Event Schema extension at all?

chapters/events.adoc
ePaul: You are right. (Maybe I was confused by there being a schema inside the schema?)

I'll fix this.

chapters/events.adoc
ePaul: Added.

ePaul:
@tfrauenstein 

> b) Event type definition in Nakadi API and API Guideline are currently redundant.
I see no technical feasible short term path to avoid this, do you?
If not, we have to make sure that both PRs are consistent and in sync before merged.
This is easier, if we avoid parallel discussion and early PR in Nakadi.
Hence, I propose to merge proposed changes into this PR and continue discussion here.

We want to avoid mentioning the specific implementation (Nakadi) in these guidelines, but in the Nakadi API they are useful. So we won't have exactly the same text here anyways.

But I'll try to see what is there and can be usefully applied here.

ePaul:
In the meeting:
* review what Nakadi did, change this to not be in conflict → @epaul
* Ask Nakadi whether using JsonPointer is possible, if not, follow Nakadi's lead about dot-separated field names → @tfrauenstein

SebastianNelk:
Thanks to @ePaul for following through on my idea/request!

tkrop:
:+1:

tfrauenstein:
👍

PR #384, opened 2018-05-11
moved 'API First' to 'Principles' section (editorial)

tfrauenstein:


chapters/general-guidelines.adoc
```diff
@@ -7,84 +7,31 @@ The titles are marked with the corresponding labels: {MUST},
...
-is in conflict with the bad practice of publishing API definition and
-asking for peer review after the service integration or even the service
-productive operation has started. It is crucial to request and get early
-feedback — as early as possible, but not before the API changes are
-comprehensive with focus to the next evolution step and have a certain
-quality (including API Guideline compliance), already confirmed via team
-internal reviews.
+You must follow the <<api-first, API First Principle>>, more specifically: 
+
+* You must define APIs outside the code first <<101, using OpenAPI as specification language>>
```
duergner: As we're about to change that section now - I would like to challenge this **MUST** role. In order to improve the chances to have API spec match the actual implementation I still see huge benefits if we would allow to e.g. have the API defined already via code. Obviously no implementation yet but just *interfaces* and *data classes* but having it this way would ensure that code changes are always reflected in API changes.

whiskeysierra: I agree. We shouldn't force designers to use a specific way or tool to define an API.

tfrauenstein: What is intended by the rule is: 
You 1st should define the API and get feedback, before you *implement* the service. 
Yes, API definition can be executed via generation + enrichment from programming language interfaces and data classes. But I would not explicitly mention it here, since there also is high risk that API definition is finally just generated from complete implementation.  

Hence, my proposal for the rule:
You must define APIs first, outside the implementation code scope, using OpenAPI...
or
You must define APIs first, before its coding its implementation, using OpenAPI...

dehora: > I would like to challenge this MUST role.

Ok, but let's not do that in a large scale refactoring pr. Changing levels should be done via a focused pull request.

duergner: @dehora agree.

@tfrauenstein sounds like a good wording. I totally agree that we should have the API defined, aligned and reviewed before it's implemented.

duergner:
:+1:

tkrop:
:+1:

PR #390, opened 2018-05-29
Fix link to infoq.com

ioggstream:
## I expect
the link to 
https://www.infoq.com/news/2014/03/rest-at-odds-with-web-apis
to work

## instead
it is broken


maxim-tschumak:
Thanks!

maxim-tschumak:
:+1:

tkrop:
:+1:

PR #392, opened 2018-06-05
Add a link from rule 111 to rule 216.

ePaul:
Both are about additionalProperties, and might seem to contradict each other.

### DIFF
--- a/chapters/compatibility.adoc
+++ b/chapters/compatibility.adoc
@@ -171,9 +171,8 @@ PUT, POST and PATCH requests.
 API formats must not declare `additionalProperties` to be false, as this
 prevents objects being extended in the future.
 
-Note that this guideline concentrates on default extensibility and does
-not exclude the use of `additionalProperties` with a schema as a value,
-which might be appropriate in some circumstances.
+Note that this guideline concentrates on default extensibility for "normal objects"
+and does not exclude the <<#216, use of `additionalProperties` for maps>>.
 
 [#112]
 == {SHOULD} Used Open-Ended List of Values (x-extensible-enum) Instead of Enumerations


chapters/compatibility.adoc
```diff
@@ -171,9 +171,8 @@ PUT, POST and PATCH requests.
 API formats must not declare `additionalProperties` to be false, as this
 prevents objects being extended in the future.
 
-Note that this guideline concentrates on default extensibility and does
-not exclude the use of `additionalProperties` with a schema as a value,
-which might be appropriate in some circumstances.
+Note that this guideline concentrates on default extensibility for "normal objects"
```
tkrop: Should we give this sentence more emphasis by using the usual **Note:**?

"normal objects" is a very fuzzy term. Can we use something more determined? May be *well-defined objects* to express that their properties are defined.

chapters/compatibility.adoc
```diff
@@ -171,9 +171,8 @@ PUT, POST and PATCH requests.
...
+and does not exclude the <<#216, use of `additionalProperties` for maps>>.
```
tkrop: May be ```use of `additionalProperties` *extensible maps object*``` to more clearly distinguish from *well-defined objects*. It might be good to adjust the wording in rule 216 too.

ePaul: I'm not sure that "extensible map objects" is the right name ... those maps are not extensible in the way the "normal objects" are, as all their values have the same type (which itself can be extended).

ePaul:
@tfrauenstein already changed this inbetween in #393 (so now we have a conflict here).

Not sure whether we need more?
Or should I still try to add the improvements proposed by @tkrop?

tkrop:
@ePaul I would say close.

PR #395, opened 2018-06-15
Improve title requirements

tkrop:
To make clear, that an API should contain not only technical terms but describes the main business purpose.


whiskeysierra:
:+1:

maxim-tschumak:
:+1:

PR #400, opened 2018-06-19
Added hint to sequential identifiers

### LINKED ISSUE #383: Rule 144 contains questionable advice
[Rule 144: Should: Only Use UUIDs If Necessary](https://opensource.zalando.com/restful-api-guidelines/#144) currently lists the following as a negative aspect:

> not ordered along their creation history and no indication of used id volume

I'd argue that this can and has been a downside for us in the past. Having a sequence that increases monotonically without gaps is not desirable in most cases, since it allows someone from the outside to regularly measure velocity and derive some insights that wouldn't otherwise be possible.

whiskeysierra:
Fixes #383


ePaul:
:+1:

whiskeysierra:
:+1:

Issue #394, opened 2018-06-07
Separate public guidelines and Zalando-internal concerns


ePaul:
Currently our guidelines are represented as "guidelines everyone can adopt", while at the same time they contain many Zalando-specific issues, and even internal links at several places.

Similarly, the evolving of those guidelines happens in a semi-open-source model: All changes have pull-requests here in Github, but at the same time we discuss them internally in our bi-weekly API Guild meetings which are only accessible to Zalando employees.

The guidelines are also used as a base in some other Zalando Open-Source projects (e.g. [Nakadi](https://github.com/zalando/nakadi/)).

Reusing those guidelines elsewhere is currently difficult, though there are [142 forks](https://github.com/zalando/restful-api-guidelines/network/members), of which at least some have edited the guidelines substantially to adapt them to their uses.

Can we make this reuse easier by separating Zalando-specific parts out? Or have them separately in the code, but somehow merged together for the viewed version?

tkrop:
I think, we could allow external collaborators to join the meeting. We need just announce it a bit more.

Regarding the question to separating internal information: I see an advantage of keeping the guideline easy applicable for us. Having a separate internal guideline application document makes it much more difficult. At the moment I would vote against it.

ddufourboivin-zalando:
I've started reading through the rules and had the same though as @ePaul . 

Maybe, without removing the internal links completely, we could visually split them. A frame and/or an Icon (tiny Zalando logo) pointing out to Zalando employees that there is extra information just for them there.

OR

We could have an internal document with the Zalando-only additions to those rules and the open-source document very quickly links to specific sections of that document, yielding only a few "Zalando employees, read further here" with a link to that document. That would minimise the Z-related content in the OSS version.

tkrop:
Zalando employee boxes sounds like a nice approach, but this would even emphasize this more.

ddufourboivin-zalando:
But having a more obvious visual about it also allo non-Zalandos to skip it more easily.

> This colour or design? Don't even read it.

ddufourboivin-zalando:
I understand your point and agree.

👍 It's a lot cleaner and more open-source-friendly if there are none.
👎 It means 2 distincts documents to read and some effort to "synchronise the reading" between them.

The 👍however overweight the 👎 here in my opinion since this **is** an open source project and the OSF-Factor (_Open Source Friendlyness_, just created that term...) should go first.

ePaul:
From the Meeting:

[...]

PR #401, opened 2018-06-20
Removes private internal Zalando documents.

dehora:
Going ahead with removing private links while https://github.com/zalando/restful-api-guidelines/issues/394 gets resolved.

### DIFF
--- a/chapters/api-operation.adoc
+++ b/chapters/api-operation.adoc
@@ -12,9 +12,7 @@ to profit from the API management infrastructure.
 An API is published by copying its **OpenAPI specification** into the reserved
 **/zalando-apis** directory of the **deployment artifact** used to deploy the
 provisioning service. The directory must only contain **self-contained YAML**
-**files** that each describe one API (see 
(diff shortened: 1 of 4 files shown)


dehora:
:+1:

whiskeysierra:
:+1:

ddufourboivin-zalando:
# ⚠️ WARNING / SUGGESTION
We should keep a trace of those internally and prepare a _Companion Document_, with the same format as those guidelines, that indicates what should be read additionally for Zalando employees.

dehora:
> We should keep a trace of those internally and prepare a Companion Document,

@ddufourboivin-zalando might be an idea to open an issue for that or added it to  #394 so it doesn't get lost here

dehora:
:+1:

tfrauenstein:
@dehora I can live with deletion of link 1 - 3 -- they are less relevant for efficient rule application.  However, in the functional naming section we should link to our functional name inventory service (such that teams can efficiently create names). We will add this.

PR #402, opened 2018-06-21
398 quotation marks

### LINKED ISSUE #398: Replace German quotation marks by English quotation marks
Example from the [DELETE](https://opensource.zalando.com/restful-api-guidelines/#delete) article:

> The semantic is best described as »please delete the resource identified by the URL«.

Those are German styled quotation marks. If this documentation is in English, it should read:

> The semantic is best described as "please delete the resource identified by the URL".

## Task

Search in all of the documentation and fix all of them, not only the example provided here.

ddufourboivin-zalando:
fix #398 

Two commits:
1. Replace the `»...«` by `"..."`
1. Replace the `"..."` by `“...”` everywhere

Options:
1. We keep the ASCII quotes `"..."`
    - Pros
      - Easier to type using any keymap
      - Probably what is already used when pasting text from other sources
    - Cons
      - Not typographically accurate
      - There is already a mix of `"..."` and `“...”` in this documentation
1. We switch everything to `“...“`
    - Pros
      - Typographically accurate
      - Align the entire documentation to one quotation standard
    - Cons
      - Bigger diff (not really a problem)
      - Hard to maintain (people will probably have the reflex of using `"..."` in the future).

I did it in 2 different commits, which will be useful if we go with the 1st option (rollback the 2nd).

Let's discuss!

### DIFF
--- a/chapters/events.adoc
+++ b/chapters/events.adoc
@@ -795,7 +795,7 @@ by a data change event and a RESTful API.
 
 Event consumers must be able to process duplicate events.
 
-Most message brokers and data streaming systems offer “at-least-once”
+Most message brokers and data streaming systems offer "at-least-once"
 delivery. That is, one particular event is delivered to the consumers
 one or more times. Other circumstances can also cause duplicate events.
 
--- a/chapters/events.adoc
+++ b/chapters/events.adoc
@@ -282,11 +282,11 @@ GeneralEvent:
 
 Event types based on the General Event Category define their custom
 schema payload at the top-level of the document, with the
-link:#event-metadata[`metadata`] field being reserved for standard
-information (the contents of link:#event-metadata[`metadata`] are
+<<event-metadata,metadata>> field being reserved for standard
+information (the contents of <<event-metadata,metadata>> are
 described further down in this section).
 
-In the example fragment below, the reserved link:#event-metadata[`metadata`]
+In the example fragment below, the reserved <<event-metadata,metadata>>
 field is shown with fields “a” and “b” being defined as part of the custom
 schema:
 
@@ -434,7 +434,7 @@ EventMetadata:
 
 Please note than intermediaries acting between the producer of an event
 and its ultimate consumers, may perform operations like validation of
-events and enrichment of an event's link:#event-metadata[`metadata`]. For
+events and enrichment of an event's <<event-metadata,metadata>>. For
 example brokers such as Nakadi, can validate and enrich events with
 arbitrary additional fields that are not specified here and may set default
 or other values, if some of the specified fields are not supplied. How
@@ -723,7 +723,7 @@ on the use of `additionalProperties`.
 
 The `eid` (event identifier) value of an event must be unique.
 
-The `eid` property is part of the standard link:#event-metadata[metadata]
+The `eid` property is part of the standard <<event-metadata,metadata>>
 for an event and gives the event an identifier. Producing clients must
 generate this value when sending an event and it must be guaranteed to
 be unique from the perspective of the owning application. In particular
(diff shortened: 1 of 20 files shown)


whiskeysierra:
Authors of the guidelines write in markdown and markdown is code. We very much need to stick to the obvious choice of `"..."`. What I could imagine is some post processing that rewrites quotes before publication.

ddufourboivin-zalando:
I personally do not have an opinion. Or if one, stick to `"..."`. This task started as _getting rid of the German-styled quotes_ and after a comment from @ePaul , went to _apply correct typography_.

I would vote for `"..."` because it is easier to maintain. But then, we need to stick to this styling and not have an hybrid, as it is at the moment.

chapters/events.adoc
```diff
@@ -286,8 +286,9 @@ link:#event-metadata[`metadata`] field being reserved for standard
 information (the contents of link:#event-metadata[`metadata`] are
 described further down in this section).
 
-In the example fragment below, the reserved `metadata` field is shown
-with fields “a” and “b” being defined as part of the custom schema:
+In the example fragment below, the reserved link:#event-metadata[`metadata`]
```
maxim-tschumak: <<event-metadata,metadata>> should also work

ddufourboivin-zalando: @maxim-tschumak : https://github.com/zalando/restful-api-guidelines/pull/404

maxim-tschumak:
:+1:

tkrop:
@ddufourboivin-zalando Thanks for the work. However, most of the `"..."` are in reality for emphasizing terms not for quoting. I would prefer these to use `_..._` instead (or `*...*` in strong cases). When it comes to quoting, it should always use `_"..."_` to keep it simple in code, even if I like the German quotes.

ddufourboivin-zalando:
@tkrop : I will rework this branch to use `"..."` everywhere (basically reversing the commit where I change them to `“...”`). Let's align on the character to use for quoting first and discuss the use of _quotation_ vs _emphasis_ in a next discussion (my proposal).

ddufourboivin-zalando:
# Next question ...

Which brings us to the next consideration.

*Should we align the existing `“...”` to the decided `"..."` or leave them be?*

Please vote by 👍 or 👎 this comment (smiley in up-right corner).

ePaul:
@whiskeysierra 
> Authors of the guidelines write in markdown and markdown is code. We very much need to stick to the obvious choice of "...".

I don't really follow that argumentation. Is it »Code should only ever use ASCII«?

whiskeysierra:
> I don't really follow that argument. Is it »Code should only ever use ASCII«?
> (I'm fine with using ASCII codes, I just want to understand the argument here.)

Yes.

tkrop:
@ePaul not entirely. It is although that this is a English document, where we should use punctuation marks (here quotation marks) as usually used in this language :).

tkrop:
:+1:

ddufourboivin-zalando:
Since it is a document, we _should_ be using the right quotation marks `“...”`, but I also think that it would be difficult to maintain consistency because we'll forever have the reflex of using ASCII/coding quotations `"..."` when we right. So I am completely good with this decision of using the ASCII ones.

There will be another discussion (initiated by @tkrop ) about using proper _emphasis_ instead of "quotations" in this documentation where it applies (most of the actual quoted text should indeed be emphasised instead). But this is out of scope here.

I will be pushing a last time with the decided "quotation marks" and then we can close this topic.

ddufourboivin-zalando:
Ready for re-review.

@whiskeysierra @tkrop @maxim-tschumak @epaul

Issue #397, opened 2018-06-19
Fix English grammar in DELETE article


ddufourboivin-zalando:
At the beginning of the [DELETE](https://opensource.zalando.com/restful-api-guidelines/#delete) article,

> DELETE request are used

should read

> DELETE requests are used

ePaul:
Just my opinion: For such small changes directly opening a PR is less overhead (and will result in faster resolution) than creating an issue.

tkrop:
Fixed in #403.

Issue #396, opened 2018-06-19
Fix formatting in PATCH article


ddufourboivin-zalando:
In the [PATCH](https://opensource.zalando.com/restful-api-guidelines/#patch) section, this

> - successful PATCH requests will usually generate 200 or 204 (if resources have been updated
> - with or without updated content returned)

should read

> - successful PATCH requests will usually generate 200 or 204 (if resources have been updated with or without updated content returned)

ePaul:
Likely a leftover from the ASCIIdoc conversion.

tkrop:
Fixed in #403.

PR #403, opened 2018-06-21
Proof reading

ddufourboivin-zalando:
fix 397
fix 396

# Findings fixed here
- https://github.com/zalando/restful-api-guidelines/issues/397
- https://github.com/zalando/restful-api-guidelines/issues/396
- adress-fields
  - The example [there](https://opensource.zalando.com/restful-api-guidelines/#address-fields) does not make sense. `addressee` has `required` fields that are not even in its `properties` list and a second `required` block. The first one should be deleted.
- https://opensource.zalando.com/restful-api-guidelines/#179
  - remove the comma in "an identical location to the requested URI, can be used to explicitly indicate that the returned"
- Add anchor to `Event Metadata.` section and link to it from where it is referenced `$ref: '#/definitions/EventMetadata'`.
- https://opensource.zalando.com/restful-api-guidelines/#213
  - change
    - >Event type names must, respectively should conform to the functional naming depending on the audience as follows (see Must/Should: Use Functional Naming Schema for details and definition):
  - to
    - >Event type names must (or should, see Must/Should: Use Functional Naming Schema for details and definition) conform to the functional naming depending on the audience as follows:

### DIFF
--- a/chapters/common-data-types.adoc
+++ b/chapters/common-data-types.adoc
@@ -137,13 +137,6 @@ below.
 addressee:
   description: a (natural or legal) person that gets addressed
   type: object
-  required:
-    - first_name
-    - last_name
-    - street
-    - city
-    - zip
-    - country_code
   properties:
     salutation:
       description: |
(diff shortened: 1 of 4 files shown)


dehora:
:+1:

PR #404, opened 2018-06-22
Use internal link syntax

ddufourboivin-zalando:
```
<<event-metadata,metadata>>
```
instead of
```
link:#event-metadata[`metadata`]
```


maxim-tschumak:
:+1:

ddufourboivin-zalando:
:

PR #405, opened 2018-06-22
Specify other names used for `component-internal` audience

ddufourboivin-zalando:
As agreed this morning with @tfrauenstein , @tkrop and @maxim-tschumak , we need to indicate in the API Guideline that the `component-internal` audience is also referred to, sometimes, as `product internal` or `team internal`.

### DIFF
--- a/chapters/meta-information.adoc
+++ b/chapters/meta-information.adoc
@@ -114,7 +114,8 @@ for discoverability, changeability, quality of design and documentation, as
 well as permission granting. We differentiate the following API audience
 groups with clear organisational and legal boundaries: 
 
-*component-internal* (aka team/product internal)::
+*component-internal*::
+  This is often referred to as *team internal* or *product internal*.
   The API consumers with this audience are restricted to applications of the
   same *functional component* which typically represents a specific *product* 
   with clear functional scope and ownership.
--- a/chapters/meta-information.adoc
+++ b/chapters/meta-information.adoc
@@ -115,7 +115,7 @@ well as permission granting. We differentiate the following API audience
 groups with clear organisational and legal boundaries: 
 
 *component-internal*::
-  This is often referred to as *team internal* or *product internal*.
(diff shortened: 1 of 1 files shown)


chapters/meta-information.adoc
```diff
@@ -114,7 +114,7 @@ for discoverability, changeability, quality of design and documentation, as
 well as permission granting. We differentiate the following API audience
 groups with clear organisational and legal boundaries: 
 
-*component-internal*::
+*component-internal* (aka team/product internal)::
```
dehora: This is fairly new metadata, can the guidelines not settle on a canonical name instead of supporting multiple names?

tfrauenstein: The canonical name should be still "component-internal".
The "aka" is a less technical term that covers the 90% use cases and is better understood.
But since it is obviously confusing in the headline, we should move this into in the description section, but very visible.

ddufourboivin-zalando: 👍 Moving to the paragraph.

ddufourboivin-zalando:
@dehora @tfrauenstein @tkrop : Please re-review.

chapters/meta-information.adoc
```diff
@@ -115,6 +115,7 @@ well as permission granting. We differentiate the following API audience
 groups with clear organisational and legal boundaries: 
 
 *component-internal*::
+  This is often referred to as *team internal* or *product internal*.
```
tfrauenstein: 👍 , would just add "API" --> This is often referred to as *team internal API* or *product internal API*.

maxim-tschumak:
:+1:

chapters/meta-information.adoc
ddufourboivin-zalando: @tfrauenstein : Done

ePaul:
:+1:

tkrop:
:+1:

PR #406, opened 2018-06-22
Do not provide examples returning a JSON array

ddufourboivin-zalando:
The [section about `ETags`](https://opensource.zalando.com/restful-api-guidelines/#_etag_header_with_if_match_header) provides examples that return JSON arrays.

To be consequent with the rules we encourage people to use, let's provide examples that do respect [Must: Always Return JSON Objects As Top-Level Data Structures To Support Extensibility [110]](https://opensource.zalando.com/restful-api-guidelines/#110).


tkrop:
:+1:

ePaul:
:+1:

ePaul:
Thanks, this must somehow have been slipped through.

PR #408, opened 2018-06-26
Change guideline to use only http status codes from list

tkrop:
In this PR I reformulated our guidance to restrict usage of HTTP status codes to those listed in the guideline that have a well defined semantic. The reason for this is:

* Creative usage of HTTP status codes makes it difficult for API consumers to understand the meaning. 
* HTTP Status codes outside of our set are normally not well understood - by API designers, but even more important by API consumers.
* Our list is already very extensive, and contains quite exotic HTTP status codes for usage. All status codes outside this list address even more exotic use cases.
* In the past we deleted at least HTTP status code 422 from the list, because we decided that its application is bad practice in almost all situations.

(Please ignore the reformatting part of this PR.)

### DIFF
--- a/chapters/http-status-codes-and-errors.adoc
+++ b/chapters/http-status-codes-and-errors.adoc
@@ -9,15 +9,15 @@ APIs should define the functional, business view and abstract from
 implementation aspects. Success and error responses are a vital part to
 define how an API is used correctly.
 
-Therefore, you must define **all** success and service specific error responses in your API specification.
-Both are part of the interface definition and
-provide important information for service clients to handle standard as well
-as exceptional situations. 
+Therefore, you must define **all** success and service specific error
+responses in your API specification. Both are part of the interface definition
+and provide important information for service clients to handle standard as
+well as exceptional situations. 
 
 
 **Hint:** In most cases it is not useful to document all technical errors,
-especially if they are not under control of the service provider. 
-Thus unless a response code conveys application-specific functional semantics or is used
+especially if they are not under control of the service provider. Thus unless
+a response code conveys application-specific functional semantics or is used
 in a none standard way that requires additional explanation, multiple error
 response specifications can be combined using the following pattern:
 
@@ -31,8 +31,8 @@ responses:
       $ref: 'https://zalando.github.io/problem/schema.yaml#/Problem'
 ----
 
-API designers should also think about a **troubleshooting board** as part of the
-associated online API documentation. It provides information and handling
+API designers should also think about a **troubleshooting board** as part of
+the associated online API documentation. It provides information and handling
 guidance on application-specific errors and is referenced via links from the
 API specification. This can reduce service support tasks and contribute to
 service client and provider performance.
@@ -41,8 +41,8 @@ service client and provider performance.
 [#150]
 == {MUST} Use Standard HTTP Status Codes
 
-You must only use standardized HTTP status codes and consistently with 
-their intended semantics. You must not invent new HTTP status codes.
+You must only use standardized HTTP status codes consistently with their
+intended semantics. You must not invent new HTTP status codes.
 
 RFC standards define ~60 different HTTP status codes with specific semantics 
 (mainly https://tools.ietf.org/html/rfc7231#section-6[RFC7231] and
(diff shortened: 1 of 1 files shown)


dehora:
:-1: the guidelines should be be subsetting HTTP like this.

chapters/http-status-codes-and-errors.adoc
```diff
@@ -54,13 +54,16 @@ https://en.wikipedia.org/wiki/List_of_HTTP_status_codes[Wikipedia] or
 via https://httpstatuses.com/) also inculding 'unofficial codes', e.g. used
 by popular web servers like Nginx.
 
-Below we listed the most commonly used and best understood HTTP status codes 
-(consistent with RFC standard). You may use HTTP status codes not listed here, 
-but in this case you must provide clear documentation in the API definition. 
-As long as there is no need to use codes not listed here, you 
-should not describe HTTP status codes to avoid risk of inconsistent definitions and 
-reduced readability due to overload with common sense information. 
+Below we list the most commonly used and best understood HTTP status codes
```
ddufourboivin-zalando: I would add a `,` after `codes`.

chapters/http-status-codes-and-errors.adoc
```diff
@@ -54,13 +54,16 @@ https://en.wikipedia.org/wiki/List_of_HTTP_status_codes[Wikipedia] or
...
+consistent with their semantic in the RFCs. APIs should only use these to
+prevent misconceptions that arise from less commonly used HTTP status codes.
 
+As long as your HTTP status code usage is well covered by the here defined
+semantic, you should not describe it to avoid an overload with common sense
```
ddufourboivin-zalando: _by the here defined semantic_ ==> _by the semantic defined here_

chapters/http-status-codes-and-errors.adoc
```diff
@@ -54,13 +54,16 @@ https://en.wikipedia.org/wiki/List_of_HTTP_status_codes[Wikipedia] or
...
+information and the risk of inconsistent definitions. Only if the HTTP status
+codes not in this list below or the usage requires additional information
```
ddufourboivin-zalando: _Only if the HTTP status codes not in this list below_ ==> _Only if the HTTP status code is not in the list below_

ddufourboivin-zalando: _or the usage_ ==> _or its usage_

whiskeysierra:
@dehora Can you elaborate on this?

chapters/http-status-codes-and-errors.adoc
```diff
@@ -54,13 +54,16 @@ https://en.wikipedia.org/wiki/List_of_HTTP_status_codes[Wikipedia] or
...
-should not describe HTTP status codes to avoid risk of inconsistent definitions and 
-reduced readability due to overload with common sense information. 
+Below we list the most commonly used and best understood HTTP status codes,
+consistent with their semantic in the RFCs. APIs should only use these to
+prevent misconceptions that arise from less commonly used HTTP status codes.
 
+As long as your HTTP status code usage is well covered by the semantic defined
+here, you should not describe it to avoid an overload with common sense
+information and the risk of inconsistent definitions. Only if the HTTP status
+codes is not in the list below or its usage requires additional information
```
maxim-tschumak: `codes` -> `code`

chapters/http-status-codes-and-errors.adoc
```diff
@@ -54,13 +54,16 @@ https://en.wikipedia.org/wiki/List_of_HTTP_status_codes[Wikipedia] or
...
+aside the well defined semantic, the API specification must provide an clear
```
maxim-tschumak: -> `a clear`

dehora:
>  Can you elaborate on this?

Sure. So it's one thing to provide design guidance in the form, "in this circumstance use this element of http (status code, header, method, etc)". It's another to subset http absent context, "here's a list of codes, and use only these codes, and by the way here's what we think they mean".

whiskeysierra:
@dehora Are you objecting to the rule as it was before already or is this about @tkrop's changes? I feel like the rule didn't change much in nature. It also doesn't forbid to use other http status codes, it just asks (or rather *agressively demands*) everyone who wants to use them to specify/document what the status code means to them, since it might not be well understood by most engineers.

Maybe I misunderstood the rule.

dehora:
Hey @whiskeysierra,

> the rule didn't change much in nature. It also doesn't forbid to use other http status codes,

Right, I guess I'm struggling to reconcile that with @tkrop's intent:  

> I reformulated our guidance to restrict usage of HTTP status codes to those listed in the guideline that have a well defined semantic. 

and this part of the patch


```diff
+ APIs should only use these to
+ prevent misconceptions that arise from less commonly used 
+ HTTP status codes. As long as your HTTP status code usage 
+ is well covered by the semantic defined here, you should not 
+ describe it to avoid an overload with common sense information 
+ and the risk of inconsistent definitions. 
```

That's a change in the guidelines, especially the first sentence.

tkrop:
@dehora @whiskeysierra We try to align the rule a bit, so that the Zally implementation gives good feedback about the best practices. In this context a sentence: _you should not use exotic status codes (here 422)_ sound much better than _this is an exotic status codes; have you considered others?_

Sure this is worth a discussion, so I flagged this PR accordingly.

tfrauenstein:
Some confusion comes from Tronje's too "restrictive" statement in the PR -- it is more precisely addressed by the concrete guideline change. 

I think, we all agree that we *should not subset* the standard HTTP status codes. 
But that was neither the intention nor matches my reading of the change.

We wanted to *recommend* using the well understood, non exotic HTTP codes listed -- that's the reason for now using *should*. My reading is: API designers may use the non listed standard HTTP codes in more exceptional situations, if there are good reasons for it.  We also say that in case of non listed codes being used, you must clearly describe it for the specific API endpoint. Anyway, client must be prepared for receiving non listed standard HTTP status codes.

@dehora Do you still feel that this is still too restrictive? Any specific change proposals (e.g.  use "recommend" instead of "should" to avoid misunderstanding of subsetting)? 
Otherwise let's approve the PR. Thx!

tfrauenstein:
👍

ePaul:
Hmm, this seems to have the goal of "prefer to not use status codes other than the ones listed here", though it actually says "you only need to document status codes which are not listed here".

So Zally now should warn if a status code which is listed is still documented. Is this actually what we want here?

tfrauenstein:
Currently "documentation of listed status codes" is not checked by API Linter. And I would vote for not checking it in future, because we would get many should violations for a topic that is less important.

ePaul:
:+1:

PR #413, opened 2018-06-27
Relax functional naming requirement and improve wording of guidance

tkrop:
We suggest to relax the requirement for functional naming a bit, to comprise the resistance on internal adaptation.

### DIFF
--- a/chapters/naming.adoc
+++ b/chapters/naming.adoc
@@ -20,8 +20,8 @@ follows:
 |=========================================================
 | *Functional Naming* | *Audience*
 | *must*   | external-public, external-partner
-| *should* | company-internal
-| *may*    | component-internal, business-unit-internal
+| *should* | company-internal, business-unit-internal
+| *may*    | component-internal
 |=========================================================
 
 To conduct the functional naming schema, a unique `functional-name` is assigned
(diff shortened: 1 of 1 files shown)


chapters/naming.adoc
```diff
@@ -17,11 +17,12 @@ hostnames>>, <<215, permission names>>, and <<213, event names>> in APIs as
...
-|================================================================
+|=========================================================
 | *Functional Naming* | *Audience*
-| *must*   | external-public, external-partner, company-internal
-| *should* | component-internal, business-unit-internal
-|================================================================
+| *must*   | external-public, external-partner
+| *should* | company-internal
+| *may*    | component-internal, business-unit-internal
+|=========================================================
```
tfrauenstein: I would move business-unit-internal to 'should'

dehora:
> to make account for resistance on internal adaptation.

I'm probably not following here, but can you give some context for this and why the guidelines need to be adjusted as a result? 

I'm tagging https://github.com/zalando/restful-api-guidelines/issues/394 on this as it seems relevant on whether the guidelines should depend on internal documents.

tfrauenstein:
👍

tkrop:
Sorry for the wrong wording.

But we see the aspect that functional naming is less relevant for *component-internal* APIs, as they are normally access only used by consumers within the application. Thus consumers and producers are normally deployed close together making changes here easy to coordinate. For *company-internal* APIs we also see relaxed requirement for functional naming, as it is still comparable simple to migrate - if necessary.

We reconsidered our first classification, because the value of our stability feature is not considered to be high enough at the moment by the delivery teams. We will reconsider our approach as soon as we have more experiences how this feature works out.

PR #421, opened 2018-07-03
Added sentence about disallowing maps from being top level objects

whiskeysierra:
The possibility to return a map (which is non-extensible by design) as the top-level object was likely overseen before. We don't want this, this sentence makes is clear.


ePaul:
:+1:

dehora:
:-1: I'd like to understand the actual problem better. This removes a class of end to end extensibility that's problematic - a good example is the inability to extend Nakadi categories.

dehora:
:+1:

PR #422, opened 2018-07-03
 Event types include Audience definition

tfrauenstein:
Like for REST APIs, Audience should be also defined in Event APIs. Now supported by Nakadi!

### DIFF
--- a/chapters/events.adoc
+++ b/chapters/events.adoc
@@ -176,16 +176,16 @@ EventType:
       example: |
         transactions.order.order-cancelled
         customer.personal-data.email-changed
-  audience:
-        type: string
-        x-extensible-enum:
-          - component-internal
-          - business-unit-internal
-          - company-internal
-          - external-partner
-          - external-public
-        description: |
-          Intended target audience of the event type, analogue to <<219,REST API audience definition>>.
+    audience:
+      type: string
+      x-extensible-enum:
+        - component-internal
+        - business-unit-internal
+        - company-internal
+        - external-partner
+        - external-public
+      description: |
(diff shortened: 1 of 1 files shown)


whiskeysierra:
:+1:

chapters/events.adoc
```diff
@@ -175,6 +176,16 @@ EventType:
       example: |
         transactions.order.order-cancelled
         customer.personal-data.email-changed
+  audience:
```
maxim-tschumak: the indentation is wrong

tfrauenstein: done, thx. 
pls. approve.

chapters/events.adoc
```diff
@@ -175,6 +176,16 @@ EventType:
...
+    audience:
+      type: string
+      x-extensible-enum:
+        - component-internal
+        - business-unit-internal
+        - company-internal
+        - external-partner
+        - external-public
+      description: |
+        Intended target audience of the event type, analogue to <<219,REST API audience definition>>.
```
ePaul: I don't think this link works inside a source code block.
Instead it could be a markdown link (as the description in OpenAPI is Markdown), but even this would not be expanded in the guidelines.

tkrop: Yes, it will not work. Writing pure HTML may fix this.

maxim-tschumak: I tested two options to embed pure HTML - it didn't work. BTW, we should avoid embedding raw HTML since we also produce ePub and PDF.

maxim-tschumak:
:+1:

tkrop:
:+1:

PR #424, opened 2018-07-04
Mentioned X-HTTP-Method-Override for GET with body

whiskeysierra:


dehora:
:+1:

chapters/http-requests.adoc
```diff
@@ -38,8 +38,10 @@ choice. The request information can either be provided distributed to
 multiple query parameters or a single structured URL encoded string.
 2.  POST with body content: when a GET with URL encoded query parameters
 is not possible, a POST with body content must be used. In this case the
-endpoint must be documented with the hint `GET with    body` to
-transport the GET semantic of this call.
+endpoint must be documented with the hint `GET with body` to
+transport the GET semantic of this call. Clients are also highly
+encouraged to use the `X-HTTP-Method-Override: GET` header to allow
```
tkrop: Can you link to a specifying document?

whiskeysierra:
> I only did a quick research, but this header is not really a standard

True, but it wouldn't be the first non-standard header that we promote. From the top of my head we also have `X-RateLimit-*`.

> I'm not sure whether it is somewhere specified sufficiently

No RFC, obviously. But several big players define it within their specifications and docs:
- Google:
  - https://developers.google.com/gdata/docs/2.0/basics?csw=1#UpdatingEntry
  - https://cloud.google.com/apis/docs/system-parameters
  - https://cloud.google.com/apis/docs/http#long_request_urls
- [Microsoft](https://msdn.microsoft.com/en-us/library/dd541471.aspx), although it uses a different name
- [IBM](https://www.ibm.com/support/knowledgecenter/en/SS6PEW_9.4.0/com.ibm.help.custom.restapis.doc/c_SpecifyingHTTPHeaders.html)

> or even supported consistently by frameworks and tools

- Spring, via HiddenHttpMethodFilter
- gRPC as well, see link from above
- Django
- Play

I came across the need for this header when detecting safe requests still within our client. My goal is to allow [hedged or backup requests](https://github.com/zalando/riptide/tree/master/riptide-backup) and I only want to do them for safe requests.

> readers might feel encouraged to use it in other cases too - which definitely should not be our intention.

We mention it as a side note within another rule and very specifically for the use of *GET with body*. Readers can take everything out of context and misuse. 

> Lastly, our API guidelines with outh2 (and our company guidelines of securing transports with TLS) heavily oppose the application of WebCaches which will make this hint majorly useless and

As said above, my intent is not primarily caching, but detection of safe requests. This can be applied not only to backup requests but also to the more general technique of fan-out or just plain old retries.

> distracting noise, in worst case complicating APIs as well as services and clients that apply it.

It shouldn't be used blindly. If a client cares for it, nobody stops them from using it. Servers should only document and implement this, if it brings value (primarily in the way that it's implemented by the framework for free already).

tkrop:
@whiskeysierra okay, I agree that no harm is done and there may be some value when using frameworks. However, please link to a clear specification of the `X-HTTP-Method-Override` semantic to guide API consumers and producers.

whiskeysierra:
> link to a clear specification

The best external spec that I could find is this:

>Restricted clients, or clients behind restricted clients, which cannot use PUT or DELETE SHOULD translate PUT and DELETE to POST operations with an additional X-HTTP-Method-Override: header:
>
>```
>POST /... HTTP/1.1
>...
>X-HTTP-Method-Override: PUT
>```
>Servers SHOULD respond to POST+X-HTTP-Method-Override as if the content of the header were the actual operation sent.
>
> https://opensocial.github.io/spec/2.5.1/Core-API-Server.xml#rfc.section.2.1.1.1

Would that be acceptable?

tkrop:
@whiskeysierra I have found https://tools.ietf.org/html/draft-ietf-scim-api-06 through https://trac.ietf.org/trac/scim/ticket/65.

whiskeysierra:
> The header is primarily there to overcome firewall restrictions on certain HTTP verbs and that services honoring that header should basically "forward" an incoming request to the corresponding controller for the mentioned verb. This obviously would not work for GET over POST with body, as a GET controller doesn't know how to handle a request body.
> 
> Rereading Google's and IBM's definition, it looks like that they also have the GET over POST use case in mind; but then I believe it would be important to enforcing that the body MUST use content type application/x-www-form-urlencoded so that a generic "forwarding handler" can convert the body into a long request string parameter before passing it to the GET controller.

That's not what I had in mind, but after reading the *specs* again, I believe you're right. Limiting GET with body to that media type is not what we want, really. There are two problems that we're trying to solve:
1. URL length restriction
2. Complex, structured payload for GET

The way this header is supposed to be used in conjunction with GET actually only solves the first, but not the second problem.

I had interpreted the header as something that describes the request in a more technical, correct way and that a server would still expect a POST with a body.

> Therefore, I suggest to close this PR.

I'm inclined to agree.

tkrop:
> Originally my understanding was that X-HTTP-Method-Override is what Microsoft calls "verb tunneling": The header is primarily there to overcome firewall restrictions on certain HTTP verbs and that services honoring that header should basically "forward" an incoming request to the corresponding controller for the mentioned verb.

This is a MS typical solution: breaking security restriction only until the security owner finds out :).

@whiskeysierra I agree, so lets close this.

Issue #407, opened 2018-06-22
Version number usage breaks GET-what-you-PUT guideline


ddufourboivin-zalando:
In [this section](https://opensource.zalando.com/restful-api-guidelines/#_version_numbers).

```
< GET /orders

> HTTP/1.1 200 OK
> [
>   { code: "O0000042", version: 1, ... },
>   { code: "O0000043", version: 42, ... }
> ]

< PUT /orders/O0000042
< { code: "O0000042", version: 1, ... }
```

I submit to you that this is wrong.

The `PUT` is then not idempotent, since the next `GET` will return the resource with a `version: 2` instead of `version: 1`. The `PUT` operation is then no longer idempotent.

To respect that rule, the `PUT` should be made with the `version` property already incremented. The server then checks if that `version` is the same or lower than the actual server-side value and decides on when there is a conflict or not.

dehora:
> I submit to you that this is wrong. 

PUT can't be wrong at the uniform interface level because it means "replace" and the request is asking the resource owner to replace the resource state at that URI with the request entity in the PUT. What is expected to happen, assuming no other agent is also manipulating the resource state, is that a subsequent GET returns the replacement `{ code: "O0000042", version: 1, ... }`

> The PUT is then not idempotent, since the next GET will return the resource with a version: 2 instead of version: 1. The PUT operation is then no longer idempotent.

So, being wrong here depends on what the example is trying to achieve and I'd suggest the guild think through the intent. At the interface level what's in the guidelines is ok, but probably not what's intended - PUT doesn't care what the meaning of `version` is.

PUT is imo hugely overrated for application semantics, people seem to endlessly confuse replace with create and update, and this is a great example of that, but that's another discussion.

whiskeysierra:
> The PUT is then not idempotent, since the next GET will return the resource with a version: 2 instead of version: 1. The PUT operation is then no longer idempotent.

> If that's important to the resource owner, then perhaps the example should return a conflict, but strictly speaking, this isn't an idempotence problem.

I believe it's rather a violation of this:

> on successful PUT requests, the server will replace the entire resource addressed by the URL with the representation passed in the payload (subsequent reads will deliver the same payload)
>
> https://opensource.zalando.com/restful-api-guidelines/#put

> To respect that rule, the PUT should be made with the version property already incremented.

That would indeed *fix* it with the downside that the server shifted the burden of version increments to the clients. I wouldn't do that, but it basically is just another cons and should probably be highlighted:

[...]

PR #426, opened 2018-07-26
More precisions on pros/cons of version numbers

ddufourboivin-zalando:
Applying Guild's conclusions, documented [here](https://github.com/zalando/restful-api-guidelines/issues/407#issuecomment-402143390).


chapters/best-practices.adoc
```diff
@@ -59,6 +59,13 @@ or if there was an update since the GET and the entity's etag has changed:
 The ETag for every entity is returned as an additional property of that entity.
 In a response containing multiple entities, every entity will then have a distinct ETag that can be used in subsequent PUT requests.
 
+The content of every entity object in such a response should be kept to a minimum (ID and ETag) so
+those object are not confused for full image of the object (which can be access through a ID specific
+query).
```
ePaul: But if we need to fetch the object separately anyway to use PUT later, then this is not better than the previous version at all. I think the idea of this version was to save those separate requests.

chapters/best-practices.adoc
```diff
@@ -89,10 +96,16 @@ or if there was an update since the GET and the entity's etag has changed:
...
-=== Version numbers
+=== Version numbers -- In the payload
 The entities contain a property with a version number. 
 When a PUT request is made, the version number has to be included in the payload and the server will check that the version in the database is not higher.
 
+In this solution, the `etag` property is not `readonly` since it is provided at PUT time as part of
+the payload.
+
+The verb for updating here could be POST instead of PUT, since the server has to compute the received
+value before saving it (incrementing the version number).
```
ePaul: A POST has a different semantics, it would mean to create a subresource under the resource in the URL.

chapters/best-practices.adoc
ddufourboivin-zalando: OK. Agreed. I will remove that extra explanation and put some properties in the `GET` example to illustrate that.

chapters/best-practices.adoc
ddufourboivin-zalando: Mmm that is what we all agreed on 3 weeks ago during the meeting. We could re-discuss it in person next Tuesday at the guild's meeting.

chapters/best-practices.adoc
ddufourboivin-zalando: Actually, just noticed that my example HAD all the properties. I forgot to remove the paragraph before committing.

ddufourboivin-zalando: @ePaul : Please see [this commit](https://github.com/zalando/restful-api-guidelines/pull/426/commits/0c94d1b0d3d48f9644356dc803f74394e27708b9)

chapters/best-practices.adoc
```diff
@@ -101,15 +110,14 @@ Example:
 > HTTP/1.1 200 OK
 > {
 >   "items": [
->     { code: "O0000042", version: 1, ... },
->     { code: "O0000043", version: 42, ... }
+>     { "code": "O0000042", "version": 1,  "foo": 42, "bar": true },
```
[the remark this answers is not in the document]
ddufourboivin-zalando: Totally.

chapters/best-practices.adoc
```diff
@@ -89,10 +92,16 @@ or if there was an update since the GET and the entity's etag has changed:
 ==== Cons
 * Information that only belongs in the HTTP header is part of the business objects
 
-=== Version numbers
+=== Version numbers -- In the payload
 The entities contain a property with a version number. 
 When a PUT request is made, the version number has to be included in the payload and the server will check that the version in the database is not higher.
 
+In this solution, the `etag` property is not `readonly` since it is provided at PUT time as part of
```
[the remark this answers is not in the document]
ddufourboivin-zalando: It's a copy/paste mistake, since it is here `id` and not `etag`. Thanks for this.

tkrop:
:+1:

chapters/best-practices.adoc
```diff
@@ -93,6 +96,12 @@ or if there was an update since the GET and the entity's etag has changed:
 The entities contain a property with a version number. 
 When a PUT request is made, the version number has to be included in the payload and the server will check that the version in the database is not higher.
 
+In this solution, the `version` property is not `readonly` since it is provided at PUT time as part of
+the payload.
+
+The verb for updating here could be POST instead of PUT, since the server has to compute the received
```
ddufourboivin-zalando: From me to me:

Indicate clearly that we mean

```
POST /foos/123
```
and not
```
POST /foos
```
in this case.

ddufourboivin-zalando: # Update

@tkrop @maxim-tschumak @ePaul and others:

I changed the wording and the verb in the example in [this commit](https://github.com/zalando/restful-api-guidelines/pull/426/commits/294494d37b613395be605507f7c987516892d12c).

Let me know what you think.

ddufourboivin-zalando: # Next Question

@tkrop , I am a bit confused since we removed the "duplicated" _Version_ example.

I think we should modify the example here. Right now, I am pushing the version number in the `POST` payload, as part of the object. I think I heard you say that is should be out of the payload and instead in the `If-Match` header. Would that be better?

chapters/best-practices.adoc
```diff
@@ -122,6 +132,7 @@ or if there was an update since the GET and the version number in the database i
 
 ==== Cons
 * Functionality that belongs into the HTTP header becomes part of the business object
+* Using POST instead of PUT for an "update" logic (not a problem in itself, but may feel unusual for the consumer)
```
tkrop: `"update"` -> `update` why do feel the need to quote here? It is an update, so what?

ddufourboivin-zalando: Fixed.

tkrop:
:+1:

Issue #428, opened 2018-07-31
Implicit filtering MUST be documented


ddufourboivin-zalando:
If any implicit filtering is performed on a list of resource or on the fields of an object, it MUST be transparently and clearly documented in the API Specification.

Example:
`GET /business-partners` must, for legal reason, list only the business partner on which you have access (let's say you are from _Foo_, you have access to _Foo Performance_ and _Foo Sport_ and _Foo Signature_; but you should never ever see in the result of that `GET` request that we are doing business also with _Bar_).

PR #429, opened 2018-07-31
Document any implicit filtering

ddufourboivin-zalando:
Add new rule 226 "Document any implicit filtering" (#428).

### DIFF
--- a/chapters/http-requests.adoc
+++ b/chapters/http-requests.adoc
@@ -255,18 +255,18 @@ the escaping of special characters and the maximal URL length.
 [#226]
 == {MUST} Document Implicit Filtering
 
-Implicit filtering should be avoided as much as possible. However, in some
-cases (usually for legal reasons), some endpoints should not be listing all
-resources they have, but only those on which the current consumer has access
-to (or some other context specific condition).
+Sometimes certain collection resources or queries will not list all the 
+possible elements they have, but only those for which the current client 
+is authorized to access.
 
 Implicit filtering could be done on:
 
 * the list of resources being return on a parent `GET` request
 * the fields returned for the resource's detail
 
-In such cases, the implicit filtering must absolutely be documented clearly
-and transparently in the API Specification (in its description).
+In such cases, the implicit filtering must be in the API Specification (in its description).
+
+Consider <<227,caching considerations>> when implicitely filtering.
 
 Example:
 
@@ -304,7 +304,7 @@ The API Specification should then specify something like this:
 
 [source,yaml]
 ----
-  /busines-partner:
+  /business-partner:
     get:
       description: >-
         Get the list of registered business partner.
--- a/chapters/http-requests.adoc
+++ b/chapters/http-requests.adoc
@@ -252,3 +252,67 @@ parameters as described below.
 When choosing the collection format, take into account the tool support,
 the escaping of special characters and the maximal URL length.
 
+[#226]
+== {MUST} Document any implicit filtering
+
+Implicit filtering should be avoided as much as possible. However, in some
+cases (usually for legal reasons), some endpoints should not be listing all
+resources they have, but only those on which the current user has access to
+(or some other context specific condition).
+
+Implicit filtering could be done on:
+
+* the list of resources being return on a parent `GET` request
+* the fields returned for the resource's detail
+
+In such cases, the implicit filtering must absolutely be documented clearly
+and transparently in the API Specification (in its description).
+
+Example:
+
+If an employee of the company _Foo_ accesses one of our business-to-business
+service and performs a `GET /business-partners`, it must, for legal reasons, not display
+any other business partner that is not owned or contractually managed by her/his company.
+It should never see that we are doing business also with company _Bar_.
+
+[source]
+----
+# An employee from `FOO` queries.
+
+< GET /business-partners
+< Authorization: Bearer A-TOKEN-FROM-A-FOO-EMPLOYEE
+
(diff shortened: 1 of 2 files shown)


chapters/http-requests.adoc
```diff
@@ -252,3 +252,67 @@ parameters as described below.
 When choosing the collection format, take into account the tool support,
 the escaping of special characters and the maximal URL length.
 
+[#226]
+== {MUST} Document any implicit filtering
```
tkrop: As far as I know we prefer titles with capital letters.

I would remove `any`.

chapters/http-requests.adoc
```diff
@@ -252,3 +252,67 @@ parameters as described below.
...
+
+Implicit filtering should be avoided as much as possible. However, in some
+cases (usually for legal reasons), some endpoints should not be listing all
+resources they have, but only those on which the current user has access to
```
tkrop: `user` -> `consumer`

chapters/http-requests.adoc
```diff
@@ -252,3 +252,67 @@ parameters as described below.
...
+(or some other context specific condition).
```
tkrop: Are there other conditions?

chapters/http-requests.adoc
```diff
@@ -252,3 +252,67 @@ parameters as described below.
...
+and transparently in the API Specification (in its description).
+
+Example:
+
+If an employee of the company _Foo_ accesses one of our business-to-business
+service and performs a `GET /business-partners`, it must, for legal reasons, not display
+any other business partner that is not owned or contractually managed by her/his company.
+It should never see that we are doing business also with company _Bar_.
+
+[source]
```
tkrop: In my humble opinion the before sentence is example enough. Anyhow, I would favor a shorter examples than full blown `curl` output.

chapters/http-requests.adoc
ddufourboivin-zalando: There could be, so I would not limit it exclusively to this (legal reasons).

chapters/http-requests.adoc
```diff
@@ -252,3 +252,61 @@ parameters as described below.
...
+        { "name": "Bar pour Elle" }
+    ]
+}
+----
+
+The API Specification should then specify something like this:
+
+[source,yaml]
+----
+  /busines-partner:
```
dehora: `busines-` -> `business`

chapters/http-requests.adoc
```diff
@@ -252,3 +252,61 @@ parameters as described below.
 When choosing the collection format, take into account the tool support,
 the escaping of special characters and the maximal URL length.
 
+[#226]
+== {MUST} Document Implicit Filtering
+
+Implicit filtering should be avoided as much as possible. However, in some
+cases (usually for legal reasons), some endpoints should not be listing all
+resources they have, but only those on which the current consumer has access
+to (or some other context specific condition).
```
dehora:
```patch
+ Implicit filtering should be avoided as much as possible. However, in some
+ cases (usually for legal reasons), some endpoints should not be listing all
```

I don't think the should qualifier is needed. This implicit filtering is standard behaviour for product APIs that support multiple tenants, especially SaaS products (eg Stripe/Intercom/GitHub/Twilio all do this) that document a single set of resources, bearing in mind that API as a product is a top level ambition. The leading sentence could be

```
Sometimes certain collection resources or queries will not list all the 
possible elements they have, but only those for which the current client 
is authorized to access.
```

chapters/http-requests.adoc
```diff
@@ -252,3 +252,61 @@ parameters as described below.
...
+
+Implicit filtering could be done on:
+
+* the list of resources being return on a parent `GET` request
+* the fields returned for the resource's detail
+
+In such cases, the implicit filtering must absolutely be documented clearly
+and transparently in the API Specification (in its description).
```
dehora: I would expect some guidance on caching in these cases. It's generally not safe for middleboxes and proxies to blindly cache results against a common resource URI without something like a vary header on the conditional element of the request (eg the Authorization or Cookie header). That gets into more sophisticated and more complicated caching techniques which might be best avoided.

whiskeysierra: > If an API is intended to support caching, it must take care to specify this ability by defining the caching boundaries i.e. time-to-live and cache constraints, by providing the Cache-Control and Vary headers
> [..]
> Caching has to take many aspects into account, e.g. general cacheability of response information, our guideline to protect endpoints using SSL and OAuth authorization
>
> https://opensource.zalando.com/restful-api-guidelines/#_must_document_caching_if_supported

I'd just link there. The paragraph explains it already pretty well, imho.

dehora: > I'd just link there. 

:+1:

tkrop:
:+1:

chapters/http-requests.adoc
ddufourboivin-zalando: All for it. But I just noticed that [that rule](https://opensource.zalando.com/restful-api-guidelines/#_must_document_caching_if_supported) does not have a number. Is that normal?

ddufourboivin-zalando: OK found the problem. There is no `#`. I am fixing this in this PR.

```adoc
[217]
== {MUST} Document Caching, if Supported
```

should be

```adoc
[#217]
== {MUST} Document Caching, if Supported
```

dehora:
:+1:

tkrop:
:+1:

PR #431, opened 2018-08-10
Emphazise specification via default error handling

tkrop:
In API reviews I often can find error specifications that just repeat the standard and provide no value for consumers. So I suggest to more emphasize the paragraph that suggest to remove this error specifications.


ddufourboivin-zalando:
👍

whiskeysierra:
:+1:

dehora:
:+1:

PR #432, opened 2018-08-10
Prevent incorrect rule id anchors

tkrop:
To prevent repeating failures when assigning rule id anchors as last experienced in #429.


ddufourboivin-zalando:
👍

ddufourboivin-zalando:
@tkrop : I fixed the reason for the failing _Travis_ builds. If you agree with my modification, give your :+1: and let's merge.

tkrop:
Thanks, I had no time to fix this.

tkrop:
:+1:

PR #433, opened 2018-08-15
feat: document retry-after with 503 (#377)

### LINKED ISSUE #377: Document `Retry-After` with 503
## I suggest

  - a rule to  explicit the usage of `Retry-After` header when 503

## Feedback welcome ;)

tkrop:
Fixes #377.

### DIFF
--- a/chapters/http-status-codes-and-errors.adoc
+++ b/chapters/http-status-codes-and-errors.adoc
@@ -168,10 +168,11 @@ unexpected server execution problem (here, client retry may be sensible)
 implies future availability, e.g. new feature). |All
 
 |503 |Service Unavailable - service is (temporarily) not available (e.g.
-due to overload) -- client retry may be sensible. If possible, the
-service should indicate by setting the
-https://tools.ietf.org/html/rfc7231#section-7.1.3['Retry-After'] how
-long the client should wait befor retry. |All
+if a required component or downstream service is not available) -- client
+retry may be sensible. If possible, the service should indicate how long
+the client should wait by setting the
+https://tools.ietf.org/html/rfc7231#section-7.1.3['Retry-After']
+header. |All
 |=======================================================================
 
 [#220]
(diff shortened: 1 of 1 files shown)


maxim-tschumak:
:+1:

chapters/http-status-codes-and-errors.adoc
```diff
@@ -167,8 +167,11 @@ unexpected server execution problem (here, client retry may be sensible)
 |501 |Not Implemented - server cannot fulfill the request (usually
 implies future availability, e.g. new feature). |All
 
-|503 |Service Unavailable - server is (temporarily) not available (e.g.
-due to overload) -- client retry may be sensible. |All
+|503 |Service Unavailable - service is (temporarily) not available (e.g.
+due to overload) -- client retry may be sensible. If possible, the
```
whiskeysierra: How does this relate to `429 Too Many Requests`?

ddufourboivin-zalando:
👍

chapters/http-status-codes-and-errors.adoc
tkrop: On first sight I would say there is none - at least with respect to issue #377.

Do you suggest to add the same advice for `429`?

Interesting idea to simplify client design. The risk is, that the server might easily create improper advice by always looking up the next expected free slot which may lead to clients requesting to aggressively. Anyhow, lets open a new issue for that.

whiskeysierra: No, but the example *due to overload* sounds more like 429 to me. I'd expect 503 when one of critical dependencies is down, e.g. the database with a scheduled maintenance window.

whiskeysierra: Overload due to client requests is a client error, hence 4xx.
Being unable to serve requests due to downstream dependencies is a server error, hence 5xx.

tkrop: Good point. I will change this.

maxim-tschumak:
:+1:

ddufourboivin-zalando:
👍

Issue #434, opened 2018-08-15
Clarify guidance on process recource vs. data resource update.


tkrop:
In the context of #391 we had stumbled about a very specific problem how to best **update** *resources* which leads to multiple discussions (see https://github.com/zalando/restful-api-guidelines/issues/391#issuecomment-403156420 and https://github.bus.zalan.do/Sokoban/rutherfordium/pull/17#discussion_r551990) including the API guild meeting.

The core discussion can be summarized by `POST` vs. `PUT` vs. `PATCH` and the idea of distinguishing between *process resources* and *data resources*.

tkrop:
Hi @whiskeysierra @meshcalero, I'm working on my concept to guide by distinguishing on resources. Please give feedback.

The main idea is as follows:

**Data resources** are resources or properties of resources that represent knowledge or facts, e.g. master data, product data, article reviews, e-mail messages, process documentation. Updates to these resources can happen on external requests and does only change the resource data. It does never trigger subsequent actions or processes except for propagating the changed information to listeners registered to this data resource. Data changes may be rejected to ensure the format or consistency of the changed resource.

**Process resources** are resources or properties of resources that are involved into (business) processes by reflecting the process state or by providing process control information. Updates to these resources can be external and happen either on the process state to reflect the advancement of the (business) process or on control information to reflect changes in the business process. Changes to the process state or control information must be validated to ensure the correctness of the conducted (business) process. Any change is usually triggering subsequent actions or processes to advance the (business) process further until its final state is reached.

Base on this definition I would see the following guidance for `POST` vs. `PUT` vs. `PATCH`:

Updates to **data resources** should follow the first three suggestions in the [`PATCH`-guideline](https://opensource.zalando.com/restful-api-guidelines/#patch):

1. `PUT` only for very simple, small resources (were only all properties together cover the resource purpose). This follows @whiskeysierra best practices that consumers should only be forced to implement against properties they do not need to know for their purpose.
2. `PATCH` with JSON Merge Patch as before.
3. `PATCH` with JSON Patch as before.


Updates to **process resources** (in my opinion) should follow only the last suggestion in the [`PATCH`-guideline](https://opensource.zalando.com/restful-api-guidelines/#patch):

1. `POST` with a proper description of what is happening.

[...]

PR #435, opened 2018-08-15
feat: improve guidance on PUT vs. POST as well other verbs.

### LINKED ISSUE #391: Better guidance for PUT vs. POST for creation
We currently don't give a clear guidance on the usage of HTTP methods for resource creation.
Currently our APIs have a wild mixture of both.

A naive POST has a problem in case of a connection issue that the client doesn't know whether it was successful, and retry might give a duplicated resource.
For a PUT you generally need to know the resource's ID before creation.

How can we make this clearer?

Some ideas:

* ID generation on client side, if possible (maybe UUID, but see [rule 144](https://opensource.zalando.com/restful-api-guidelines/#144)), then use PUT
* POST for ID generation, then PUT
* idempotent POST (using some foreign keys)
* POST with a way to find out whether it worked (GET)

tkrop:
Fixes #391.

I have revisited #391 and excluded the discussion we continued about `POST` vs. `PUT` vs. `PATCH` for update in the API Guild this week (see https://github.com/zalando/restful-api-guidelines/issues/391#issuecomment-403156420 and https://github.bus.zalan.do/Sokoban/rutherfordium/pull/17#discussion_r551990). So the changes are majorly editorial, removing some inconsistencies and improving the guidance on aspects we all commonly agreed about.

Unfortunately, the enhancements are a bit hidden between the reformatting. So please read carefully and make suggestions to improve wording. I have create a new [issue](#434) to track the remaining discussion.

### DIFF
--- a/chapters/http-requests.adoc
+++ b/chapters/http-requests.adoc
@@ -11,13 +11,12 @@ follows:
 [[get]]
 === GET
 
-GET requests are used to *read* either a single resource or a set of resources
-on a list endpoint.
+GET requests are used to *read* either a single resource or a set of resources.
 
 * GET requests for individual resources will usually generate a 404 if the
 resource does not exist
-* GET requests for collection resources may return either 200 (if the listing
-is empty) or 404 (if the list is missing)
+* GET requests for collection resources may return either 200 (if the
+collection is empty) or 404 (if the collection is missing)
 * GET requests must NOT have a request body payload
 
 *Note:* GET requests on collection resources should provide sufficient
@@ -96,8 +95,6 @@ POST requests are idiomatically used to *create* single resources on a
 collection resource endpoint, but other semantics on single resources endpoint
 are equally possible. The semantic for collection endpoints is best described
 as _"please add the enclosed representation to the collection resource
-identified by the URL"_. The semantic for single resource endpoints is best
-described as _"please execute the given well specified request on the resource
 identified by the URL"_.
 
 * on a successful POST request, the server will create one or multiple new
@@ -106,9 +103,13 @@ resources and provide their URI/URLs in the response
 updated), 201 (if resources have been created), and 202 (if the request was
 accepted but has not been finished yet)
 
-*More generally:* POST should be used for scenarios that cannot be covered by
-the other methods sufficiently. In such cases, make sure to document the fact
-that POST is used as a workaround (see <<get-with-body>>).
+The semantic for single resource endpoints is best described as _"please
+execute the given well specified request on the resource identified by the
+URL"_.
+
+*Generally:* POST should be used for scenarios that cannot be covered by the
+other methods sufficiently. In such cases, make sure to document the fact that
+POST is used as a workaround (see <<get-with-body>>). 
 
 *Note:* Resource IDs with respect to POST requests are created and maintained
 by server and returned with response payload.
@@ -121,6 +122,7 @@ the server. Here, a _foreign key_ is a natural partner resource ID, e.g. the
 shopping cart ID is a _foreign key_ candidate for an order, while an
 _idempotency key_ is an artificial request key that is preserved when
 re-sending the POST request.
+
 
 [[patch]]
 === PATCH
@@ -200,7 +202,7 @@ and resource collections.
 body.
 
 *Hint:* This is particular useful to efficiently lookup whether large resources
-or list resources have been updated in conjunction with the
+or collection resources have been updated in conjunction with the
 https://tools.ietf.org/html/rfc7232#section-2.3[`ETag`-header].
 
 [[options]]
@@ -277,7 +279,7 @@ is authorized to access.
 
 Implicit filtering could be done on:
 
-* the list of resources being return on a parent `GET` request
+* the collection of resources being return on a parent `GET` request
 * the fields returned for the resource's detail
 
 In such cases, the implicit filtering must be in the API Specification (in its description).
--- a/chapters/http-requests.adoc
+++ b/chapters/http-requests.adoc
@@ -7,198 +7,214 @@
 Be compliant with the standardized HTTP method semantics summarized as
 follows:
 
+
 [[get]]
 === GET
 
-GET requests are used to read a single resource or query set of
-resources.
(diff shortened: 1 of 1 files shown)


chapters/http-requests.adoc
```diff
@@ -7,198 +7,214 @@
...
+of <<182,`ETag` and `If-(None-)Match`>> headers should be considered to signal
+the server stricter demands to expose conflicts and prevent lost updates.
+The section <<optimistic-locking>> also describes some alternatives to this
+approach. 
+
+*Note:* In the rare cases where PUT is although used for resource creation,
+the resource IDs are maintained by the client and passed as a URL path segment.
+Putting the same resource twice is required to be idempotent and to result in
+the same single resource instance. If PUT is applied for creating a resource,
+only URIs should be allowed as resource IDs. If URIs are not available POST
```
whiskeysierra: >  If PUT is applied for creating a resource, only URIs should be allowed as resource ID

I believe that will cause confusion. My first thought was this `/web-sites/http://example.org` (ignoring escaping). Have we specified the term resource ID somewhere? Most people refer to the last path segment not the whole URI.

chapters/http-requests.adoc
```diff
@@ -7,198 +7,214 @@
...
 follows:
 
+
 [[get]]
 === GET
 
-GET requests are used to read a single resource or query set of
-resources.
+GET requests are used to *read* either a single resource or a set of resources
+on a list endpoint.
```
meshcalero: not sure I like the distinction between a single resource or a list endpoint. For me a resource represents a specific concept. Whether this concept is a single entity or a concept of entities doesn't matter. I would use the term "set of resources" only for things like Multi-GET or embedded resources.

chapters/http-requests.adoc
```diff
@@ -7,198 +7,214 @@
...
-same resource twice is by itself *not* required to be idempotent and may
-result in multiple resource instances. Anyhow, if external URIs are
-present that can be used to identify duplicate requests, it is best
-practice to implement POST in an idempotent way.
+POST requests are idiomatically used to *create* single resources on a
+collection resource endpoint, but other semantics on single resources endpoint
+are equally possible. The semantic for collection endpoints is best described
+as _"please add the enclosed representation to the collection resource
+identified by the URL"_. The semantic for single resource endpoints is best
+described as _"please execute the given well specified request on the resource
```
meshcalero: > The semantic for single resource endpoints is best described as _"please execute the given well specified request on the resource
I would move this sentence below the two the list below, as what is stated in the list only applies to creates. For single resource POST we can not make any assumptions what it does and returns.

chapters/http-requests.adoc
tkrop: Okay, I change it back.

tkrop: I wanted to say, I remove the term *list endpoint*.

> I would use the term "set of resources" only for things like Multi-GET or embedded resources.

Can you give me a hint, why *set of resources* should be not be used here to address the use case you mentioned here?

chapters/http-requests.adoc
```diff
@@ -7,198 +7,214 @@
...
+* GET requests for collection resources may return either 200 (if the listing
+is empty) or 404 (if the list is missing)
+* GET requests must NOT have a request body payload
 
-* GET requests for individual resources will usually generate a 404 if
-the resource does not exist
-* GET requests for collection resources may return either 200 (if the
-listing is empty) or 404 (if the list is missing)
-* GET requests must NOT have request body payload
+*Note:* GET requests on collection resources should provide sufficient
```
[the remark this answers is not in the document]
tkrop: `list` -> 27 (8 in `http-requests.adoc`)
`collection` -> 33 (16 in `http-requests.adoc`)

Any other opinions about this?

chapters/http-requests.adoc
tkrop: Good catch. I removed this at another location, but oversaw this usage.

chapters/http-requests.adoc
tkrop: Okay, I put it to the next paragraph below the list.

chapters/http-requests.adoc
```diff
@@ -7,198 +7,214 @@
...
+by server and returned with response payload.
+
+*Tip:* Posting the same resource twice is by itself *not* required to be
+_idempotent_ and may result in multiple resource instances. However, it is
+best practice to implement POST _idempotent_, by temporary or permanently
+storing _foreign keys_ or _idempotency keys_ to identify re-send requests on
+the server. Here, a _foreign key_ is a natural partner resource ID, e.g. the
+shopping cart ID is a _foreign key_ candidate for an order, while an
+_idempotency key_ is an artificial request key that is preserved when
+re-sending the POST request.
```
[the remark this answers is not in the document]
tkrop: Good point. I tried this by the example, but may forgot an abstract definition.

chapters/http-requests.adoc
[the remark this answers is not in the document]
tkrop: The application of this request is easier than thought. I had only to switch 4 appearances from `list` to `collection`, which I think is more suitable.

ddufourboivin-zalando:
👍

meshcalero:
To me the first sentence on GET is still not fully appropriate

> GET requests are used to read either a single resource or a set resources

From my point of view a GET always reads a single resource. A resource represents some concept, which could be a single entity or a collection (usually of other resources).

I would just use the sentence "GET requests are used to read a resource", but if others prefer the current text, I wouldn't veto either.

zeitlinger:
I'd write:

GET requests are used to read either a single or a collection resource.

.. because the following paragraphs refer to collection resources, not sets of resources.

tkrop:
@meshcalero I took over @zeitlinger  suggestions. Do you approve?

zeitlinger:
👍

tkrop:
:+1:

meshcalero:
:+1:

tkrop:
:+1:

zeitlinger:
:+1:

PR #440, opened 2018-08-21
feat: extend the colection format rule (154) to support OpenAPI 3 (#439)

### LINKED ISSUE #439: Adopt collection format of query parameters rule (154) to support OpenAPI3

maxim-tschumak:
Closes #439

### DIFF
--- a/chapters/http-requests.adoc
+++ b/chapters/http-requests.adoc
@@ -229,12 +229,12 @@ for further hints on how to support the different HTTP methods on
 resources.
 
 [#154]
-== {SHOULD} Define Collection Format of Query Parameters
+== {SHOULD} Define Collection Format of Query Parameters and Headers
 
-Sometimes, query parameters allow to provide a list of values, either by
-providing a comma-separated list (`csv`) or by repeating the parameter
-multiple times with different values (`multi`). The API specification
-should explicitly define one type as follows:
+Sometimes, query parameters and headers allow to provide a list of values,
+either by providing a comma-separated list (`csv`) or by repeating the
+parameter multiple times with different values (`multi`). The API
+specification should explicitly define one type as follows:
 
 [,cols="20%,20%,25%,35%",options="header",]
 |=======================================================================
--- a/chapters/http-requests.adoc
+++ b/chapters/http-requests.adoc
@@ -229,28 +229,18 @@ for further hints on how to support the different HTTP methods on
 resources.
 
 [#154]
-== {SHOULD} Use and Specify Explicitly the Form-Style Query Format for Collection Parameters
+== {SHOULD} Define Collection Format of Query Parameters
 
-There are different ways of supplying a set of values as a query
-parameter. The form-style query format should be used as described in
-https://tools.ietf.org/html/rfc6570#section-3.2.8[RFC6570 standard].
-The set of parameter values should be supplied either in comma-separated
-or multiple parameter instances format. One particular type should be selected
-and stated explicitly in the API specification.
-
-In the OpenAPI 3, the 
-https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#style-values[`style`]
-and `explode` fields of the
-https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#parameter-object[parameter object]
-are used to specify the format of parameter supply.
-In the OpenAPI 2, it is the http://swagger.io/specification/[`collectionFormat`]
-property.
+Sometimes, query parameters allow to provide a list of values, either by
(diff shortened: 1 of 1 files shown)


ddufourboivin-zalando:
👍

chapters/http-requests.adoc
```diff
@@ -229,24 +229,28 @@ for further hints on how to support the different HTTP methods on
 resources.
 
 [#154]
-== {SHOULD} Explicitly define the Collection Format of Query Parameters
+== {SHOULD} Use and Specify Explicitly the Form-Style Query Format for Collection Parameters
 
 There are different ways of supplying a set of values as a query
```
tkrop: To simplify language I suggest the following text:

*Sometimes, query parameters allow to provide a list of values, either by providing a comma-separated list (`csv`) or by repeating the parameter multiple times with different values (`multi`). The API specification should explicitly define one type as follows:*

I think the following two sentences are optional:

*In the Open API 3.0, the https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#style-values[`style: form`] in combination with `explode: true | false` is used.*

*In the Open API 2.0 the https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameter-object[`collectionFormat: csv | multi`] is used.*

chapters/http-requests.adoc
```diff
@@ -229,24 +229,28 @@ for further hints on how to support the different HTTP methods on
...
+property.
+
+[,cols="20%,20%,25%,35%",options="header",]
 |=======================================================================
-|Collection Format |Description |Example
-|`csv` |Comma separated values |`?parameter=value1,value2,value3`
-
-|`multi` |Multiple parameter instances
-|`?parameter=value1&parameter=value2&parameter=value3`
+|Description |OpenAPI 3 |OpenAPI 2 |Example
```
tkrop: May be better

1. `Open API 3.0 | Open API 2.0`
2. The example can use `key` or `param` instead of `parameter`.
3. `Multiple parameter instances` => `Multiple parameters`.

chapters/http-requests.adoc
```diff
@@ -229,24 +229,18 @@ for further hints on how to support the different HTTP methods on
 resources.
 
 [#154]
-== {SHOULD} Explicitly define the Collection Format of Query Parameters
+== {SHOULD} Define Collection Format of Query Parameters
```
tkrop: I'm still not sure whether we should restrict this to query parameters only. Shouldn't path or header parameters follow the same rule if used?

chapters/http-requests.adoc
```diff
@@ -229,24 +229,18 @@ for further hints on how to support the different HTTP methods on
...
 |=======================================================================
-|Collection Format |Description |Example
-|`csv` |Comma separated values |`?parameter=value1,value2,value3`
-
-|`multi` |Multiple parameter instances
-|`?parameter=value1&parameter=value2&parameter=value3`
+|Description |OpenAPI 3.0 |OpenAPI 2.0 |Example
+|Comma separated values |`style: form, explode: false` |`collectionFormat: csv` |`?param=value1,value2`
+|Multiple parameters |`style: form, explode: true` |`collectionFormat: multi` |`?param=value1&param=value2`
 |=======================================================================
```
tkrop: In reviews I have experienced, that at least some readers cannot apply this concept in Open API. So may be we should give them at least one very concrete example:

For Open API 3.0 a `csv` query parameter will look as follows:
```
parameters:
    - name: customer_id
      in: query
      type: array
      style: form
      explode: false
      items:
        type: string
```

chapters/http-requests.adoc
whiskeysierra: Headers maybe, but arrays in path segments sounds incredibly wrong to me.

chapters/http-requests.adoc
whiskeysierra: If it's `explode: false` shouldn't it be `customer_ids` then?

chapters/http-requests.adoc
maxim-tschumak: I also wouldn't do this for path parameters, but it sounds reasonable for headers.

tkrop: @whiskeysierra I agree, but the URI format is open to this, and if we cannot predict use cases here, my approach would to not restrict the application of the rule. The only result would be, that readers heaving the use case would not specify it, making it even harder to detect in a review.

whiskeysierra:
:+1:

maxim-tschumak:
:+1:

tkrop:
:+1:

PR #453, opened 2018-09-25
Add a general guidance on resource identifiers

### LINKED ISSUE #436: Give better guidance on resource identifiers
Spawned from the following comment: https://github.com/zalando/restful-api-guidelines/pull/435/files#r210293544:

The following sentences in https://opensource.zalando.com/restful-api-guidelines/#put 

> If PUT is applied for creating a resource, only URIs should be allowed as resource IDs. If URIs are not available POST should be preferred.

are not clear. It seems the meaning is already implied in the rest of the paragraph, and we can remove these sentences.

tkrop:
Closes #436.

This PR contains a guideline change. It contains a new rule to restrict the characters used for a resource identifier. In general resource identifiers should already be restricted to this set, but this assumption may be false in practice.

I'm although not sure, whether Zally should enforce this rule by checking whether resource-id definitions are always restrained to `[\w:.-]*`.

### DIFF
--- a/chapters/resources.adoc
+++ b/chapters/resources.adoc
@@ -65,6 +65,14 @@ definition. For example, "sales-order-items" is superior to
 "order-items" in that it clearly indicates which business object it
 represents. Along these lines, "items" is too general.
 
+[#228]
+== {MUST} Use URL-friendly Resource Identifiers: [a-zA-Z0-9:._-]*
+
+To simplify encoding of resource IDs in URLs, their representation
+must only consist of ASCII strings of letters, numbers, underscore,
+minus, colon, and period.
+
+
 [#143]
 == {MUST} Identify resources and Sub-Resources via Path Segments
 
--- a/chapters/http-requests.adoc
+++ b/chapters/http-requests.adoc
@@ -83,9 +83,7 @@ approach.
 *Note:* In the rare cases where PUT is although used for resource creation,
 the resource IDs are maintained by the client and passed as a URL path segment.
 Putting the same resource twice is required to be idempotent and to result in
-the same single resource instance. If PUT is applied for creating a resource,
-only URIs should be allowed as resource IDs. If URIs are not available POST
-should be preferred.
+the same single resource instance.
 
 
 [[post]]


chapters/resources.adoc
```diff
@@ -65,6 +65,14 @@ definition. For example, "sales-order-items" is superior to
 "order-items" in that it clearly indicates which business object it
 represents. Along these lines, "items" is too general.
 
+[#228]
+== {MUST} Use URL-friendly Resource Identifiers: [a-zA-Z0-9:._-]*
+
+To simplify encoding of resource IDs in URLs, their representation
+must only consist of ASCII strings of letters, numbers, underscore,
+minus, colon, and period.
```
whiskeysierra: Colon is a reserved character in URLs:

> If data for a URI component would conflict with a reserved
   character's purpose as a delimiter, then the conflicting data must be
   percent-encoded before the URI is formed.
>
>  reserved    = gen-delims / sub-delims
    gen-delims  = ":" / "/" / "?" / "#" / "[" / "]" / "@"
>
>
> https://tools.ietf.org/html/rfc3986#section-2.2

tkrop: I used https://tools.ietf.org/html/rfc3986#section-3.3 to verify this restricted pattern:

>      segment       = *pchar
>      segment-nz    = 1*pchar
>      segment-nz-nc = 1*( unreserved / pct-encoded / sub-delims / "@" )
>                    ; non-zero-length segment without any colon ":"
>
>      pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"

whiskeysierra: Ah, it looks like they have less restrictive rules for path segments. This shouldn't be an issue since we don't use resource ids in the first path segemnt, do we?

> in which case the
   first path segment cannot contain a colon (":") character

tkrop: @whiskeysierra No we don't, but this wouldn't be a problem too, as we only allow [absolute paths](https://opensource.zalando.com/restful-api-guidelines/#217). :smile_cat: Anyhow, I only added `:` to potentially be compliant with our proposal - if some day ... you know what I mean.

whiskeysierra:
:+1:

PR #454, opened 2018-09-25
feat: link pagination with query parameters and improve wording (#445)

### LINKED ISSUE #445: Pagination should link to rule 137 and introduce standard keywords.
I often review APIs that provide their own terminology for pagination. Rule [137](https://opensource.zalando.com/restful-api-guidelines/#137) introduces the standard terminology but is not reflected in the pagination section. Therefore we should link this and use this terminology consistently.

tkrop:
Closes #445.

This change tries to better communicate a consistent look and feel of APIs.

While it replaces a `MAY` by a `MUST`, I think this was the intended semantic for providing conventional query parameters supporting query parameters for searching, sorting, filtering, and paginating. Thus I'm not sure whether this is a guideline change.

I also remove the alternate `size` for `limit`, and `page` for `offset` to enforce a stricter look and feel. If voted for I will reintroduce them again.

### DIFF
--- a/chapters/naming.adoc
+++ b/chapters/naming.adoc
@@ -160,7 +160,6 @@ paginating, you must stick to the following naming conventions:
   `embed` correctly is difficult, so do it with care. See <<158>> below.
 * `offset` — numeric offset of the first element on a page. See <<pagination>>
   section below.
-  paque value, never to be inspected/constructed, addresses a single page
 * `cursor` — an opaque pointer to a page, never to be inspected/constructed by
   clients. It usually (encrypted) encodes the identifier of the first or last
   page element, the pagination direction, and the applied query filters to
--- a/chapters/naming.adoc
+++ b/chapters/naming.adoc
@@ -149,15 +149,19 @@ If you provide query support for searching, sorting, filtering, and
 paginating, you must stick to the following naming conventions:
 
 * `q` — default query parameter (e.g. used by browser tab completion);
-should have an entity specific alias, like sku
-* `sort` — comma-separated list of fields to sort. To indicate sorting
-direction, fields may be prefixed with + (ascending) or - (descending,
-default), e.g. /sales-orders?sort=+id
-* `fields` — to retrieve a subset of fields. See <<157>> below.
-* `embed` — to expand embedded entities (ie.: inside of an article
-entity, expand silhouette code into the silhouette object). Implementing
-"expand" correctly is difficult, so do it with care. See <<158>> below.
-* `cursor` — key-based page start. See <<pagination>> section below.
-* `offset` — numeric offset page start. See <<pagination>> section below.
-* `limit` — to restrict the number of entries. See <<pagination>> section
-below.
+  should have an entity specific alias, like sku
+* `sort` — comma-separated list of fields to define the sort order. To
+  indicate sorting direction, fields may be prefixed with + (ascending) or
+  - (descending, default), e.g. /sales-orders?sort=+id
+* `fields` — to retrieve only a subset of fields of a resource. See <<157>>
+  below.
+* `embed` — to expand or embedded sub-entities (ie.: inside of an article
+  entity, expand silhouette code into the silhouette object). Implementing
+  `embed` correctly is difficult, so do it with care. See <<158>> below.
+* `offset` — numeric offset of the first element on a page. See <<pagination>>
+  section below.
+* `cursor` — encoded pointer to the first or last element on a page, the
+  pagination direction, and the applied query filters. See <<pagination>>
+  section below.
+* `limit` — client suggested limit to restrict the number of entries on
+  a page. See <<pagination>> section below.
--- a/chapters/naming.adoc
+++ b/chapters/naming.adoc
@@ -143,27 +143,23 @@ The trailing slash must not have specific semantics. Resource paths must
 deliver the same results whether they have the trailing slash or not.
 
 [#137]
-== {MAY} Use Conventional Query Strings
+== {MUST} Stick to Conventional Query Parameters
 
-If you provide query support for sorting, pagination, filtering
-functions or other actions, use the following standardized naming
-conventions:
+If you provide query support for searching, sorting, filtering, and
+paginating, you must stick to the following naming conventions:
 
 * `q` — default query parameter (e.g. used by browser tab completion);
 should have an entity specific alias, like sku
-* `limit` — to restrict the number of entries. See Pagination section
-below. Hint: You can use size as an alternate query string.
+* `sort` — comma-separated list of fields to sort. To indicate sorting
+direction, fields may be prefixed with + (ascending) or - (descending,
+default), e.g. /sales-orders?sort=+id
+* `fields` — to retrieve a subset of fields. See <<157>> below.
+* `embed` — to expand embedded entities (ie.: inside of an article
+entity, expand silhouette code into the silhouette object). Implementing
+"expand" correctly is difficult, so do it with care. See <<158>> below.
 * `cursor` — key-based page start. See <<pagination>> section below.
 * `offset` — numeric offset page start. See <<pagination>> section below.
 Hint: In combination with limit, you can use page as an alternative to
 offset.
-* `sort` — comma-separated list of fields to sort. To indicate sorting
-direction, fields may be prefixed with + (ascending) or - (descending,
-default), e.g. /sales-orders?sort=+id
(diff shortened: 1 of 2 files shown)


chapters/naming.adoc
```diff
@@ -143,27 +143,23 @@ The trailing slash must not have specific semantics. Resource paths must
...
+* `fields` — to retrieve a subset of fields. See <<157>> below.
 * `embed` — to expand embedded entities (ie.: inside of an article
 entity, expand silhouette code into the silhouette object). Implementing
-"expand" correctly is difficult, so do it with care.
-
-
+"expand" correctly is difficult, so do it with care. See <<158>> below.
+* `cursor` — key-based page start. See <<pagination>> section below.
+* `offset` — numeric offset page start. See <<pagination>> section below.
+Hint: In combination with limit, you can use page as an alternative to
```
whiskeysierra: `page` with code ticks and doesn't page have a different semantic then offset?

My understanding:
- offset is the absolute number of records
- page is offset/limit
- limit can be understood as a page size

tkrop: Damned, I missed this. I wanted to remove this too.

whiskeysierra:
> Do you think we should keep `page` and `size` as an alternative?

No, but we may add better descriptions to each parameter to make their semantic clear.

tkrop:
@whiskeysierra good idea.

chapters/naming.adoc
```diff
@@ -149,15 +149,19 @@ If you provide query support for searching, sorting, filtering, and
...
-entity, expand silhouette code into the silhouette object). Implementing
-"expand" correctly is difficult, so do it with care. See <<158>> below.
-* `cursor` — key-based page start. See <<pagination>> section below.
-* `offset` — numeric offset page start. See <<pagination>> section below.
-* `limit` — to restrict the number of entries. See <<pagination>> section
-below.
+  should have an entity specific alias, like sku
+* `sort` — comma-separated list of fields to define the sort order. To
+  indicate sorting direction, fields may be prefixed with + (ascending) or
+  - (descending, default), e.g. /sales-orders?sort=+id
```
whiskeysierra: Why is descending the default?

whiskeysierra: (I realize that you didn't introduce that, just asking.)

chapters/naming.adoc
```diff
@@ -149,15 +149,19 @@ If you provide query support for searching, sorting, filtering, and
...
+* `fields` — to retrieve only a subset of fields of a resource. See <<157>>
+  below.
+* `embed` — to expand or embedded sub-entities (ie.: inside of an article
+  entity, expand silhouette code into the silhouette object). Implementing
+  `embed` correctly is difficult, so do it with care. See <<158>> below.
+* `offset` — numeric offset of the first element on a page. See <<pagination>>
+  section below.
+* `cursor` — encoded pointer to the first or last element on a page, the
```
whiskeysierra: Isn't that an implementation detail? From an abstract perspective the cursor just addresses a single page, doesn't it?

chapters/naming.adoc
tkrop: @whiskeysierra I also only guess, but I think order is often based on `date`. Thus descending might make sense to see the most resent results first. But I would like to question this as you. Lets remove this!

chapters/naming.adoc
tkrop: @whiskeysierra Yes, it is an implementation detail, that describes best practice. It is based on my experience in the workshops that many developers have no idea how to start. Thus, I think it might be helpful to have a description about what a cursor is about. When I start to explain this, I often see people starting to understand this concept. What do you think.

whiskeysierra: I'm skipping over my belief that we shouldn't document `cursor` at all and I might open up that discussion separately.

Let's try to split apart the API and implementation aspects more clearly then. Clients should only care about the contract that a cursor provides (opaque value, never to be inspected/constructed, addresses a single page) while providers care about implementation hints (cursor encodes direction although I'm still not a big fan of that, forward and backward cursors address the first/last item of the page respectively).

Mixing both in a sentence creates some confusion that we can try to avoid.

chapters/naming.adoc
```diff
@@ -143,27 +143,27 @@ The trailing slash must not have specific semantics. Resource paths must
...
+  sorting direction, fields may be prefixed with `+` (ascending) or `-`
+  (descending), e.g. /sales-orders?sort=+id
+* `fields` — to retrieve only a subset of fields of a resource. See <<157>>
+  below.
+* `embed` — to expand or embedded sub-entities (ie.: inside of an article
+  entity, expand silhouette code into the silhouette object). Implementing
+  `embed` correctly is difficult, so do it with care. See <<158>> below.
+* `offset` — numeric offset of the first element on a page. See <<pagination>>
+  section below.
+  paque value, never to be inspected/constructed, addresses a single page
```
whiskeysierra: *opaque*

chapters/naming.adoc
```diff
@@ -143,27 +143,27 @@ The trailing slash must not have specific semantics. Resource paths must
...
+  have an entity specific alias, like sku
+* `sort` — comma-separated list of fields to define the sort order. To indicate
+  sorting direction, fields may be prefixed with `+` (ascending) or `-`
+  (descending), e.g. /sales-orders?sort=+id
+* `fields` — to retrieve only a subset of fields of a resource. See <<157>>
+  below.
+* `embed` — to expand or embedded sub-entities (ie.: inside of an article
+  entity, expand silhouette code into the silhouette object). Implementing
+  `embed` correctly is difficult, so do it with care. See <<158>> below.
+* `offset` — numeric offset of the first element on a page. See <<pagination>>
```
whiskeysierra: Why do we define offset and cursor independently? Wouldn't it be beneficial for both to be indistinguishable to clients so that servers can migrate from one to the other?

chapters/naming.adoc
```diff
@@ -143,27 +143,27 @@ The trailing slash must not have specific semantics. Resource paths must
...
+  section below.
+  paque value, never to be inspected/constructed, addresses a single page
+* `cursor` — an opaque pointer to a page, never to be inspected/constructed by
+  clients. It usually (encrypted) encodes the identifier of the first or last
+  page element, the pagination direction, and the applied query filters to
```
whiskeysierra: I wouldn't encrypt the query filters in the cursor. Those are query parameters that clients should in fact modify (at least on the very first request when they start their pagination).

chapters/naming.adoc
tkrop: In case of offset, there may be a use case for setting the value manually. Unless we propose a more advanced scheme for pagination links with, the offset is not opaque.

chapters/naming.adoc
tkrop: The behavior of the pagination gets absolutely unpredictable, if you allow to change the query filters, e.g. the selector page element might even not belong to the page making the cursor fail. If you want to change the query filters, you have to paginate from begin. Thus, it is best practice to include the filters into the cursor, and neglect all query filters given in addition.

chapters/naming.adoc
tkrop: Sorry, intended to remove this fragment. Not supposed to be part of `offset` definition.

ddufourboivin-zalando:
👍

tkrop:
:+1:

chapters/naming.adoc
whiskeysierra: > If you want to change the query filters, you have to paginate from begin.

Agreed, but that requirement in itself doesn't dictate that all other query parameters need to be encoded/encrypted inside the cursor. It just needs to be part of the contract that clients are not allowed/meant to modify the pagination links in any which way.

chapters/naming.adoc
```diff
@@ -143,27 +143,26 @@ The trailing slash must not have specific semantics. Resource paths must
...
-
-
+== {MUST} Stick to Conventional Query Parameters
+
+If you provide query support for searching, sorting, filtering, and
+paginating, you must stick to the following naming conventions:
+
+* `q` — default query parameter (e.g. used by browser tab completion); should
+  have an entity specific alias, like sku
+* `sort` — comma-separated list of fields to define the sort order. To indicate
```
maxim-tschumak: Does it have to be _comma-separated_? We specify the format for collection query parameters in [154](https://opensource.zalando.com/restful-api-guidelines/#154). A link to the rule would be nice.

maxim-tschumak:
:+1:

PR #455, opened 2018-09-26
add X-Mobile-Advertising-Id header

hafizh:
Add mobile advertising id header.

### DIFF
--- a/chapters/proprietary-headers.adoc
+++ b/chapters/proprietary-headers.adoc
@@ -86,10 +86,9 @@ as generic aspect. Current range is iOS, Android, Windows, Linux, MacOS.
 Provided by mobile apps only. It is a unique, customer-resettable identifier provided
 by mobile device's operating system for advertising purposes. It is not sent if 
 the customer disables it in the settings for respective platform. The value 
-can either be Apple IDFA (iOS, see 
-https://developer.apple.com/documentation/adsupport/asidentifiermanager[Apple developer, 
-AdSupport]) or the Goggle AddID (Android, see 
-https://support.google.com/admanager/answer/6274238[Goggle Ad Manager]). 
+can either be https://developer.apple.com/documentation/adsupport/asidentifiermanager[Apple 
+IDFA] (iOS) or the https://support.google.com/googleplay/android-developer/answer/6048248[Goggle 
+AddID] (Android)
 |b89fadce-1f42-46aa-9c83-b7bc49e76e1f
 
 |[[x-app-domain]]{X-App-Domain}|Integer|
--- a/chapters/proprietary-headers.adoc
+++ b/chapters/proprietary-headers.adoc
@@ -82,6 +82,10 @@ e.g. smartphone Android vs. iOS. Via this header info should be passed-through
 as generic aspect. Current range is iOS, Android, Windows, Linux, MacOS.
 |Android
 
+|X-Advertising-ID |String |An app specific, unique customer resettable 
+id for advertising purposes. It is not sent if the customer disables it 
+in the settings for respective platform. |b89fadce-1f42-46aa-9c83-b7bc49e76e1f
+
 |[[x-app-domain]]{X-App-Domain}|Integer|
 The app domain (i.e. shop channel context) of the request. Note, app-domain is
 a legacy concept that will be replaced in new platform by combinations of main
--- a/chapters/proprietary-headers.adoc
+++ b/chapters/proprietary-headers.adoc
@@ -82,9 +82,15 @@ e.g. smartphone Android vs. iOS. Via this header info should be passed-through
 as generic aspect. Current range is iOS, Android, Windows, Linux, MacOS.
 |Android
 
-|X-Advertising-ID |String |An app specific, unique customer resettable 
-id for advertising purposes. It is not sent if the customer disables it 
-in the settings for respective platform. |b89fadce-1f42-46aa-9c83-b7bc49e76e1f
+|[[x-mobile-advertising-id]]{X-Mobile-Advertising-Id}|String |
+Provided by mobile apps only. It is a unique, customer-resettable identifier provided
+by mobile device's operating system for advertising purposes. It is not sent if 
+the customer disables it in the settings for respective platform. The value 
+can either be Apple IDFA (iOS, see 
+https://developer.apple.com/documentation/adsupport/asidentifiermanager[Apple developer, 
+AdSupport]) or the Goggle AddID (Android, see 
+https://support.google.com/admanager/answer/6274238[Goggle Ad Manager]). 
+|b89fadce-1f42-46aa-9c83-b7bc49e76e1f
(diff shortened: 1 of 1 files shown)


tkrop:
@hafizh Thanks for suggesting. We have to talk in our next API Guild meeting  about this suggestion. Can you give more argumentation here, why and what you need this for?

hafizh:
This is a header Apps pass for all requests, that also all underlying dependencies accept and pass to nakadi events, tracking or recommentation services. So, @bocytko suggested we add it here, so we don't have to align on the correct naming, with every team expecting it and standardize this.

chapters/proprietary-headers.adoc
```diff
@@ -87,6 +87,11 @@ differ between device platform, e.g. smartphone Android vs. iOS. Via
 this header info should be passed-through as generic aspect. Current
 range is iOS, Android, Windows, Linux, MacOS |Android
 
+
+|X-Advertising-ID |String |An app specific, unique customer resettable 
+id for advertising purposes. It is not sent if the customer disables it 
+in the settings for respective platform. |b89fadce-1f42-46aa-9c83-b7bc49e76e1f
```
ePaul: Can we get a bit more detail? "An ID for advertising purposes" is totally non-clear to me, maybe because I'm very far away from the customer-facing applications (we certainly don't want advertising in our internal or supplier-facing tools).

tkrop:
@hafizh Can you join our guild meeting to elaborate on this topic? I send you an invite.

tkrop:
@bocytko can you say something to this issue? Otherwise we will close this issue.

chapters/proprietary-headers.adoc
bocytko: See mobile apps advertising id: [1](https://support.google.com/admanager/answer/6274238?hl=en), [2](https://www.google.de/search?q=mobile+apps+advertising+id&oq=mobile+apps+advertising+id&aqs=chrome..69i57j0l5.6048j0j7&sourceid=chrome&ie=UTF-8).

bocytko:
@ruiaraujo can you help here with the description of the header (maybe you have it already in the FSA or skipper)?

@tkrop are you refusing to accept this header altogether or are your just unsatisfied with the description?

tkrop:
@bocytko We do not refuse generally. But, because we discourage the use of proprietary headers except for one well defined use case described in our [Properietary Headers Section](https://opensource.zalando.com/restful-api-guidelines/#proprietary-headers), we expect the requirements around it to be explained. Currently, we can not decide from the PR whether this `X-Advertising-ID` is matching our use case for a Proprietary Header or whether it should be modeled as query/path or body parameter.

It would help if @ruiaraujo, @hafizh,  or you can join our next API guild meeting to explain the usage pattern and give more context.

ruiaraujo:
@tkrop How can we get an invite?

@bocytko We are not handling it yet but we will.

tkrop:
@ruiaraujo done.

tfrauenstein:
@hafizh @ruiaraujo As you remember, we discussed this PR in Guild meeting on 06.11. We felt that clearer definition of use cases, format, semantics and client / server obligations is needed to standardize the "x-Advertising-id' in the Guidelines. 

Some more concrete questions:
- How does the  "x-Advertising-id" is related to the (privacy regulation compliant) IDFA (Apple IDentifier For mobile Advertising) for iOS and the GAID (Google mobile Advertising IDentifier) for Android? 
- Is it simply a placeholder for either the IDFA or GAID depending on the mobile app platform? If yes, and make it clear in the definition (incl. references to format definition etc.)
- It is supposed to be only provided by customer facing mobile apps communicating with online services, is it? If yes, we should call it "x-Mobile-Advertising-id'.
- Is it really a pass through parameter that should be supported by all mobile called backend services including its called services (either via APIs or events)?

Will you provide a PR update that provides a clearer definition and explicit statements around the questions above? Otherwise we should close the PR. Thank you!

hafizh:
@tfrauenstein Thank you for taking your time on this. The main take away from the Guild meeting for me was to find out if this header/id is something that also applies for web. If so to come up and align on a more generic naming and semantics. However, this turned out to be challenging since a lot of teams needs to be involved in this. Also looking into it more we didn't see it as applicable for web. 
To answer your questions: 

> How does the "x-Advertising-id" is related to the (privacy regulation compliant) IDFA (Apple IDentifier For mobile Advertising) for iOS and the GAID (Google mobile Advertising IDentifier) for Android?
> Is it simply a placeholder for either the IDFA or GAID depending on the mobile app platform? If yes, and make it clear in the definition (incl. references to format definition etc.)

Yes, it is a placeholder for IDFA for iOS and GAID for Android. Will update description. 

> It is supposed to be only provided by customer facing mobile apps communicating with online services, is it? If yes, we should call it "x-Mobile-Advertising-id'.

That sounds reasonable. Will update. 

> Is it really a pass through parameter that should be supported by all mobile called backend services including its called services (either via APIs or events)?

In a way, yes! But in general it is an optional parameter. The apps can always send it when available. Weather the backend supports it or not depends on their use cases. If they provide such features that are linked to this parameter or not. 
I will try to articulate this into the description as well.

whiskeysierra:
:+1:

maxim-tschumak:
:+1:

chapters/proprietary-headers.adoc
```diff
@@ -82,6 +82,15 @@ e.g. smartphone Android vs. iOS. Via this header info should be passed-through
 as generic aspect. Current range is iOS, Android, Windows, Linux, MacOS.
 |Android
 
+|[[x-mobile-advertising-id]]{X-Mobile-Advertising-Id}|String |
+Provided by mobile apps only. It is a unique, customer-resettable identifier provided
+by mobile device's operating system for advertising purposes. It is not sent if 
+the customer disables it in the settings for respective platform. The value 
+can either be https://developer.apple.com/documentation/adsupport/asidentifiermanager[Apple 
+IDFA] (iOS) or the https://support.google.com/googleplay/android-developer/answer/6048248[Goggle 
+AddID] (Android)
```
tfrauenstein: Basically looks fine. But I propose following editorial update:

It is either the https://developer.apple.com/documentation/adsupport/asidentifiermanager[IDFA] (Apple Identifier for mobile Advertising) for iOS, or the https://support.google.com/googleplay/android-developer/answer/6048248[GAID]  (Google mobile Advertising Identifier) for Android. It is a unique, customer-resettable identifier provided by mobile device’s operating system for advertising purposes, and usually passed by mobile apps via http header when calling backend services. Called services should be ready to pass this parameter through when calling other services. It is not sent if the customer disables it in the settings for respective mobile platform.

tfrauenstein:
@hafizh thx -- basically looks fine, however I proposed concrete editorial update

tfrauenstein:
👍

tkrop:
:+1:

PR #461, opened 2018-11-01
fix: clarify the usage of legacy hostnames for APIs (#460)

### LINKED ISSUE #460: Make it clear, which APIs are allowed to use legacy hostnames
https://opensource.zalando.com/restful-api-guidelines/#224 should allow "legacy" hostnames for **component-internal** APIs (instead of internal)

maxim-tschumak:
Fixes #460


tkrop:
:+1:

maxim-tschumak:
:+1:

PR #463, opened 2018-11-13
Fix verbatim formatting

FRosner:
# Problem

The "and" between `ETag` and `If-Match` is formatted as verbatim together with the enclosing words.

image

# Fix

I included an additional whitespace to make `ETag` and `If-Match` appear as separate verbatim formatted words.

image


whiskeysierra:
:+1:

tkrop:
:+1:

PR #476, opened 2019-01-03
should not use /api as base path (#462)

### LINKED ISSUE #462: Remove rule "May Use /api as first Path Segment [135]"
https://opensource.zalando.com/restful-api-guidelines/#135

Alternatively we may choose to switch it from "May" to "May not". It doesn't serve any significant purpose anymore and just pollutes the URL space.

maxim-tschumak:
Closes #462


ePaul:
:+1:

tkrop:
:+1:

PR #479, opened 2019-01-15
feat: add idempotency guidance (#456)

### LINKED ISSUE #456: Idempotency Guidelines
In several past reviews we ended up discussing (variants of) the following three major techniques on how to implement idempotency:

1. Conditional Requests, [RFC 7232](https://tools.ietf.org/html/rfc7232)
2. *Foreign Key* Pattern
3. Idempotency Key, https://stripe.com/docs/api#idempotent_requests

I'd like to use this issue to kick-off a focused discussion on those techniques. My goal is to extend the guidelines with a table that compares them in an objective manner:

|                                 | Conditional Requests | *Foreign Key* Pattern | Idempotency Key |
|------------------------------------|--------------------------|----------------------------|---------------------------|
| Applied with HTTP methods       | `PUT`/`PATCH`                  | `POST`                   | `PUT`/`POST`/`PATCH`       |
| HTTP Standard                   | ✔️ Yes                | ✖️ No                     | ✖️ No                     |
| Prevents concurrent creation  | ✔️ Yes              | ✔️ Yes                  | ✖️ No                    |
| Prevents lost updates           | ✔️ Yes                | ✖️ No                    | ✖️ No                   |
| Supports safe retries           | ⚠️ With effort                  | ⚠️ With effort        | ✔️ Yes                 |
| Inspectable (by intermediaries) | ✔️ Yes                | ✖️ No                 | ✔️ Yes                |
| Exact same response             | ✖️ No             | ✖️ No     | ⚠️ With effort           |
| Requires additional `GET`<br>(before `PUT`/`POST`/`PATCH`) | ✖️ Yes   | ✔️ No   | ✔️ No        |

I don't claim that the table is complete. Let's use it as a starting point and extend/modify it as we go along. @tkrop It would also be helpful if you write 1-2 sentences on the foreign key pattern (I know I asked you this before somewhere but I can't find your answer anymore. Sorry!)

tkrop:
Fixes #456.

### DIFF
--- a/chapters/common-headers.adoc
+++ b/chapters/common-headers.adoc
@@ -123,32 +123,28 @@ in https://tools.ietf.org/html/rfc7240[RFC7240] to indicate whether the
 preference was applied.
 
 [#182]
-== {MAY} Consider Using ETag Together With If-Match/If-None-Match Header
+== {MAY} Consider Using `ETag` Together With `If-Match`/`If-None-Match` Header
 
-When creating or updating resources it may be necessary to expose
-conflicts and to prevent the 'lost update' or 'initially created' problem. 
-Following https://tools.ietf.org/html/rfc7232[RFC 7232 "HTTP: Conditional Requests"]
-this can be best accomplished by using the
-https://tools.ietf.org/html/rfc7232#section-2.3[`ETag`] header together
-with the https://tools.ietf.org/html/rfc7232#section-3.1[`If-Match`] or
-https://tools.ietf.org/html/rfc7232#section-3.2[`If-None-Match`] conditional header. The
-contents of an `ETag: <entity-tag>` header is either (a) a hash of the
-response body, (b) a hash of the last modified field of the entity, or
-(c) a version number or identifier of the entity version.
+When creating or updating resources it may be necessary to expose conflicts
+and to prevent the 'lost update' or 'initially created' problem. Following
+https://tools.ietf.org/html/rfc7232[RFC 7232 "HTTP: Conditional Requests"]
+this can be best accomplished by using the {ETag} header together {If-Match}
+or {If-None-Match} conditional header. The contents of an {ETag}`: <entity-tag>`
+header is either (a) a hash of the response body, (b) a hash of the last
+modified field of the entity, or (c) a version number or identifier of the
+entity version.
 
-To expose conflicts between concurrent update operations via PUT, POST,
-or PATCH, the `If-Match: <entity-tag>` header can be used to force the
-server to check whether the version of the updated entity is conforming
-to the requested `<entity-tag>`. If no matching entity is found, the
-operation is supposed a to respond with status code 412 - precondition
-failed.
+To expose conflicts between concurrent update operations via {PUT}, {POST},
+or {PATCH}, the `If-Match: <entity-tag>` header can be used to force the server
+to check whether the version of the updated entity is conforming to the
+requested `<entity-tag>`. If no matching entity is found, the operation is
+supposed a to respond with status code 412 - precondition failed.
 
-Beside other use cases, the `If-None-Match:` header with parameter `*`
-can be used in a similar way to expose conflicts in resource creation.
-If any matching entity is found, the operation is supposed a to respond
-with status code 412 - precondition failed.
+Beside other use cases, `If-None-Match: *` can be used in a similar way to
+expose conflicts in resource creation. If any matching entity is found, the
+operation is supposed a to respond with status code 412 - precondition failed.
 
-The `ETag`, `If-Match`, and `If-None-Match` headers can be defined as
+The {ETag}, {If-Match}, and {If-None-Match} headers can be defined as
 follows in the API definition:
 
 [source,yaml]
@@ -191,4 +187,51 @@ components:
       example: "7da7a728-f910-11e6-942a-68f728c1ba70", *
 ----
 
-Please also see the section <<optimistic-locking>> for a discussion about alternative approaches.
+Please see <<optimistic-locking>> for a detailed discussion and options.
+
+[#230]
+== {MAY} Consider Using `Idempotency-Key` Header
+
+When creating or updating a resource it can be helpful or necessary to prevent
+duplicate execution to ensure idempotency of otherwise non-idempotent requests
+from perspective of the client. This can be simply achieved by sending a client
+specific unique request key via the {Idempotency-Key} header.
+
+The unique request key is not part of the resource. It is stored temporarily,
+e.g. for 24 hours, together with the response and a request hash (optionally).
+The service looks up the key to ensure idempotent behavior by sending the
+response without executing the request a second time. This allows clients to
+safely retry requests after timeouts and receive the same response.
+
+*Note:* To grant the single execution semantic, the request cache and the
+resource have to be created or updated in the same transaction. In this case
+clients may require fallback strategies for retries.
+
+The {Idempotency-Key} header should be defined as follows:
+
+[source,yaml]
+----
+components:
+  headers:
+    Idempotency-Key:
+      description: |
+        The idempotency key is a UUID version 4 (random) created by the client
+        to identify a request. It is used by the service to identify subsequent
+        retries of the same request and ensure idempotent behavior by sending
+        the same response without executing the request a second time.
+
+        Idempotency keys expire after 24 hours. Clients are responsible to stay
+        within this limits, if they require idempotent behavior.
+
+      type: string
+      format: uuid
+      required: false
+      example: "7da7a728-f910-11e6-942a-68f728c1ba70"
+----
+
+*Note:* The {Idempotency-Key} header unlike other headers in this section
+is not standardized in an RFC. Our only reference are the usage in the
+https://stripe.com/docs/api/idempotent_requests[Stripe API]. However, as it
+fit not into our section about <<proprietary-headers>>, and we did not want
+to change the header name and semantic, we decided to treat it as any other
+common header.
--- a/chapters/common-headers.adoc
+++ b/chapters/common-headers.adoc
@@ -138,11 +138,12 @@ To expose conflicts between concurrent update operations via {PUT}, {POST},
 or {PATCH}, the `If-Match: <entity-tag>` header can be used to force the server
 to check whether the version of the updated entity is conforming to the
 requested `<entity-tag>`. If no matching entity is found, the operation is
-supposed a to respond with status code 412 - precondition failed.
+supposed a to respond with status code {412} - precondition failed.
 
 Beside other use cases, `If-None-Match: *` can be used in a similar way to
 expose conflicts in resource creation. If any matching entity is found, the
-operation is supposed a to respond with status code 412 - precondition failed.
+operation is supposed a to respond with status code {412} - precondition
+failed.
 
 The {ETag}, {If-Match}, and {If-None-Match} headers can be defined as
 follows in the API definition:
@@ -193,22 +194,22 @@ Please see <<optimistic-locking>> for a detailed discussion and options.
 == {MAY} Consider Using `Idempotency-Key` Header
 
 When creating or updating resources it can be helpful or necessary to prevent
-duplicate execution to ensure idempotent response behavior. Generally, this can 
-be achieved by sending a client specific _unique request key_ – that is not
-part of the resource – using the {Idempotency-Key} header.
+duplicate execution to ensure <<idempotent>> behavior. Generally, this can be
+achieved by sending a client specific _unique request key_ – that is not part
+of the resource – using the {Idempotency-Key} header.
 
 The _unique request key_ is stored temporarily, e.g. for 24 hours, together
 with the response and a request hash (optionally) in a key cache. The service
 can now look up the _unique request key_ in the key cache and serve the
 response from the key cache, instead of re-executing the request, to ensure
-idempotent behavior. Optionally, it can check the request hash for consistency
-before serving the response. If the key is not in the key store, the request
-is executed as usual and the response is stored in the key cache.
+<<idempotent>> behavior. Optionally, it can check the request hash for
+consistency before serving the response. If the key is not in the key store,
+the request is executed as usual and the response is stored in the key cache.
 
 This allows clients to safely retry requests after timeouts etc while receive
 the same response multiple times.
 
-*Note:* To grant a reliable idempotent execution semantic, the resource and
+*Note:* To grant a reliable <<idempotent>> execution semantic, the resource and
 the key cache have to be updated with hard transaction semantics – considering
 all potential pitfalls of failures, timeouts, and concurrent requests in a
 distributed systems. This makes a correct implementation very hard
--- a/chapters/common-headers.adoc
+++ b/chapters/common-headers.adoc
@@ -212,7 +212,8 @@ This allows clients to safely retry requests after timeouts, network outages,
 etc. while receive the same response multiple times. *Note:* The request retry
 in this context requires to send the exact same request, i.e. updates of the
 request that would change the result are off-limits. The request hash in the
-key cache can protection against this misbehavior.
+key cache can protection against this misbehavior. The service is recommended
+to reject such a request using status code {400}.
 
 *Important:* To grant a reliable <<idempotent>> execution semantic, the
 resource and the key cache have to be updated with hard transaction semantics
@@ -229,10 +230,15 @@ components:
   headers:
     Idempotency-Key:
       description: |
-        The idempotency key is a UUID version 4 (random) created by the client
-        to identify a request. It is used by the service to identify subsequent
+        The idempotency key is a free identifier created by the client to
+        identify a request. It is used by the service to identify subsequent
         retries of the same request and ensure idempotent behavior by sending
         the same response without executing the request a second time.
+
+        Clients should be careful as any subsequent requests with the same key
+        may return the same response without further check. Therefore, it is
+        recommended to use an UUID version 4 (random) or any other random
+        string with enough entropy to avoid collisions.
 
         Idempotency keys expire after 24 hours. Clients are responsible to stay
         within this limits, if they require idempotent behavior.
--- a/chapters/common-headers.adoc
+++ b/chapters/common-headers.adoc
@@ -193,26 +193,32 @@ Please see <<optimistic-locking>> for a detailed discussion and options.
 [#230]
 == {MAY} Consider Using `Idempotency-Key` Header
 
-When creating or updating resources it can be helpful or necessary to prevent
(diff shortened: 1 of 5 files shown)


chapters/http-requests.adoc
```diff
@@ -193,52 +189,95 @@ resource must be accessible after this operation on its endpoint.
...
+  (see <<182>>).
+
+To decide, which pattern is suitable for your use case, please consult the
+following table showing the major properties of each pattern:
+
+[,cols="40%,20%,20%,20%",options="header",]
+|========================================================================================
+|                                      | Conditional Key | Secodary Key | Idempotency Key
+| Applicable with                      | {PUT}/{PATCH} | {POST} | {POST}/{PUT}/{PATCH}
+| HTTP Standard                        | {NO}  | {NO}  | {NO}
```
whiskeysierra: Conditional headers are standard HTTP. See https://tools.ietf.org/html/rfc7232

chapters/http-requests.adoc
```diff
@@ -193,52 +189,95 @@ resource must be accessible after this operation on its endpoint.
...
+  e.g. a _hash_ or _version number_, often stored with it. It allows to detect
+  concurrent creations and updates to ensure idempotent request behavior
+  (see <<182>>).
+
+To decide, which pattern is suitable for your use case, please consult the
+following table showing the major properties of each pattern:
+
+[,cols="40%,20%,20%,20%",options="header",]
+|========================================================================================
+|                                      | Conditional Key | Secodary Key | Idempotency Key
```
whiskeysierra: Typo *secondary*

chapters/http-requests.adoc
```diff
@@ -193,52 +189,95 @@ resource must be accessible after this operation on its endpoint.
...
+| Applicable with                      | {PUT}/{PATCH} | {POST} | {POST}/{PUT}/{PATCH}
+| HTTP Standard                        | {NO}  | {NO}  | {NO}
+| Prevents concurrent creation         | {YES} | {YES} | {NO}
+| Prevents concurrent (lost) updates   | {YES} | {NO}  | {NO} 
+| Supports safe retries                | {YES} | {YES} | {YES}
+| Supports exact same response         | {NO}  | {NO}  | {YES}
+| Can be inspected (by intermediaries) | {YES} | {NO}  | {YES}
+| Usable without additional {GET}
+(before {PUT}/{POST}/{PATCH})          | {YES}  | {YES} | {NO}
```
whiskeysierra: I believe those are all flipped by accident.

chapters/http-requests.adoc
```diff
@@ -193,52 +189,95 @@ resource must be accessible after this operation on its endpoint.
...
+  concurrent creations and updates to ensure idempotent request behavior
+  (see <<182>>).
+
+To decide, which pattern is suitable for your use case, please consult the
+following table showing the major properties of each pattern:
+
+[,cols="40%,20%,20%,20%",options="header",]
+|========================================================================================
+|                                      | Conditional Key | Secodary Key | Idempotency Key
+| Applicable with                      | {PUT}/{PATCH} | {POST} | {POST}/{PUT}/{PATCH}
```
whiskeysierra: What about `DELETE`?

chapters/http-requests.adoc
whiskeysierra: Actually secondary key is correct.

tkrop: Yes, flipped around when switching column 1 and 3.

chapters/http-requests.adoc
tkrop: Might make sense for complete comparison with `Conditional Key` and `Idempotency Key`. However, it is a rare use case to `DELETE` resources conditional and it is a rare use case to `DELETE` resources at all.

@tfrauenstein suggested in the meeting to remove the `PUT` from the table and just add a sentence that patterns applicable for `PATCH` can and should be applied to `PUT` - and now `DELETE`. Before adding `DELETE` to the table, I would now vote for adding such a sentence with `PUT` and `DELETE` instead of adding it to the table.

@zeitlinger @ePaul @gewald @tfrauenstein what are you thinking?

chapters/common-headers.adoc
```diff
@@ -191,4 +188,58 @@ components:
...
+        The idempotency key is a UUID version 4 (random) created by the client
+        to identify a request. It is used by the service to identify subsequent
+        retries of the same request and ensure idempotent behavior by sending
+        the same response without executing the request a second time.
+
+        Idempotency keys expire after 24 hours. Clients are responsible to stay
+        within this limits, if they require idempotent behavior.
+
+      type: string
+      format: uuid
```
duergner: What's the reason to enforce UUID format here?

chapters/common-headers.adoc
```diff
@@ -191,4 +188,58 @@ components:
...
+can now look up the _unique request key_ in the key cache and serve the
+response from the key cache, instead of re-executing the request, to ensure
+<<idempotent>> behavior. Optionally, it can check the request hash for
+consistency before serving the response. If the key is not in the key store,
+the request is executed as usual and the response is stored in the key cache.
+
+This allows clients to safely retry requests after timeouts etc while receive
+the same response multiple times.
+
+*Note:* To grant a reliable <<idempotent>> execution semantic, the resource and
```
duergner: Given the importance of this part I'm not sure whether a simple *Note* is actually enough to emphasize this.

chapters/common-headers.adoc
```diff
@@ -191,4 +188,58 @@ components:
...
+
+The _unique request key_ is stored temporarily, e.g. for 24 hours, together
+with the response and a request hash (optionally) in a key cache. The service
+can now look up the _unique request key_ in the key cache and serve the
+response from the key cache, instead of re-executing the request, to ensure
+<<idempotent>> behavior. Optionally, it can check the request hash for
+consistency before serving the response. If the key is not in the key store,
+the request is executed as usual and the response is stored in the key cache.
+
+This allows clients to safely retry requests after timeouts etc while receive
```
maxim-tschumak: timeouts, networks outages, etc.

chapters/common-headers.adoc
```diff
@@ -191,4 +188,58 @@ components:
...
+[#230]
+== {MAY} Consider Using `Idempotency-Key` Header
+
+When creating or updating resources it can be helpful or necessary to prevent
+duplicate execution to ensure <<idempotent>> behavior. Generally, this can be
+achieved by sending a client specific _unique request key_ – that is not part
+of the resource – using the {Idempotency-Key} header.
+
+The _unique request key_ is stored temporarily, e.g. for 24 hours, together
+with the response and a request hash (optionally) in a key cache. The service
```
maxim-tschumak: Providing the same response is not an idempotency requirement, in my understanding. (e.g. DELETE is idempotent, but may return 404 on the second DELETE request on the same resource)

Important is that the server's state remains the same.

chapters/common-headers.adoc
```diff
@@ -191,4 +188,58 @@ components:
...
+can now look up the _unique request key_ in the key cache and serve the
+response from the key cache, instead of re-executing the request, to ensure
+<<idempotent>> behavior. Optionally, it can check the request hash for
```
maxim-tschumak: I think the _unique request key_ **must** be enough. Calculating the request hash is not straight forward. The implementation has to take into account what parts of the request should the hash be based on (headers, timestamps in the request bodies, paths, path parameters, etc.) - this is very specific for a particular endpoint.

chapters/common-headers.adoc
```diff
@@ -191,4 +188,58 @@ components:
...
+consistency before serving the response. If the key is not in the key store,
+the request is executed as usual and the response is stored in the key cache.
+
+This allows clients to safely retry requests after timeouts etc while receive
+the same response multiple times.
+
+*Note:* To grant a reliable <<idempotent>> execution semantic, the resource and
+the key cache have to be updated with hard transaction semantics – considering
+all potential pitfalls of failures, timeouts, and concurrent requests in a
+distributed systems. This makes a correct implementation very hard
```
maxim-tschumak: Please add `.` at the end.

chapters/common-headers.adoc
```diff
@@ -191,4 +188,58 @@ components:
...
+
+[#230]
+== {MAY} Consider Using `Idempotency-Key` Header
+
+When creating or updating resources it can be helpful or necessary to prevent
+duplicate execution to ensure <<idempotent>> behavior. Generally, this can be
+achieved by sending a client specific _unique request key_ – that is not part
+of the resource – using the {Idempotency-Key} header.
+
+The _unique request key_ is stored temporarily, e.g. for 24 hours, together
```
maxim-tschumak: I'm not sure if a temporary cache is good enough: idempotency should not have time constraints, IMO.

I think a simple persisting of all the processed requests is a better solution.

ePaul:
:+1: for the actual rule changes, I didn't review the formatting stuff in detail.

chapters/common-headers.adoc
ePaul: Maybe we need to find a better name for this, like "idempotent with same-response", or "strongly idempotent"?

chapters/http-requests.adoc
tkrop: Changed it to just mentioning `PUT` and `DELETE` after talking with @tfrauenstein.

chapters/common-headers.adoc
tkrop: My thinking was, that providing freedom here encourages people to get creative with unexpected side effects. I wanted to prevent this and thereby support efficient library development for this feature. However, thinking twice about this, prefixing might come in handy too, which is excluded by this.

I'm open to accept change request to relax this requirement.
@tfrauenstein @whiskeysierra @duergner @maxim-tschumak @zeitlinger @ePaul @gewald Please vote with thumbs up/down here for relaxing.

chapters/common-headers.adoc
tkrop: We discussed this in the meeting :).

I agree that a persistent storage would be preferable. However, as this data must be stored for every request, it will create an event stream that grow to an much bigger extend than the actual data, which is even not needed in 99.9% of the time. Anyhow, as this has a high impact on performance and maintainability of this event stream data, I think it is best practice to restrict the active (visible) portion of this stream to a small time window and communicate this restriction.

chapters/common-headers.adoc
maxim-tschumak: I would "vote" for **not** restricting it to UUID. I also favor the prefixing + internal (compound) identifier (e.g. `order-id`) approach. 

I see also a potential problem with a generated UUID: it only works if the retry is implemented on request-level. A retry mechanism in the application logic is not possible since it would have to "generate the same UUID".

chapters/common-headers.adoc
tkrop: @maxim-tschumak @ePaul The main feature of the `Idempotency-Key` is, that it provides not only `idempotent` server state but also the *same response* to a client. If this is not clear from the section, please state, because than I need to make this clearer her.

maxim-tschumak: @ePaul Agree if we really want to enforce this.

maxim-tschumak: @tkrop It's clear from the text. I just don't really understand that this has great value? (aka why?)

chapters/common-headers.adoc
tkrop: I agree in so far, that we talk about a looking up a restricted time interval. However, I have extracted this idea from looking on stripe that also checks the request. The reason for proposing this is, that it allows infrastructure like implementation that can apply this pattern to billions of requests without risking a collision.

I do not see a big problem in suggesting this, as it is optional and determining what can/should be hashed in general is not as difficult as it sounds.

chapters/common-headers.adoc
maxim-tschumak: My thinking was: usually, you have multiple instances of a server. Because of this, you have to synchronize the processed requests _cache_. A server has usually a persistence (state) anyway and it can be used for such technical information as processed request keys (per operation/resource/etc.).

chapters/common-headers.adoc
tkrop: Will update this.

chapters/common-headers.adoc
maxim-tschumak: It's clear from the text :)

chapters/common-headers.adoc
tkrop: Will do.

chapters/common-headers.adoc
tkrop: Will change this to *Important:*!!

tkrop:
@duergner @maxim-tschumak @ePaul Thanks for reviewing so far. Has one had the time to although have a look at spelling mistakes and grammar so far?

chapters/common-headers.adoc
tfrauenstein: 👍 I would relax it, but keep the hint that we recommend to use UUID version 4, or other random string values without (practically relevant risk of) collusions.

tfrauenstein:
👍  Nice work, thx.

chapters/common-headers.adoc
tkrop: Yes, that is where the key cache will normally be located - I mentioned the transactional requirement between writing the cache entry and the resource, didn't I ;).

How does this influence, the above statement of me?

maxim-tschumak: If this information is persisted anyway - why should it have a TTL?

tkrop: The reason is, that you have to clean up the storage to keep your speed and this limits the visible time. So better to define a limit upfront instead of surprising your clients after a performance problem.

chapters/http-requests.adoc
```diff
@@ -165,80 +159,167 @@ http://tools.ietf.org/html/rfc6902[JSON Patch] can shown its full power while
...
+  allows to ensure <<idempotent>> behavior by looking up the resource in case
+  of multiple independent resource creations from different clients (see
+  <<231>>).
+* A client specific *idempotency key* provided via {Idempotency-Key} header
+  in the request. The key is not part of the resource but stored temporarily
+  pointing to the original response to ensure <<idempotent>> behavior when
+  retrying a request (see <<230>>).
+
+*Note:* While *conditional key* and *secondary key* are focused on handling
+concurrent requests, the *idempotency key* is focused on providing exact same
```
vadeg: I think it conflicts with [note above](https://github.com/zalando/restful-api-guidelines/pull/479/files#diff-c98f51639d3d10ea65dae4c0de86666fR224):
```
*Note:* this does not mean that it is returning the same response or status code.
```

chapters/common-headers.adoc
tkrop: Discussed personally and added some helping remarks.

chapters/common-headers.adoc
tkrop: Completed sentence.

chapters/common-headers.adoc
tkrop: Discussed personally and added some helping remarks.

chapters/http-requests.adoc
```diff
@@ -165,80 +159,167 @@ http://tools.ietf.org/html/rfc6902[JSON Patch] can shown its full power while
...
+| {POST}      | {NO}  | {AT} No, however you <<229>>.
+| {PUT}       | {NO}  | {YES}
+| {PATCH}     | {NO}  | {AT} No, however you <<229>>.
+| {DELETE}    | {NO}  | {YES}
+| {OPTIONS}   | {YES} | {YES}
+| {TRACE}     | {YES} | {YES}
+|====================================================
+
+[#229]
+== {SHOULD} Consider To Design `POST` and `PATCH` Idempotent
```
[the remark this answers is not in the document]
tkrop: Thanks, nice discussion start.

I personally would like it to be understood this way. There are to many drawbacks when having non-idempotent behavior, e.g. for create zombie or duplicate resources, for update incorrect values, etc.

I would vote clearly for `SHOULD`.

chapters/common-headers.adoc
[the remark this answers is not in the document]
tkrop: Yes, I forgot to mention this. Had it on my list, but ... will add this.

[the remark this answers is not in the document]
tkrop: Done.

tkrop:
@vadeg @maxim-tschumak @tfrauenstein @duergner @whiskeysierra @ePaul @zeitlinger @gewald Thanks for all the great feedback and discussions. I have tried to incorporate all of them. Lets go into the last round for approvals, prove reading, and or voting for https://github.com/zalando/restful-api-guidelines/pull/479#discussion_r248955796.

I would really like to merge this work today :smile_cat:!

maxim-tschumak:
:+1:

tfrauenstein:
👍

vadeg:
👍

chapters/http-requests.adoc
```diff
@@ -165,80 +159,167 @@ http://tools.ietf.org/html/rfc6902[JSON Patch] can shown its full power while
...
+* A resource specific *conditional key* provided via <<182,`If-Match` header>>
+  in the request. The key is in general a meta information of the resource,
+  e.g. a _hash_ or _version number_, often stored with it. It allows to detect
+  concurrent creations and updates to ensure <<idempotent>> behavior (see
+  <<182>>).
+* A resource specific *secondary key* provided as resource property in the
+  request body. The _secondary key_ is stored permanently in the resource. It
+  allows to ensure <<idempotent>> behavior by looking up the resource in case
+  of multiple independent resource creations from different clients (see
+  <<231>>).
```
cjbooms: Can you explain this a bit better? What is a a "resource specific *secondary key*" ?

tkrop: Exactly. Do you think I should add "(property with uniqueness constraint)" here? Or would it be enough to add this to rule 231, that is linked here?

tkrop: @cjbooms tried to be clearer about uniqueness requirements. Please have a look.

cjbooms: I think `property with uniqueness constraint` enforced server-side is a more concise definition. It covers the wide variety of ways this could be achieved, and is not at all ambiguous.

tkrop: @cjbooms I have reworded this: please read https://github.com/zalando/restful-api-guidelines/pull/479/files#diff-c98f51639d3d10ea65dae4c0de86666fR266 ff and https://github.com/zalando/restful-api-guidelines/pull/479/files#diff-c98f51639d3d10ea65dae4c0de86666fR310 ff to look whether it is sufficient.

maxim-tschumak:
:+1:

tfrauenstein:
👍

PR #482, opened 2019-01-21
feat: improve guidance for caching support (#451)

### LINKED ISSUE #451: Document use cases and best practices for cache support.
To support performance in APIs for master data resources, we should describe how to specify caching support.

tkrop:
fixes #451.

The main contributions are in [chapters/performance.adoc:173-309](https://github.com/zalando/restful-api-guidelines/pull/482/files#diff-f409f6bb629f68bac5ace6b270f40a3cR173) and a small extend in [chapters/http-requests.adoc:229-250](https://github.com/zalando/restful-api-guidelines/pull/482/files#diff-c98f51639d3d10ea65dae4c0de86666fR229).

Sorry, that I again mixed some simple reformatting stuff in this commit - especially in chapters/common-headers.adoc. Unfortunately, the diff cannot show this very well.

However, I also mixed an extension to `Prefer` in [chapters/common-headers.adoc:99-155](https://github.com/zalando/restful-api-guidelines/pull/482/files#diff-667f49c57a44c87e7eebac00e5f86cbbR99).

### DIFF
--- a/chapters/common-headers.adoc
+++ b/chapters/common-headers.adoc
@@ -1,69 +1,60 @@
 [[common-headers]]
 = Common Headers
 
-This section describes a handful of headers, which we found raised the
-most questions in our daily usage, or which are useful in particular
-circumstances but not widely known.
+This section describes a handful of headers, which we found raised the most
+questions in our daily usage, or which are useful in particular circumstances
+but not widely known.
 
 [#178]
 == {MUST} Use Content Headers Correctly
 
-Content or entity headers are headers with a `Content-` prefix. They
-describe the content of the body of the message and they can be used in
-both, HTTP requests and responses. Commonly used content headers include
-but are not limited to:
+Content or entity headers are headers with a `Content-` prefix. They describe
+the content of the body of the message and they can be used in both, HTTP
+requests and responses. Commonly used content headers include but are not
+limited to:
 
-* https://tools.ietf.org/html/rfc6266[`Content-Disposition`] can
-indicate that the representation is supposed to be saved as a file, and
-the proposed file name.
-* https://tools.ietf.org/html/rfc7231#section-3.1.2.2[`Content-Encoding`]
-indicates compression or encryption algorithms applied to the content.
-* https://tools.ietf.org/html/rfc7230#section-3.3.2[`Content-Length`]
-indicates the length of the content (in bytes).
-* https://tools.ietf.org/html/rfc7231#section-3.1.3.2[`Content-Language`]
-indicates that the body is meant for people literate in some human
-language(s).
-* https://tools.ietf.org/html/rfc7231#section-3.1.4.2[`Content-Location`]
-indicates where the body can be found otherwise
-(<<179>> for more details]).
-* https://tools.ietf.org/html/rfc7233#section-4.2[`Content-Range`] is
-used in responses to range requests to indicate which part of the
-requested resource representation is delivered with the body.
-* https://tools.ietf.org/html/rfc7231#section-3.1.1.5[`Content-Type`]
-indicates the media type of the body content.
+* {Content-Disposition} can indicate that the representation is supposed to be
+  saved as a file, and the proposed file name.
+* {Content-Encoding} indicates compression or encryption algorithms applied to
+  the content.
+* {Content-Length} indicates the length of the content (in bytes).
+* {Content-Language} indicates that the body is meant for people literate in
+  some human language(s).
+* {Content-Location} indicates where the body can be found otherwise (<<179>>
+  for more details]).
+* {Content-Range} is used in responses to range requests to indicate which part
+  of the requested resource representation is delivered with the body.
+* {Content-Type} indicates the media type of the body content.
 
 [#133]
 == {MAY} Use Standardized Headers
 
-Use http://en.wikipedia.org/wiki/List_of_HTTP_header_fields[this list]
-and mention its support in your OpenAPI definition.
+Use http://en.wikipedia.org/wiki/List_of_HTTP_header_fields[this list] and
+mention its support in your OpenAPI definition.
 
 [#179]
 == {MAY} Use Content-Location Header
 
-The Content-Location header is _optional_ and can be used in successful
-write operations (PUT, POST or PATCH) or read operations (GET, HEAD) to
+The {Content-Location} header is _optional_ and can be used in successful write
+operations ({PUT}, {POST}, or {PATCH}) or read operations ({GET}, {HEAD}) to
 guide caching and signal a receiver the actual location of the resource
-transmitted in the response body. This allows clients to identify the
-resource and to update their local copy when receiving a response with
-this header.
+transmitted in the response body. This allows clients to identify the resource
+and to update their local copy when receiving a response with this header.
 
-The Content-Location header can be used to support the following use
-cases:
+The Content-Location header can be used to support the following use cases:
 
-* For reading operations GET and HEAD, a different location than the
-requested URI can be used to indicate that the returned resource is
-subject to content negotiations, and that the value provides a more
-specific identifier of the resource.
-* For writing operations PUT and PATCH, an identical location to the
-requested URI can be used to explicitly indicate that the returned
-resource is the current representation of the newly created or updated
-resource.
-* For writing operations POST and DELETE, a content location can be used
-to indicate that the body contains a status report resource in response
-to the requested action, which is available at provided location.
+* For reading operations {GET} and {HEAD}, a different location than the
+  requested URI can be used to indicate that the returned resource is subject
+  to content negotiations, and that the value provides a more specific
+  identifier of the resource.
+* For writing operations {PUT} and {PATCH}, an identical location to the
+  requested URI can be used to explicitly indicate that the returned resource
+  is the current representation of the newly created or updated resource.
+* For writing operations {POST} and {DELETE}, a content location can be used to
+  indicate that the body contains a status report resource in response to the
+  requested action, which is available at provided location.
 
-*Note*: When using the Content-Location header, the Content-Type header
+*Note*: When using the {Content-Location} header, the {Content-Type} header
 has to be set as well. For example:
 
 [source,http]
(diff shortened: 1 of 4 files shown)


chapters/http-requests.adoc
```diff
@@ -234,17 +237,17 @@ intended by the operation so that it can be held accountable.
 
 Method implementations must fulfill the following basic properties:
 
-[cols="25%,25%,50%",options="header",]
+[cols="15%,15%,35%,35%",options="header",]
 |====================================================
-| HTTP Method | Safe  | Idempotent
-| {GET}       | {YES} | {YES}
-| {HEAD}      | {YES} | {YES}
-| {POST}      | {NO}  | {AT} No, however you <<229>>.
```
zeitlinger: would rename AT to SHOULD

chapters/http-requests.adoc
```diff
@@ -234,17 +237,17 @@ intended by the operation so that it can be held accountable.
...
-| {PUT}       | {NO}  | {YES}
-| {PATCH}     | {NO}  | {AT} No, however you <<229>>.
-| {DELETE}    | {NO}  | {YES}
-| {OPTIONS}   | {YES} | {YES}
-| {TRACE}     | {YES} | {YES}
+| Method    | Safe  | Idempotent | Cacheable
+| {GET}     | {YES} | {YES}      | {YES}, but <<227>>
+| {HEAD}    | {YES} | {YES}      | {YES}, but <<227>>
+| {POST}    | {NO}  | {AT} No, but <<229>> | {AT} May, but <<227>>
+| {PUT}     | {NO}  | {YES}      | {NO}
```
zeitlinger: why are PUT and DELETE not cacheable?

chapters/performance.adoc
```diff
@@ -4,48 +4,47 @@
...
-resources (see <<182>>
-* pagination for incremental access of larger (result) lists
+* compression of request and response bodies (see <<156>>) 
+* querying field filters to retrieve a subset of resource attributes (see
+  <<157>> below)
+* {ETag} and {If-Match}/{If-None-Match} headers to avoid re-fetching of
+  unchanged resources (see <<182>>)
+* <<pagination>> for incremental access of larger collections of data items
+* caching of master data items, i.e. resources that change rarely or not
+  at all after creation (see <<227>>).
```
zeitlinger: Is the reference to the Prefer header missing (for minimal or representation)?

chapters/http-requests.adoc
ePaul: Because they usually change the state of their resource?
(I guess you could cache a 404 here for some time.)

chapters/performance.adoc
```diff
@@ -4,48 +4,47 @@
...
-Depending on your use case and payload size, you can significantly
-reduce network bandwidth need by supporting filtering of returned entity
-fields. Here, the client can determine the subset of fields he wants to
-receive via the fields query parameter — example see
+Depending on your use case and payload size, you can significantly reduce
+network bandwidth need by supporting filtering of returned entity fields. Here,
+the client can determine the subset of fields he wants to receive via the
+fields query parameter — example see
 https://cloud.google.com/appengine/docs/python/taskqueue/rest/performance#partial-response[Google
 AppEngine API's partial response]:
```
zeitlinger: it should be "_embed" instead of "_embedded" as we discussed last time (IIRC)

Note that this is not part of this changeset

chapters/http-requests.adoc
zeitlinger: yes, 404 for delete and the payload of PUT for subsequent GETs

zeitlinger:
finished the review now - using the diff from IDEA fixed the problems of seeing what has changed.

chapters/http-requests.adoc
tkrop: AT is short cut for ATTENTION, which is the icon it refers to.

chapters/performance.adoc
tkrop: Sound like a good extension. I'm inclined to add this. Any objections.

chapters/performance.adoc
tkrop: I have opended #483 to discuss this outside of this PR.

chapters/common-headers.adoc
```diff
@@ -78,66 +69,73 @@ Content-Location: /products/123/images?format=raw
...
 ----
 
-Supporting APIs may return the `Preference-Applied` header also defined
-in https://tools.ietf.org/html/rfc7240[RFC7240] to indicate whether the
-preference was applied.
+Supporting APIs may return the {Preference-Applied} header also defined in
+{RFC7240}[RFC-7240] to indicate whether the preference was applied.
 
 [#182]
 == {MAY} Consider Using `ETag` Together With `If-Match`/`If-None-Match` Header
```
maxim-tschumak: The same argumentation as in the previous comment: let's make it `MAY support ETag`.

maxim-tschumak: I propose to change `use` to `support` in the whole guideline

chapters/common-headers.adoc
```diff
@@ -152,36 +150,39 @@ follows in the API definition:
...
-        The RFC7232 ETag header field in a response provides the current entity-
-        tag for the selected resource. An entity-tag is an opaque identifier for
-        different versions of a resource over time, regardless whether multiple
-        versions are valid at the same time. An entity-tag consists of an opaque
-        quoted string, possibly prefixed by a weakness indicator.
+        The RFC 7232 ETag header field in a response provides the entity-tag of
+        a selected resource. The entity-tag is an opaque identifier for versions
+        and representations of the same resource over time, regardless whether
+        multiple versions are valid at the same time. An entity-tag consists of
+        an opaque quoted string, possibly prefixed by a  weakness indicator (see
```
maxim-tschumak: [very minor] `a  weakness` -> `a weakness` (two whitespaces)

chapters/common-headers.adoc
```diff
@@ -152,36 +150,39 @@ follows in the API definition:
...
+        [RFC 7232 Section 2.3](https://tools.ietf.org/html/rfc7232#section-2.3).
 
       type: string
       required: false
-      example: W/"xy", "5", "7da7a728-f910-11e6-942a-68f728c1ba70"
+      example:  W/"xy", "5", "5db68c06-1a68-11e9-8341-68f728c1ba70"
```
maxim-tschumak: another extra whitespace

chapters/http-requests.adoc
```diff
@@ -226,6 +226,9 @@ An operation can be...
 * [[safe, safe]]{RFC-safe} - the operation semantic is defined to be read-only,
   meaning it must not have _intended side effects_, i.e. changes, to the server
   state.
+* [[cacheable, cacheable]]{RFC-cacheable} - to indicate that responses are
+  allowed to be stored for future reuse. In general, requests to safe methods
+  that do not depend on a current or authoritative response are cacheable.
```
maxim-tschumak: It's not quite clear for me, what is meant under `authoritative response`. Maybe rephrase it a bit?

chapters/http-requests.adoc
maxim-tschumak: You would also get a 404 if you are not authorized to delete a resource.

chapters/performance.adoc
```diff
@@ -171,22 +170,140 @@ GET /order/123?embed=(items) HTTP/1.1
...
-transparent HTTP caching should be avoided for RESTful APIs unless the
-API designer has proven to know better.
+As a consequence, client side as well as transparent web caching should be
+avoided, unless the service supports and requires it to protect itself, e.g.
+in case of a heavily used and therefore rate limited master data service, i.e.
+data items that rarely or not at all change after creation.
 
-As default, API providers should set the `Cache-Control: no-cache` header.
+As default, API providers and consumers should always set the {Cache-Control}
+header set to {Cache-Control-no-store} and assume the same setting, if no
```
maxim-tschumak: should be `Cache-Control: no-store`

chapters/performance.adoc
```diff
@@ -171,22 +170,140 @@ GET /order/123?embed=(items) HTTP/1.1
...
 
-If an API is intended to support caching, it must take care to specify
-this ability by defining the caching boundaries i.e. time-to-live and
-cache constraints, by providing the `Cache-Control` and `Vary` headers
-(Please read https://tools.ietf.org/html/rfc7234[RFC-7234] carefully).
+Caching has to take many aspects into account, e.g. general <<cacheable,
+cacheability>> of response information, our guideline to protect endpoints
+using SSL and <<104, OAuth authorization>>, resource update and invalidation
+rules, existence of multiple consumer instances. As a consequence, caching is
+in best case complex, in worst case inefficient.
```
maxim-tschumak: I think the biggest peril of using caching is the high probability of becoming **inconsistent** at some point. Should we mention it more explicitly?

chapters/performance.adoc
```diff
@@ -171,22 +170,140 @@ GET /order/123?embed=(items) HTTP/1.1
...
+        a request message, aside the target URL and HTTP method, might have
+        influenced the response. A client or proxy cache must respect this
+        information, to ensure that it delivers the correct cache entry (see
+        [RFC-7231 Section
+        7.1.4](https://tools.ietf.org/html/rfc7231#section-7.1.4)).
+
+      type: string
+      required: false
+      example: "accept-encoding, accept-language"
+  - ETag:
```
maxim-tschumak: `ETag` header is defined twice. I would prefer to have only one definition (here) and a reference.

chapters/performance.adoc
```diff
@@ -171,22 +170,140 @@ GET /order/123?embed=(items) HTTP/1.1
...
+different media types, or requires other request specific headers. To support
+correct caching you have to carefully choose the value. However, a good first
+default may be:
+
+[source,http]
+----
+Vary: accept, accept-encoding
+----
+
+Anyhow, this is only relevant, if you encourage clients to install generic
```
maxim-tschumak: It's not relevant for us (Zalando).

index.adoc
```diff
@@ -94,17 +106,40 @@
 
 // Attributes to improve design and linking of common headers.
 :ETag: pass:[<a href="https://tools.ietf.org/html/rfc7232#section-2.3"><code>ETag</code></a>]
+:entity-tag: pass:[<a href="https://tools.ietf.org/html/rfc7232#section-2.3"><code>&lt;entity-tag&gt;</code></a>]
 :If-Match: pass:[<a href="https://tools.ietf.org/html/rfc7232#section-3.1"><code>If-Match</code></a>]
 :If-None-Match: pass:[<a href="https://tools.ietf.org/html/rfc7232#section-3.2"><code>If-None-Match</code></a>]
+
+:Accept: pass:[<a href="https://tools.ietf.org/html/rfc7231#section-5.3.2"><code>Accept</code></a>]
```
maxim-tschumak: Does it work for PDFs? We should always look at all produced artifacts.

chapters/common-headers.adoc
```diff
@@ -78,66 +69,73 @@ Content-Location: /products/123/images?format=raw
...
 
-More details in RFC 7231
-https://tools.ietf.org/html/rfc7231#section-7.1.2[7.1.2 Location],
-https://tools.ietf.org/html/rfc7231#section-3.1.4.2[3.1.4.2
-Content-Location]
+More details in RFC 7231 {RFC-7231}#section-7.1.2[7.1.2 Location],
+{RFC-7231}#section-3.1.4.2[3.1.4.2 Content-Location]
 
 [#181]
 == {MAY} Use the Prefer header to indicate processing preferences
```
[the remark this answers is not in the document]
tkrop: Yes good suggestion. I look what I can do.

chapters/common-headers.adoc
tkrop: Lets restrict this to the part we are currently working on :)

chapters/http-requests.adoc
tkrop: Yes, that is not the easiest sentence: I must admit, I have stolen it from https://tools.ietf.org/html/rfc7231#section-4.2.3 :smile_cat:. What this means is, that you can cache responses for requests to *safe methods*, if you do not need a current answer or an answer from the original server. However, I'm not sure whether this makes the sentence, but I will try some improvement.

chapters/performance.adoc
tkrop: This is source code. I do not like to introduce *attributes* with blanks, but it will be translated as you suggested it.

chapters/performance.adoc
tkrop: In many cases even this is acceptable ...

Do you want to define an action for improvement on this remark?

chapters/performance.adoc
tkrop: Replaced against Link.

chapters/performance.adoc
tkrop: Added *inconsistency* problem more explicit.

index.adoc
tkrop: Works for me so far. Have we ever checked the general PDF style? We could improve it a lot.

chapters/http-requests.adoc
```diff
@@ -234,17 +238,17 @@ intended by the operation so that it can be held accountable.
 
 Method implementations must fulfill the following basic properties:
```
tfrauenstein: editorial: following basic properties:  --> following basic properties according to https://tools.ietf.org/html/rfc7231[HTTP 1.1 RFC]:

chapters/http-requests.adoc
tfrauenstein: fine -- we should not be less restrictive than HTTP 1.1. standard.

chapters/http-requests.adoc
```diff
@@ -234,17 +238,17 @@ intended by the operation so that it can be held accountable.
...
-| {POST}      | {NO}  | {AT} No, however you <<229>>.
-| {PUT}       | {NO}  | {YES}
-| {PATCH}     | {NO}  | {AT} No, however you <<229>>.
-| {DELETE}    | {NO}  | {YES}
-| {OPTIONS}   | {YES} | {YES}
-| {TRACE}     | {YES} | {YES}
+| Method    | Safe  | Idempotent | Cacheable
+| {GET}     | {YES} | {YES}      | {YES}, but <<227>>
+| {HEAD}    | {YES} | {YES}      | {YES}, but <<227>>
+| {POST}    | {NO}  | {AT} No, but <<229>> | {AT} May, but <<227>>
```
tfrauenstein: Yes, "May" according to HTTP standard. 
However, we also should provide following restrictive information here: 

{AT} May, but use cacheable only, if specific POST endpoint is safe (see <<227>>). 
Hint: overwhelming majority of cache implementations only support GET and HEAD.

tkrop:
:+1:

tfrauenstein:
👍

regispl:
Hey, 

The guidance you're giving sounds good to me in general, but there's a couple of things that 'd like to comment on:
1. I'm not sure I get what's the motivation for the detailed "small vs. medium vs. large data size" guidelines - would it make sense to focus on the "best possible" guidance, rather than providing this detailed explanation on case-to-case basis with only slight differences between them (e.g. guidance for small vs. large is almost identical)? So what I mean is putting together whatever is in the "In general (...)" section plus `HEAD` support and pagination (which, by the way, is already a "Must" according to other API Guideline's section) and considering it a "baseline"? 
This is even more confusing when you read the "general" section saying that "you should support ETag Together With If-Match / If-None-Match Header on all cacheable endpoints" but then `ETag`s are - for some reason - specifically mentioned one more time for "small", but omitted for "large". 
In general, while the advices are sound and correct, I find this section a bit too detailed and quite difficult to follow. 
2. Following up on the above - I find the "small vs. medium vs. large" distinction very "subjective" and open for interpretation and as such they don't seem to be useful for the readers. As mentioned above, I'd personally remove it completely (or at least make it way less detailed - one "default" best advice and one or two exceptions listed separately), but if it's to stay, would you consider providing a more detailed guidance on what those mean? 
3. The "Provide efficient methods to warm up and update caches (...)" part makes me think about the "thundering herd" problem on client's startup to me, which always scares me a bit as a service owner. I was wondering if it would make sense to **explicitly** mention the importance of Rate Limiting here (it's mentioned few paragraphs above in slightly different context)? 
4. Related to the point above: would it make sense to "promote" asynchronous eventing here as a possible alternative for using direct REST API calls & cache for accessing datasets (especially the large ones)? While no silver bullet and not the best option for every single case, when it can be used, the benefits of "notify on change" approach are massive - no periodic GETs, no thundering herds on startup to warm-up cache, no problem when cache expires and the service owning the data is down / slow, no cache "management" (respecting different headers, expiry etc.) at all...

Just a few ideas / doubts that came to my mind when reading this section 👍

PR #497, opened 2019-03-26
feat: add new section about durable immutable remote references (#487)

### LINKED ISSUE #487: Introduce Guideline for Usage of Durable Immutable External References
We currently discourage external references (e.g., `$ref: 'https://zalando.github.io/problem/schema.yaml#/Problem'`) in API specifications - except for the above example.

The reason for this is, that such references must be **durable** and **immutable** with respect to content, that is delivered from such a location. With the advance of our API repository, we are prepared to allow access to **published** API (fragments) via references.

tkrop:
fixes #487.

### DIFF
--- a/chapters/general-guidelines.adoc
+++ b/chapters/general-guidelines.adoc
@@ -22,9 +22,10 @@ all apis with `x-api-audience =/= component-internal` (see <<219, API Audience>>
 == {MUST} Provide API Specification using OpenAPI
 
 We use the http://swagger.io/specification/[OpenAPI specification] as standard
-to define API specifications files. API designers have to provide the API
-specification files using *YAML* to improve readability. We encourage to use
-*OpenAPI 3.0* version, but still support *OpenAPI 2.0* (a.k.a. Swagger 2).
+to define API specification files. API designers are required to provide the API
+specification using a single *self-contained YAML* file to improve readability.
+We encourage to use *OpenAPI 3.0* version, but still support *OpenAPI 2.0*
+(a.k.a. Swagger 2).
 
 The API specification files should be subject to version control using a source
 code management system - best together with the implementing sources. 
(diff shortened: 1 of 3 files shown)


tkrop:
I must admit: Adding this aspect to the general section suggests to delete the `API operation` section and move its content to the general section to have this main aspects in one place.

chapters/general-guidelines.adoc
```diff
@@ -42,23 +43,46 @@ existing APIs the https://editor.swagger.io/[Swagger Editor] or our
...
 
 
+[#234]
+== {MAY} use Durable and Immutable Remote References
+
+Normally, API specification files must be *self-contained*, i.e. files
+should not contain references to local or remote content, e.g
+`$ref: 'https://github.com/zalando/zally/blob/master/server/src/main/resources/api/zally-api.yaml'`.
+The reason is, that the content referred to is _in general_ *not durable* and
+*immutable*. As a consequence, the semantic of an API may change in unexpected
```
maxim-tschumak: and *not immutable*

chapters/general-guidelines.adoc
```diff
@@ -42,23 +43,46 @@ existing APIs the https://editor.swagger.io/[Swagger Editor] or our
...
+should not contain references to local or remote content, e.g
+`$ref: 'https://github.com/zalando/zally/blob/master/server/src/main/resources/api/zally-api.yaml'`.
+The reason is, that the content referred to is _in general_ *not durable* and
+*not immutable*. As a consequence, the semantic of an API may change in
+unexpected ways.
+
+However, you may use remote references to resources accessible by the following
+service URLs.
+
+* `https://infrastructure-api-repository.zalandoapis.com/`
```
whiskeysierra: Is this backed by a corresponding Zally rule?

maxim-tschumak: Not yet, but we will follow up on this. (every guidelines' merge to master creates an issue in Zally repository to trigger a review and a potential change/new check)

tkrop: Up to now, we only check for
 
* `https://zalando.github.io/problem/schema.yaml#/Problem` and
* `https://opensource.zalando.com/problem/schema.yaml#/Problem`.

Makes me wonder whether I have referenced the right one, or whether I should enumerate both.

tkrop:
May be I should rename the rule to `MUST only use Durable and Immutable Remote Reference`.

tfrauenstein:
👍

maxim-tschumak:
:+1:

PR #501, opened 2019-05-30
feat: add HEAD for response status code 304 (#501)

tkrop:
fixes #501. Adding `HEAD` for status code `304` as mentioned in [RFC 7232 Section 4.1](https://tools.ietf.org/html/rfc7232#section-4.1).


zeitlinger:
:+1:

tkrop:
:+1:

PR #503, opened 2019-06-05
feat: improve prefer header specification (#503)

tkrop:
fixes #502. Fixed spelling. Added note to copy only supported behaviors after reviewing an API that copied the full template and adding a separate description to restrict the supported use cases instead of striping down the template.

In addition, I have put some broken link and visual fixes into this pull request, that we stumbled over during the review of @maxim-tschumak.

### DIFF
--- a/chapters/common-headers.adoc
+++ b/chapters/common-headers.adoc
@@ -95,7 +95,7 @@ The {Prefer} header can defined like this in an API definition:
 components:
   headers:
   - Prefer:
-      description: |
+      description: >
         The RFC7240 Prefer header indicates that a particular server behavior
         is preferred by the client but is not required for successful completion
         of the request (see [RFC 7240](https://tools.ietf.org/html/rfc7240).
@@ -110,16 +110,25 @@ components:
           resource (representation) in the response body on success.
         * **wait=<delta-seconds>** is used to suggest a maximum time the server
           has time to process the request synchronously.
-        * **handling=<strinct|lenient>** is used to suggest the server to be
+        * **handling=<strict|lenient>** is used to suggest the server to be
           strict and report error conditions or lenient, i.e. robust and try to
           continue, if possible.
 
-      type: string
+      type: array
+      style: form
+      items:
+        type: string
       required: false
 ----
 
+*Note:* Please copy only the behaviors into your {Prefer} header specification
+that are supported by your API endpoint. If necessary, specify different
+{Prefer} headers for each supported use case.
+
 Supporting APIs may return the {Preference-Applied} header also defined in
 {RFC7240}[RFC 7240] to indicate whether a preference has been applied.
+
+
 
 [#182]
 == {MAY} Consider to Support `ETag` Together With `If-Match`/`If-None-Match` Header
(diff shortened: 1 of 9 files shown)


chapters/common-headers.adoc
```diff
@@ -110,17 +110,26 @@ components:
...
         * **wait=<delta-seconds>** is used to suggest a maximum time the server
           has time to process the request synchronously.
-        * **handling=<strinct|lenient>** is used to suggest the server to be
+        * **handling=<strict|lenient>** is used to suggest the server to be
           strict and report error conditions or lenient, i.e. robust and try to
           continue, if possible.
 
-      type: string
+      type: array
+      style: form
```
tfrauenstein: We cannot change to array type here:
a) it is not consistent with RFC which is not a json array format (though multiple prefer aspects may be defined here)
b) it is an incompatible change

tkrop: @tfrauenstein Thanks for the comment. The pull request is actually supposed to make it compliant with the [RFC 7240 Section 2](https://tools.ietf.org/html/rfc7240#section-2). The standard defines:

> A client MAY use multiple instances of the Prefer header field in a single message, or it MAY use a single Prefer header field with multiple comma-separated preference tokens.  If multiple Prefer header fields are used, it is equivalent to a single Prefer header field with the comma-separated concatenation of all of the tokens.

which translates for headers `type: array` with `style: simple` and `explode: true|false` (see [Style Values](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#style-values)) which in OpenAPI is not defining a json array, but a comma separated list of values (see [Style Examples](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#style-examples)).

I must admit I did the `style: form` wrong in the first commit, it must be `style: simple` which is the default for headers. I will change it accordingly and omit the `style` definition.

To give context: @ePaul stumbled across this in the guild meeting when we talked about a reviewed API. His team is defining it slightly different using an `x-extensible-enum` list of acceptable values. However, I think this is redundant to the description. As the description cannot be removed without loss of information, I skipped the `x-extensible-enum` values for the guideline. Besides it would be incomplete, as `wait=<seconds>` not that straight forward to enumerate.

chapters/common-headers.adoc
```diff
@@ -110,17 +110,26 @@ components:
...
+      items:
+        type: string
       required: false
 ----
 
+*Note:* Please copy only the behaviors into your {Prefer} header specification
+that are supported by your API endpoint. If necessary, specify different
+{Prefer} headers for each supported use case.
+
```
maxim-tschumak: The links to RFC is missing, please add one.

tkrop: @maxim-tschumak I have already linked twice in this Section to the RFC 7240 - once at the begin of it and once one line below -, and the `Prefer` keyword itself is also clickable and leads to the RFC 7240. Do you really think need more linking?

chapters/common-headers.adoc
tkrop: Updated.

chapters/common-headers.adoc
maxim-tschumak: See line 128 - the link is not valid, I would just fix it

maxim-tschumak:
:+1:

tkrop:
:+1:

PR #508, opened 2019-06-18
feat: correct rule 154 (#505)

### LINKED ISSUE #505: Rule #154 is incorrect for headers
Rule [#154 - Should: Define Collection Format of Query Parameters and Headers](https://opensource.zalando.com/restful-api-guidelines/#154) is incorrect for headers. [Open API 3.0](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#style-values) does not support `style: form` for headers. The equivalent specification to query parameters `style: form` for headers is `style: simple`.

In addition, I like to challenge whether we should enforce collection formats by making this a `MUST` rule instead of `SHOULD`.

tkrop:
fixes #505.

In addition, this pull request suggests that users explicitly *must* define the collection format. The OpenAPI specification 2.0 was dropped, to foster migration to 3.0.x after supporting it now for nearly a year as default.

### DIFF
--- a/chapters/http-requests.adoc
+++ b/chapters/http-requests.adoc
@@ -331,19 +331,36 @@ original one implementing a well defined behavior. Using {204} without content
 would be a similar well defined option.
 
 [#154]
-== {SHOULD} Define Collection Format of Query Parameters and Headers
+== {MUST} Define Collection Format of Query Parameters and Headers
 
-Sometimes, query parameters and headers allow to provide a list of values,
-either by providing a comma-separated list (`csv`) or by repeating the
-parameter multiple times with different values (`multi`). The API
-specification should explicitly define one type as follows:
+Header, path, and query parameters  allow to provide a collection of values,
+either by providing a comma-separated list of values (`csv`) or by repeating
+the parameter multiple times with different values (`multi`) as follows:
 
-[,cols="20%,20%,25%,35%",options="header",]
-|=======================================================================
-|Description |OpenAPI 3.0 |OpenAPI 2.0 |Example
-|Comma separated values |`style: form, explode: false` |`collectionFormat: csv` |`?param=value1,value2`
-|Multiple parameters |`style: form, explode: true` |`collectionFormat: multi` |`?param=value1&param=value2`
-|=======================================================================
+[,cols="14%,30%,39%,17%",options="header",]
+|=========================================================================
+| Parameter Type | Comma-separated Values | Multiple Parameters | Standard
+| Header | `Header: value1,value2` | `Header: value1, Header: value2`
+| {RFC-7230}#section-3.2.2[RFC 7230 Section 3.2.2]
+
+| Path | `key=value1,value2` | `/key=value1;key=value2`
+| {RFC-6570}#section-3.2.7[RFC 6570 Section 3.2.7]
+
+| Query | `?param=value1,value2` | `?param=value1&param=value2`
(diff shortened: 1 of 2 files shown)


chapters/http-requests.adoc
```diff
@@ -331,19 +331,36 @@ original one implementing a well defined behavior. Using {204} without content
...
+either by providing a comma-separated list of values (`csv`) or by repeating
+the parameter multiple times with different values (`multi`) as follows:
+
+[,cols="14%,30%,39%,17%",options="header",]
+|=========================================================================
+| Parameter Type | Comma-separated Values | Multiple Parameters | Standard
+| Header | `Header: value1,value2` | `Header: value1, Header: value2`
+| {RFC-7230}#section-3.2.2[RFC 7230 Section 3.2.2]
+
+| Path | `key=value1,value2` | `/key=value1;key=value2`
```
maxim-tschumak: The rule title don't say anything about path parameters - we should either add path params to the title or remove those from the table

chapters/http-requests.adoc
```diff
@@ -331,19 +331,36 @@ original one implementing a well defined behavior. Using {204} without content
...
-[,cols="20%,20%,25%,35%",options="header",]
-|=======================================================================
-|Description |OpenAPI 3.0 |OpenAPI 2.0 |Example
-|Comma separated values |`style: form, explode: false` |`collectionFormat: csv` |`?param=value1,value2`
-|Multiple parameters |`style: form, explode: true` |`collectionFormat: multi` |`?param=value1&param=value2`
-|=======================================================================
+== {MUST} Define Collection Format of Query Parameters and Headers
+
+Header, path, and query parameters  allow to provide a collection of values,
+either by providing a comma-separated list of values (`csv`) or by repeating
```
maxim-tschumak: `csv`, `multi` - this is Swagger (Open API 2) terminology - I would just remove it.

chapters/http-requests.adoc
```diff
@@ -331,19 +331,36 @@ original one implementing a well defined behavior. Using {204} without content
...
+either by providing a comma-separated list of values or by repeating the
+parameter multiple times with different values as follows:
+
+[,cols="14%,30%,39%,17%",options="header",]
+|=========================================================================
+| Parameter Type | Comma-separated Values | Multiple Parameters | Standard
+| Header | `Header: value1,value2` | `Header: value1, Header: value2`
+| {RFC-7230}#section-3.2.2[RFC 7230 Section 3.2.2]
+
+| Path | `key=value1,value2` | `/key=value1;key=value2`
```
ePaul: I've never found a reason to use a collection path parameter (normally I would use a query parameter instead).

Also, normally path parameters in OpenAPI don't have a key-value syntax – are we introducing something new here?

tkrop: I agree. This is a very rare case. I just added this line for completeness. Should I remove it?

maxim-tschumak:
:+1:

chapters/http-requests.adoc
ePaul: Yes, as long as we don't have anywhere else those key-value-parameters in a path, I would omit them here.

whiskeysierra:
:+1:

vadeg:
:

zeitlinger:
:+1:

tkrop:
:+1:

zeitlinger:
:+1:

Issue #161, opened 2016-12-09
Consider use of RESTful term in the guidelines


dehora:
The guidelines lead with term RESTful, and while there are some use of the constraints, use of hypermedia in particular is limited (mostly to drive pagination). 

They guidelines don't argue against REST, but nor do they advocate strongly for example, for content models that use a state machine driven approach, where transitions are performed by clients working with document links. 

It's arguable then, the term could be removed from the leading material and point to the discussion elsewhere in the document on tradeoffs.

If done, this might require a new github project as "restful" is in the project url.

hjacobs:
Linking the trigger of this: https://twitter.com/stilkov/status/806979435429773312 :smirk:

hjacobs:
Still waiting for an answer to https://twitter.com/ebowman/status/807154435755536384 ...

mohabusama:
The reply: https://twitter.com/stilkov/status/807219063185408000

Github API,, of course :)

whiskeysierra:
> Github API,, of course :)

Totally disagree. They just provide links for reading. There is not state changed initiated through links.

tfrauenstein:
I personally think this discussion is rather academic and provides no value for us. However, I propose to complement our guidelines with the footnote that we stick to usage of the term for our REST approximations (like the very industry majority) --> see #169

PR #510, opened 2019-06-24
feat: improved cursor guideline for get-with-body (#504)

### LINKED ISSUE #504: Add guidance on overlong cursors requiring `GET with body`
Because a *search query* and the subsequent `cursor` can be quite extensive, it may be good to give advice how to handle this cases consistently. The idea is to implement a body based search query containing a cursor as follows:

```
{
  "cursor": "<encode-search-query-with-resource-pointer-and-direction>"
  "limit": <numer>
  "query-params": ... // ignored, if cursor is provided
}
```
The response contains the encode [`first`?, `next`, `prev`, and `last`? cursors] as usual without making use of [pagination links](https://opensource.zalando.com/restful-api-guidelines/#161).

We should also add `cursor`, `first`, `next`, `prev`, and `last` to [Use common field names and semantics](https://opensource.zalando.com/restful-api-guidelines/#174) with a reference to pagination.

tkrop:
fixes #504.

I have added a *pagination* section to the common fields, and gave examples how the *pagination requests* and *responses* should look like.

@whiskeysierra @meshcalero @ePaul please have a look.

### DIFF
--- a/chapters/common-data-types.adoc
+++ b/chapters/common-data-types.adoc
@@ -126,23 +126,23 @@ allows API client developers to build up a common understanding of Zalando's
 resources. There is very little utility for API consumers in having different
 names or value types for these fields across APIs.
 
-[[pagination-fields]]
-=== Pagination Fields
+[[simple-hypertext-control-fields]]
+=== Simple Hypertext Control and Pagination Fields
 
-To foster a consistent look and feel response object for paginating over
-collections should follow a common pattern using the following pagination
-field semantics:
+To foster a consistent look and feel of simple hypertext controls responses
+for paginating over collection values should follow a common pattern using the
+following field semantics:
 
-* [[first]]{first}: the cursor or link in a pagination response pointing to
-  the first page.
-* [[prev]]{prev}: the cursor or link in a pagination response pointing to the
-  previous page.
-* [[self]]{self}: the cursor or link in a pagination response pointing to the
-  current page.
-* [[next]]{next}: the cursor or link in a pagination response pointing to the
-  next page.
-* [[last]]{last}: the cursor or link in a pagination response pointing to the
-  last page.
+* [[first]]{first}: the link or cursor in an object or pagination response
+  pointing to the first object or page.
+* [[prev]]{prev}: the link or cursor in an object or pagination response
+  pointing to the previous object or page.
+* [[self]]{self}: the link or cursor in an object or pagination response
+  pointing to the same object or the current page.
+* [[next]]{next}: the link or cursor in an object or pagination response
+  pointing to the next object or page.
+* [[last]]{last}: the link or cursor in an object or pagination response
+  pointing to the last page.
 
 To distinguish between pagination links and cursors the cursor set should be
 wrapped in a {cursors} object as follows:
@@ -154,6 +154,7 @@ wrapped in a {cursors} object as follows:
 * [[items]]{items}: array of resources, holding all the items of the current
   page (can be replaced by resource name).
 
+For a standard response page this may look as follows:
 
 [source,yaml]
 ----
@@ -204,8 +205,9 @@ ResponsePage:
        items: ...
 ----
 
-The response page may contain additional metadata about the collection or the
-current page, if necessary.
+Normally, only cursors or simple hypertext controls are used. The response page
+may contain additional metadata about the collection or the current page, if
+necessary.
 
 [[address-fields]]
 === Address Fields
--- a/chapters/common-data-types.adoc
+++ b/chapters/common-data-types.adoc
@@ -91,7 +91,7 @@ provided that otherwise requires examining the Open API file.
 
 Example JSON schema:
 
-[source,json]
+[source,yaml]
 ----
 tree_node:
   type: object
@@ -121,11 +121,91 @@ tree_node:
     parent_node_id: '534321'
 ----
 
-These properties are not always strictly necessary, but making them
-idiomatic allows API client developers to build up a common
-understanding of Zalando's resources. There is very little utility for
-API consumers in having different names or value types for these fields
-across APIs.
+These properties are not always strictly necessary, but making them idiomatic
+allows API client developers to build up a common understanding of Zalando's
+resources. There is very little utility for API consumers in having different
+names or value types for these fields across APIs.
+
+[[pagination-fields]]
+=== Pagination Fields
+
+To foster a consistent look and feel response object for paginating over
+collections should follow a common pattern using the following pagination
+field semantics:
+
+* [[first]]{first}: the cursor or link in a pagination response pointing to
+  the first page.
+* [[prev]]{prev}: the cursor or link in a pagination response pointing to the
+  previous page.
+* [[self]]{self}: the cursor or link in a pagination response pointing to the
+  current page.
+* [[next]]{next}: the cursor or link in a pagination response pointing to the
+  next page.
+* [[last]]{last}: the cursor or link in a pagination response pointing to the
+  last page.
+
+To distinguish between pagination links and cursors the cursor set should be
+wrapped in a {cursors} object as follows:
+
+* [[cursors]]{cursors}: pagination object containing the cursors for the
+  {first}, {prev}, {self}, {next}, and {last} page.
+* [[cursor]]{cursor}: the cursor used in a {GET-with-body} search request
+  pointing to the requested page.
+* [[items]]{items}: array of resources, holding all the items of the current
+  page (can be replaced by resource name).
+
+
+[source,yaml]
+----
+ResponsePage:
+  type: object
+  properties: 
+    first:
+      description: pagination link pointing to the first page.
+      type: string
+      format: uri
+    prev:
+      description: pagination link pointing to the previous page.
+      type: string
+      format: uri
+    self:
+      description: pagination link pointing to the current page.
+      type: string
+      format: uri
+    next:
+      description: pagination link pointing to the next page.
+      type: string
+      format: uri
+    last:
+      description: pagination link pointing to the last page.
+      type: string
+      format: uri
+
+    cursors:
+      first:
+        description: cursor pointing to the first page.
+        type: string
+      prev:
+        description: cursor pointing to the previous page.
+        type: string
+      self:
+        description: cursor pointing to the current page.
+        type: string
+      next:
+        description: cursor pointing to the next page.
+        type: string
+      last:
+        description: cursor pointing to the last page.
+        type: string
+
+     items:
+       description: array of collection items
+       type: array
+       items: ...
+----
+
+The response page may contain additional metadata about the collection or the
+current page, if necessary.
 
 [[address-fields]]
 === Address Fields
--- a/chapters/common-data-types.adoc
+++ b/chapters/common-data-types.adoc
@@ -126,88 +126,82 @@ allows API client developers to build up a common understanding of Zalando's
 resources. There is very little utility for API consumers in having different
 names or value types for these fields across APIs.
 
-[[simple-hypertext-control-fields]]
-=== Simple Hypertext Control and Pagination Fields
+[[link-relation-fields]]
+=== Link Relation Fields
 
-To foster a consistent look and feel of simple hypertext controls responses
-for paginating over collection values should follow a common pattern using the
-following field semantics:
+To foster a consistent look and feel using simple hypertext controls for
+paginating and iterating over collection values the response objects should
+follow a common pattern using the below field semantics:
 
-* [[first]]{first}: the link or cursor in an object or pagination response
-  pointing to the first object or page.
-* [[prev]]{prev}: the link or cursor in an object or pagination response
-  pointing to the previous object or page.
-* [[self]]{self}: the link or cursor in an object or pagination response
-  pointing to the same object or the current page.
-* [[next]]{next}: the link or cursor in an object or pagination response
-  pointing to the next object or page.
-* [[last]]{last}: the link or cursor in an object or pagination response
-  pointing to the last page.
+* [[self]]{self}:the link or cursor in a pagination response or object
+  pointing to the same collection object or page.
+* [[first]]{first}: the link or cursor in a pagination response or object
+  pointing to the first collection object or page.
+* [[prev]]{prev}: the link or cursor in a pagination response or object
+  pointing to the previous collection object or page.
+* [[next]]{next}: the link or cursor in a pagination response or object
+  pointing to the next collection object or page.
+* [[last]]{last}: the link or cursor in a pagination response or object
+  pointing to the last collection object or page.
 
-To distinguish between pagination links and cursors the cursor set should be
-wrapped in a {cursors} object as follows:
+Pagination responses should contain the following additional array field to
+transport the page content:
 
-* [[cursors]]{cursors}: pagination object containing the cursors for the
-  {first}, {prev}, {self}, {next}, and {last} page.
-* [[cursor]]{cursor}: the cursor used in a {GET-with-body} search request
-  pointing to the requested page.
 * [[items]]{items}: array of resources, holding all the items of the current
-  page (can be replaced by resource name).
+  page ({items} may be replaced by a resource name).
 
-For a standard response page this may look as follows:
+To simplify user experience, the applied query filters may be returned using
+the following field (see also {GET-with-body}):
+
+* [[query]]{query}: object containing the query filters applied in the search
+  request to filter the collection resource.
+
+As Result, the standard response page using <<161, pagination links>> is
+defined as follows:
 
 [source,yaml]
 ----
 ResponsePage:
   type: object
   properties: 
+    self:
+      description: Pagination link pointing to the current page.
+      type: string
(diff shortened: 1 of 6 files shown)


chapters/common-data-types.adoc
```diff
@@ -121,11 +121,91 @@ tree_node:
...
+
+To foster a consistent look and feel response object for paginating over
+collections should follow a common pattern using the following pagination
+field semantics:
+
+* [[first]]{first}: the cursor or link in a pagination response pointing to
+  the first page.
+* [[prev]]{prev}: the cursor or link in a pagination response pointing to the
+  previous page.
+* [[self]]{self}: the cursor or link in a pagination response pointing to the
```
whiskeysierra: Is that useful?

chapters/common-data-types.adoc
```diff
@@ -121,11 +121,91 @@ tree_node:
...
+  current page.
+* [[next]]{next}: the cursor or link in a pagination response pointing to the
+  next page.
+* [[last]]{last}: the cursor or link in a pagination response pointing to the
+  last page.
+
+To distinguish between pagination links and cursors the cursor set should be
+wrapped in a {cursors} object as follows:
+
+* [[cursors]]{cursors}: pagination object containing the cursors for the
```
whiskeysierra: Doesn't https://opensource.zalando.com/restful-api-guidelines/#164 dictate that we should put a `cursor` property in the `next`/`prev`/... properties next to their respective `href` property?

Example:

```json
{
  "prev": {
    "href": "https://..",
    "cursor": ".."
  },
  "next": {
    "href": "https://..",
    "cursor": ".."
  }
}
```

CC @meshcalero

chapters/common-data-types.adoc
```diff
@@ -121,11 +121,91 @@ tree_node:
...
+    next:
+      description: pagination link pointing to the next page.
+      type: string
+      format: uri
+    last:
+      description: pagination link pointing to the last page.
+      type: string
+      format: uri
+
+    cursors:
```
whiskeysierra: For me, the need to declare cursors separate from the links shows that we didn't follow our own rules when it comes to extensions. In particular:

- Must: Use Common Hypertext Controls [164] https://opensource.zalando.com/restful-api-guidelines/#164
- Must: Always Return JSON Objects As Top-Level Data Structures To Support Extensibility [110] https://opensource.zalando.com/restful-api-guidelines/#110 (kind of)

chapters/hyper-media.adoc
```diff
@@ -152,10 +152,10 @@ understandability and usability of an API.
 Hypertext controls for pagination inside collections and self-references
 should use a simple URI value in combination with their corresponding
 http://www.iana.org/assignments/link-relations/link-relations.xml[link
-relations] (`next`, `prev`, `first`, `last`, `self`) instead of the
-extensible common hypertext control
+relations] ({next}, {prev}, {first}, {last}, and {self}) instead of the
```
whiskeysierra: Why are we not explaining the reasoning behind this here?

chapters/pagination.adoc
```diff
@@ -32,92 +31,106 @@ you must stick to the common query parameter names defined in <<137>>.
...
+     the database.
+  ** Sharded or NoSQL databases.
+* Cursor-based navigation may not work if you need the total count of results.
+
+*Hint:* The cursor used for pagination is an opaque pointer to a page, that
+must never be inspected/constructed by clients. It usually (encrypted) encodes
+the identifier of the first or last page element, the pagination direction, and
+the applied query filters to recreate the collection. 
+
+The page information for cursor-based pagination should consist of a set of
```
whiskeysierra: Do we advocate this for all cases or just *GET with body*?

chapters/common-data-types.adoc
tkrop: Hmm, good question. Especially since we use it in three different contexts with slightly different semantic:

1. As `self`-reference of a resource, which is a broader use case than addressed here,
2. As `self`-reference of a page resource in pagination, which is described here, but is only a restricted version of the before, and
3. As pure cursor pointing without link relation context, but still pointing to a page resource in pagination.

@whiskeysierra I'm wondering, whether we could and should generalize the definition of all link relations / cursors here?

chapters/common-data-types.adoc
tkrop: @whiskeysierra https://opensource.zalando.com/restful-api-guidelines/#165 is specifying that for pagination simplified hypertext controls should be used.

whiskeysierra: It's not described why that was chosen and now it bites us.

chapters/common-data-types.adoc
tkrop: I can't see the contradiction here. The intend of this element is to clearly distinguish between the simple hypertext control usage and the cursor-only usage, that is getting necessary because cursors can get to long for hypertext controls.

whiskeysierra: > The intend of this element is to clearly distinguish between the simple hypertext control usage and the cursor-only usage

Why is it beneficial to keep those two hard separated? It doesn't really matter whether it's a pagination that uses `GET` (+ `href`) or `GET with body`(+ `href` and `cursor`). In both cases what we want to express is the ability to navigate to another page. Adding the cursor as an additional property to the existing page links would be an obvious extension that requires minimal change.

The fact that we would now end up with two places where we use the same names (*next*, *prev*, etc) is already a smell.

chapters/hyper-media.adoc
tkrop: Good question. As you can see, I added the `.` to the sentence. May be @meshcalero originally planned to give more explanation here but forgot. On the other side, generally simplifying the hypertext controls for use cases where all other control keywords are clear, makes sense for me without further explanation too.

I think the point is, that we still missing the ful set of hypertext controls that @meshcalero has presented in his talk in the guideline.

chapters/pagination.adoc
tkrop: Yes, in some way this is intended. My aim was to keep the original structure to define rule `#160` for general cursor-based pagination giving now the precise example, and than extend the guideline by rule `#161` to suggest using link relations for pagination where possible.

However, I must agree this may be misleading. I add a reference to the next rule here and state that the cursors may be dropped in favor of link relations.

chapters/common-data-types.adoc
tkrop: So in effect you suggest to use the [common hypertext controls](https://opensource.zalando.com/restful-api-guidelines/#164) instead of the [simplified hypertext controlls](https://opensource.zalando.com/restful-api-guidelines/#165) for cursors?

@meshcalero what do you think?

whiskeysierra: > So in effect you suggest to use the [common hypertext controls](https://opensource.zalando.com/restful-api-guidelines/#164) instead of the [simplified hypertext controlls](https://opensource.zalando.com/restful-api-guidelines/#165) for cursors?

Yes! I haven't seen a good reasoning for the simplified version actually.

chapters/common-data-types.adoc
tkrop: Changing a [simplified hypertext controlls](https://opensource.zalando.com/restful-api-guidelines/#165) into a [common hypertext controls](https://opensource.zalando.com/restful-api-guidelines/#164) and vice versa is a breaking change. So one has to decide what to support for each object right from the start.

For pagination I am a fan of the simple version, even if this means to be inconsistent in the use of hypertext controls.

whiskeysierra: > So one has to decide what to support for each object right from the start.

If we would have gone with the common controls right from the start we wouldn't have a problem. I'm curious why we did that since we usually advocate for extensibility.

whiskeysierra: > For pagination I am a fan of the simple version

Why?

chapters/common-data-types.adoc
tkrop: I gave generalization a try here.

chapters/hyper-media.adoc
tkrop: I tried to add more motivation here.

chapters/pagination.adoc
tkrop: Added a sentence to the end.

chapters/common-data-types.adoc
tkrop: Because it reduces the cognitive load and specification overhead - added this to [simplified hypertext controlls](https://opensource.zalando.com/restful-api-guidelines/#165).

chapters/common-data-types.adoc
meshcalero: @whiskeysierra : The motivation for the simplified version is avoid noise: API are not just for machines but also for humans (developers seeing a API response want to get to the essence and not have it buried somewhere in syntactic sugar). But the reason why the rule is only a **should** and not a **must** rule is that there exist valid use cases where you would do differently; and this example is one of them. We might even add this for clarity to the "simplified version" rule.

I also believe that having the cursor embedded in the controls is the cleanest solution, but there is one detail that I believe is worth to consider: Response size. If implemented efficiently, the cursor for all pagination controls can be the same (by encoding the offset etc. in the URI and only keep the initial complex filter in the "cursor"). Given that the cursor is huge (otherwise you would not need a POST) this blows up the response size. GZIP will take care of most of the impact for the transport, but JSON parsing time and memory consumption is still there (and can be measurable on mobile devices). If we want to avoid that, then you would prefer to have a single cursor attribute for all pagination controls, which would then reside outside of the individual controls. And then you could keep the simplified version for the individual controls too. On the other side: Putting this into a rule would enforce a specific implementation for cursors (make the cursor identically for all pagination controls; but also here might exit valid reasons not to do it) which I would like to avoid.

Might be a premature optimization, but I still believe worth to consider. @tkrop @whiskeysierra What do you think?

chapters/common-data-types.adoc
dehora: >> Why?
>
> Because it reduces the cognitive load and specification overhead - added this to simplified hypertext controlls.

It does neither. The guidelines have two ways to declare links, where one is dedicated to pagination via well known property names which I suspect is by far the most common of use of links in APIs following this guideline, and another, via structure which I suspect is less used. That increases load and overhead—"here's how to do links, except for the most common case". There should be one way to add links for any purpose and that would have to be @meshcalero's variant because it generalises.

chapters/common-data-types.adoc
tkrop: To answer @meshcalero question of *premature optimization* ...

I think the number of use cases for this optimization is close to 100% of the `GET with body` use cases. When I was writing my proposal I actually was also concerned about the explosion in response data that could be inflicted by prolonging the cursors. On the other hand, I have reviewed a number of APIs in the past that artificially restricted the number of parameters to a small set sacrificing the potential batch request efficiency to limit the `cursor` size.

In the same context @ePaul yesterday challenged the encoding of `query` parameters within the `cursor`, which I yesterday defended because I think there is value in encoding query parameters in a way that makes consequent requests fault tolerant and easy to handle.

@whiskeysierra @dehora @meshcalero I'm open for a change here, if we can agree on a common format.

So @whiskeysierra and @dehora are opting for a common definition for `GET` and `GET with body` discarding the `simple hypertext controls:

```
{
  "prev": {
    "href": "https://...",
    "cursor": "..."
  },
  "next": {
    "href": "https://...",
    "cursor": "..."
  }
}
```

@ePaul yesterday suggested a slightly more complex variant challenging the proposal to encode the `query` within the `cursor` as follows:
```
{
  "prev": {
    "href": "https://...",
    "body": {
      "cursor": "...",
      "query": "..."
    }
  },
  "next": {
    "href": "https://...",
    "body": {
      "cursor": "...",
      "query": "..."
    }
  }
}
```

@meshcalero is considering to join `GET` and `GET with body` with a modified simple hypertext control to shrink the response size by reusing the `body.query` while encoding only the `pointer` and the `direction` in the `cursor` that may look as follows:
```
{
  "prev": "https://...?cursor=...",
  "next": "https://...?cursor=...",
  "body": {
     "query": "..."
  }
}

@whiskeysierra, @dehora, @ePaul, and @meshcalero: Other options? Which variant should we go for?

File: chapters/common-data-types.adoc (hunk 3 above)
Comment (tkrop) [reply chapters/common-data-types.adoc]: @dehora I'm generally for consistency, but also for efficiency. And - I consider the simple hypertext controls as a consistent efficient alternative, when the other parameters can be simply derived from the context.

However, if all agree that simple hypertext controls should be discouraged, I follow this approach and rewrite the rules accordingly. @whiskeysierra, @meshcalero, @ePaul and other: please vote here for discouraging!

**Comment by whiskeysierra** [comment]
I actually like all three of these options. My primary concern was the
duplication of the link relation properties. All three address that.

File: chapters/common-data-types.adoc (hunk 3 above)
Comment (meshcalero) [reply chapters/common-data-types.adoc]: I would not mind to discourage the use of simple hypertext controls, but I recall that in one of my previous project we did introduce them based on user-feedback on our initial API draft where we had only hypertext control objects, as many users did not like that 50+% of all hypertext controls then did look like

`"self" : {
    "href": "https://..../
}`

instead of just 

`"self" : "https://.../"`

Esp. the `self` and pagination controls were heavily criticized. But also others had the same problem. That's how we back then came up with the rule, that any control with a [IANA registered link-rel-type](http://www.iana.org/assignments/link-relations/link-relations.xhtml) as name is allowed to use the simplified representation.

Back then I was the originally the one fighting for consistency, but I had to learn that users don't care about this consistency as they anyway don't scan payload in a generic way for controls. As long as a given content type always uses the same flavor for a specific control, users preferred the shorter version. 

So in the end user feedback convinced me that conciseness and readability of the payload is sometimes more important than consistency, esp. for public api, where the first impression is critical for developers even considering to use your API. This might not be the biggest concern for Zalando's API, but on the other side it feels a bit odd to me that we give up something where I have learned that it can have quite some value for users, for something like pagination within GET with body, which is a feature that anyway has a smell, even when I accept that it has valid use cases.

But as I still like consistency a lot, I would not veto against discouraging the simple version either.

Comment (meshcalero) [reply chapters/common-data-types.adoc]: @whiskeysierra I'm not sure the simple representation is introducing a problem with respect to extensibility: If you start with the simplified version, then also the end point will be a GET. If you later extend this to a GET with body, you will not break any exisiting client of the GET endpoint, as you would still serve here the simple version and only in the GET with body version you would use the object representation.

Comment (whiskeysierra) [reply chapters/common-data-types.adoc]: That's technically correct (the best kind of correct!), I didn't think about it like that. I'd probably argue that's it's still inconsistent and inconvenient for clients that need to switch. But the argument is less strong.

Comment (tkrop) [reply chapters/common-data-types.adoc]: @meshcalero thanks for giving this nice chain of experiences that match my intuition on user feelings.

File: chapters/common-data-types.adoc (hunk 2 above)
Comment (zeitlinger) [reply chapters/common-data-types.adoc]: @tkrop thanks for the great (offline) summary.

I like the last suggestion and would rename `query` to `filter`.

A complete example (which helped me understand the topic) could look as follows

```
{
  "prev": "https://my.host/users?cursor=<prevCur>"
  "next": "https://my.host/users?cursor=<nextCur>"
  "body": {
    "filter": {
      "yearJoined": 2019,
      "active": true
      "ascending": false
    }
  }
}
```
where <prevCur> is the base64 (or optionally encrypted version) of
```
{
  "direction": "prev",
  "id": 20
  "filterHash": <hash of complete body> 
}
```  

`filterHash` would be recommended to guard against users fiddling with the filter conditions.

**Comment by tkrop** [comment]
@dehora @meshcalero @zeitlinger @whiskeysierra @tfrauenstein Thanks for all the valuable input.

I would like to finish this issue. Can we commit on [the proposal](https://github.com/zalando/restful-api-guidelines/pull/510#discussion_r300605431) of @zeitlinger to split up the `cursor` for `GET with body` collection resources into the `query` in the body and the `cursor` containing only the *pointer*, the *direction*, and a *query hash* (to protect the pagination) provided in pagination links?

**Comment by whiskeysierra** [comment]
It's based on this statement from @meshcalero 

>  to shrink the response size

Not sure if that should be the primary objective. We had similar discussions before:

> The primary advantage of non-absolute URI is reduction of the payload size, which is better achievable by following the recommendation to use gzip compression
>
> https://opensource.zalando.com/restful-api-guidelines/#217

Not sure whether base64 encoding would preserve the *compressibility* though. 

In the end I'd also be OK with that.

**Comment by whiskeysierra** [comment]
:+1:

File: chapters/common-data-types.adoc [hunk 6]
```diff
@@ -121,11 +121,87 @@ tree_node:
...
+    next:
+      description: Pagination link pointing to the next page.
+      type: string
+      format: uri
+    last:
+      description: Pagination link pointing to the last page.
+      type: string
+      format: uri
+
+     query:
```
maxim-tschumak: The indentation seems to be incorrect - please check.

maxim-tschumak:
:+1:

chapters/common-data-types.adoc
tkrop: I fix this while merging on off @whiskeysierra PRs.

PR #512, opened 2019-06-24
doc: Update imperative statement

addityasingh:
Update imperative statement

### DIFF
--- a/chapters/deprecation.adoc
+++ b/chapters/deprecation.adoc
@@ -74,7 +74,7 @@ Clients should monitor the `Warning` header in HTTP responses to see if
 an API will be deprecated in future.
 
 [#191]
-== {MUST} Not Start Using Deprecated APIs
+== {MUST} Don't Start Using Deprecated APIs
 
 Clients must not start using deprecated parts of an API.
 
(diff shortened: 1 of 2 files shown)


chapters/deprecation.adoc
```diff
@@ -74,7 +74,7 @@ Clients should monitor the `Warning` header in HTTP responses to see if
 an API will be deprecated in future.
 
 [#191]
-== {MUST} Not Start Using Deprecated APIs
+== {MUST} Don't Start Using Deprecated APIs
```
whiskeysierra: It's meant to read as one consistent sentence:

> Must not start using deprecated APIs

tkrop: Yes. The word in the brackets is part of the sentence.

tkrop: However, the other headings are not consistently using this pattern. This, I think, makes it harder to understand.

tkrop:
@addityasingh thanks for the pull request. Our intention is to integrate the keyword into the rule reading this sentence as

> Must not start using deprecated APIs.

Reason is, that the imperative sentences of rules get shorter and are easier to integrate as references into other sentences.

I must admit, this is not that clear when looking on the local context of the rule. In addition, we still haven't applied this pattern to all rules. To improve this, I have created #513 and will close this issue.

addityasingh:
Yeah. I think what was not clear for me was some of the other rules did not follow the same pattern as `Must not start using deprecated APIs.` and MUST didn't seem a part of the sentence in them. I understood them to be the norms

PR #516, opened 2019-07-17
Added naming convention for date-time properties

### LINKED ISSUE #506: Common field names propose an unusual naming convention for timestamp fields
As part of Rule Must: Use common field names and semantics https://opensource.zalando.com/restful-api-guidelines/#174 we specify the fields `created` and `modified` to be used for audit-like timestamps.

The problem that arises is that it implicitly sets a precedent on how we should name other timestamp fields and without a `_at` suffix that makes most examples that I observe rather obscure. Examples: shipped, returned, closed, deleted, ...
They all appear to express a boolean-like condition or a status, rather than a timestamp.

Right now I see 3 options all of which have individual down sides:

1. Don't use the `at` suffix anywhere
   - consistency across fields :heavy_check_mark: 
   - in line with guidelines :heavy_check_mark: 
   - timestamp fields sound like booleans :x: 

2. Use the `at` suffix everywhere
   - consistency across fields :heavy_check_mark: 
   - in violation with guidelines :x: 
   - timestamp fields sound like timestamps :heavy_check_mark: 

3. Use `at` suffix everywhere *except* for `created` and `modified`
   - inconsistency across fields :x: 
   - in line with guidelines :heavy_check_mark: 
   - (most) timestamp fields sound like timestamps :heavy_plus_sign:

Option 3 seems to violate one of the fundamental principles of our guidelines due to the inconsistency:

> Ideally, all Zalando APIs will look like the same author created them.
>
> https://opensource.zalando.com/restful-api-guidelines/#introduction

Between option 1 and 2 I'd go with option 2 which in turn requires a (breaking) guideline change.

whiskeysierra:
Fixes #506

### DIFF
--- a/chapters/common-data-types.adoc
+++ b/chapters/common-data-types.adoc
@@ -78,10 +78,12 @@ referenced object or the relationship to the referenced object followed
 by `_id` (e.g. `customer_id` not `customer_number`; `parent_node_id` for
 the reference to a parent node from a child node, even if both have the
 type `Node`)
-* `created`: when the object was created. If used, this must be a
+* `created_at`: when the object was created. If used, this must be a
 `date-time` construct.
-* `modified`: when the object was updated. If used, this must be a
+** Used to be `created`, see <<235, Naming conventions for date/time properties>>.
+* `modified_at`: when the object was updated. If used, this must be a
 `date-time` construct.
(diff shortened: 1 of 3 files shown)


maxim-tschumak:
:+1:

tkrop:
:+1:

whiskeysierra:
:+1:

PR #517, opened 2019-07-17
Rephrased rule about null/absent properties

### LINKED ISSUE #498: Rule #123 Null values should have their fields removed – still valid?
I had a discussion in my team whether to omit or not `null` valued fields. Pretty clear is that optional fields can always be dropped, as there is no guarantee about their existence. But let’s imagine we are talking about a mandatory field.

In OpenAPI 2 there was no way to define `null` as a possible way, I learned. In [OpenAPI 3](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#schemaNullable), someone can define `nullable: true` to express `null` valued fields.

Most statically typed languages probably will treat a missing field and a `null` valued field the same (e.g. Scala encodes both with `None` in the case of an `Option`). For JavaScript someone would have to come up with a solution to check two things:

1. Does the field exist at all (otherwise type is `undefined`)?
2. Is the field `null` or something else?

By not dropping `null` valued, **mandatory** fields, the first check is unnecessary and the API spec can be treated as a real contract. Another option would be to make it explicit, that `null` is not a valid value. In this scenario it is to a certain extend unclear for the API consumer what happens with the field:

* Will it be `null` in the response of a `GET` query for the same resource, as provided in the `POST`/`PUT` request to create it initially?
* Will it even exist or omitted at all in the response of a `GET` query?

At the end I argue, the API should respect the provided input (e.g. `null`), if it is valid according to the JSON spec (which is the [case](https://tools.ietf.org/html/rfc7159#section-3) for `null`) and explicitly handle it in some way expressing the business rules a/o logic. The latter can be propagating and keeping it or dropping it and therefore not ensuring that it returns the value `null` eventually.

Omitting a field is “lack of information”, while having it set to `null` does convey information… imho. Even the description of the rule states, that there are use cases where “using null” is important, e.g. [JSON Merge Patch](https://tools.ietf.org/html/rfc7396).

whiskeysierra:
Fixes #498

### DIFF
--- a/chapters/json-guidelines.adoc
+++ b/chapters/json-guidelines.adoc
@@ -102,8 +102,8 @@ unknown.
 
 Open API 3.x allows to mark properties as `required` and as `nullable` to
 specify whether properties may be absent (`{}`) or `null` (`{"example":null}`).
-If a property is defined to be `non-required` and `nullable` (see
-<<required-nullable-row-2, 2nd row of below Table>>), this rule demands 
+If a property is defined to be not `required` and `nullable` (see
+<<required-nullable-row-2, 2nd row in Table below>>), this rule demands
 that both cases must be handled in the exact same manner by specification.
 
 The following table shows all combinations and whether the examples are
@@ -119,12 +119,29 @@ valid:
 | `false` |`false`  | {YES} | {NO}
 |===========================================
 
-While API designers and implementers may be tempted to assign both cases a
-different semantics, we explicitly decide **against** this options, because we
+While API designers and implementers may be tempted to assign different
+semantics to both cases, we explicitly decide **against** that option, because we
 think that any gain in expressiveness is far outweighed by the risk of clients
-not understanding and implementing the subtle differences correctly. Moreover,
-many major libraries and frameworks as well as some strong-typed languages
-simply do not support different semantics.
+not understanding and implementing the subtle differences incorrectly.
+
+As an example, an API that provides the ability for different users to
+coordinate on a time schedule, e.g. a meeting, may have a resource for options
+in which every user has to make a `choice`. The difference between _undecided_
+and _decided against any of the options_ could be modeled as _absent_ and
+`null` respectively. It would be safer to express the `null` case with a
+dedicated https://en.wikipedia.org/wiki/Null_object_pattern[Null object], e.g.
+`{}` compared to `{"id":"42"}`.
+
+Moreover, many major libraries have somewhere between little to no support for
+a `null`/absent pattern (see
+https://stackoverflow.com/questions/48465005/gson-distinguish-null-value-field-and-missing-field[Gson],
+https://github.com/square/moshi#borrows-from-gson[Moshi],
+https://github.com/FasterXML/jackson-databind/issues/578[Jackson],
+https://developer.ibm.com/articles/j-javaee8-json-binding-3/[JSON-B]). Especially
+strongly-typed languages suffer from this since a new composite type is required
+to express the third state. Nullable `Option`/`Optional`/`Maybe` types could be
+used but having nullable references of these types completely contradicts their
+purpose.
 
 The only exception to this rule is JSON Merge Patch {RFC-7396}[RFC 7396]) which
 uses `null` to explicitly indicate property deletion while absent properties are
--- a/chapters/json-guidelines.adoc
+++ b/chapters/json-guidelines.adoc
@@ -100,33 +100,35 @@ unknown.
 [#123]
 == {MUST} Use same semantics for `null` and absent properties
 
-OpenAPI 3 allows to mark properties as `required` and `nullable`. The
-following table showcases all combinations of both and whether the examples
-are valid in the context of the schema or not:
+Open API 3.x allows to mark properties as `required` and as `nullable` to
+specify whether properties may be absent (`{}`) or `null` (`{"example":null}`).
+If a property is defined to be `non-required` and `nullable` (see
+<<required-nullable-row-2, 2nd row of below Table>>), this rule demands 
+that both cases must be handled in the exact same manner by specification.
+
+The following table shows all combinations and whether the examples are
+valid:
 
 [cols=",,,",options="header",]
-|===
-|`+required+` |`+nullable+` |`+{}+` |`+{"example":null}+`
-|`+true+` |`+true+` | ❌ |✔️
-|`+false+` |`+true+` |✔️ |✔️
-|`+true+` |`+false+` |❌ |❌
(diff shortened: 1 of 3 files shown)


chapters/json-guidelines.adoc
```diff
@@ -99,15 +99,28 @@ unknown.
...
-OpenAPI, which is in common use, doesn't support null field values (it does
-allow omitting that field completely if it is not marked as required).
-However that doesn't prevent clients and servers sending and receiving those
-fields with null values. Also, in some cases null may be a meaningful value
-- for example, JSON Merge Patch {RFC-7396}[RFC 7396]) using null to indicate
-property deletion.
-
+== {MUST} Treat `null` and absent properties equally
+
+OpenAPI 3 allows to mark properties as `required` and `nullable`. The
```
maxim-tschumak: (minor) I'd like to have a sentence which contains something like "**same semantics** for null values and unset properties", but this is just my "taste".

chapters/json-guidelines.adoc
```diff
@@ -99,15 +99,28 @@ unknown.
...
+[cols=",,,",options="header",]
+|===
+|`+required+` |`+nullable+` |`+{}+` |`+{"example":null}+`
+|`+true+` |`+true+` | ❌ |✔️
+|`+false+` |`+true+` |✔️ |✔️
+|`+true+` |`+false+` |❌ |❌
+|`+false+` |`+false+` |✔️ |❌
+|===
+
+Non-required, nullable properties (2nd row) have two valid forms,
```
maxim-tschumak: We could either highlight the row or insert a link to the row.

maxim-tschumak:
:+1:

chapters/json-guidelines.adoc
whiskeysierra: I changed the rule title itself

maxim-tschumak:
:+1:

chapters/json-guidelines.adoc
whiskeysierra: Making it bold requires to change from code to text and I have no clue how to link to a row.

m99coder:
👍

chapters/json-guidelines.adoc
tkrop: Change for this to work.

maxim-tschumak:
:+1:

chapters/json-guidelines.adoc
```diff
@@ -99,14 +99,37 @@ unknown.
...
+| `true`  |`false`  | {NO}  | {NO}
+| `false` |`false`  | {YES} | {NO}
+|===========================================
+
+While API designers and implementers may be tempted to assign both cases a
+different semantics, we explicitly decide **against** this options, because we
+think that any gain in expressiveness is far outweighed by the risk of clients
+not understanding and implementing the subtle differences correctly. Moreover,
+many major libraries and frameworks as well as some strong-typed languages
+simply do not support different semantics.
```
dehora: I found this paragraph hard to follow.  For example, when saying `any gain in expressiveness is far outweighed by the risk of clients not understanding and implementing the subtle differences correctly`, can an example be provided. Also, what are the semantics that are being referred to here?

chapters/json-guidelines.adoc
```diff
@@ -99,14 +99,37 @@ unknown.
...
-fields with null values. Also, in some cases null may be a meaningful value
-- for example, JSON Merge Patch {RFC-7396}[RFC 7396]) using null to indicate
-property deletion.
+== {MUST} Use same semantics for `null` and absent properties
+
+Open API 3.x allows to mark properties as `required` and as `nullable` to
+specify whether properties may be absent (`{}`) or `null` (`{"example":null}`).
+If a property is defined to be `non-required` and `nullable` (see
+<<required-nullable-row-2, 2nd row of below Table>>), this rule demands 
+that both cases must be handled in the exact same manner by specification.
```
dehora: When saying—`this rule demands that both cases must be handled in the exact same manner by specification`—what is that exact same manner?

chapters/json-guidelines.adoc
whiskeysierra: Example:

- Let's say there is a doodle with options
- The property `option` is absent could be interpreted as *user has not decided yet*
- The property `option` set to `null` could be interpreted as *user actively decided against any of the options*

Those subtle differences between absent and null should be avoided.

whiskeysierra:
Up to the API to define a semantic for it. But it should be the same for
both cases.

chapters/json-guidelines.adoc
dehora: Sounds like that's worth capturing in the pr content.

chapters/json-guidelines.adoc
tkrop: @dehora indistinguishable in the outcome from the consumers and servers perspective.

chapters/json-guidelines.adoc
tkrop: @whiskeysierra can you add this short example behind the paragraph? May be you can also add another one for strong typed framework or language?

whiskeysierra: Done. Please take another look.

chapters/json-guidelines.adoc
dehora: Ok, but what's the actionable recommendation beyond coherency, ie, which combinations are recommended?

chapters/json-guidelines.adoc
dehora: lgtm

dehora:
> Up to the API to define a semantic for it. 

What stops the guidelines recommending a semantics? Consider the scenario where designers have to combine multiple APIs to offer a suite of services.

whiskeysierra:
> What stops the guidelines recommending a semantics? Consider the scenario where designers have to combine multiple APIs to offer a suite of services. Why would the single hand allow multiple meanings?

What do you have in mind? Something along the line of

> *The recommended default semantic for `null`/absent properties is to represent the lack of information.*

chapters/json-guidelines.adoc
tkrop: Yes, I had the same trouble with providing the table. The point is, we are actually only talking about the second row. The others do not count, because two others just do only allow one version and the last even none of this. I thought about dropping the table, because it may be noted misleading. @dehora What do you think?

whiskeysierra:
> What do you have in mind? Something along the line of
> 
> > _The recommended default semantic for `null`/absent properties is to represent the lack of information._

@dehora Thoughts?

tfrauenstein:
👍

whiskeysierra:
:+1:

PR #518, opened 2019-07-17
Added rules about query languages

### LINKED ISSUE #515: Filter list by age field - how to
You guide does not seem to describe a quite common usecase of filtering a list, e.g.
having a simple dataset of Pets, how would you get dogs younger than 5.
```json
{
 "age": 5,
 "_id": "5d2c87038b849c37dc9e3a5a",
 "name": "Coco",
 "animal": "dog",
}
```

whiskeysierra:
Fixes #515

### DIFF
--- a/chapters/http-requests.adoc
+++ b/chapters/http-requests.adoc
@@ -390,7 +390,7 @@ The following examples should serve as ideas:
 
 * `name=Zalando`, to query for elements based on property equality
 * `age=5`, to query for elements based on logical properties
-**  Assumung that elements don't actually have an `age` but rather a `birthday`
+**  Assuming that elements don't actually have an `age` but rather a `birthday`
 * `max_length=5`, based on upper and lower bounds (`min` and `max`)
 * `shorter_than=5`, using terminology specific e.g. to _length_
 * `created_before=2019-07-17` or `not_modified_since=2019-07-17`
--- a/chapters/http-requests.adoc
+++ b/chapters/http-requests.adoc
@@ -416,19 +416,45 @@ Aspects that sets those APIs apart from the rest include but are not limited to:
 * Dynamic filters, due to a dynamic and extensible resource model
 * Free choice of operators, e.g. `and`, `or` and `not`
 
-APIs that qualify for a specific, complex query language are encouraged to define
-it as a structured, potentially nested data structure in JSON.
+APIs that qualify for a specific, complex query language are encouraged to use nested JSON data structures and define them using OpenAPI directly. The provides the following benefits:
 
-The reason for favoring data structures over text-based query languages is that
-traditionally there are more applications generating queries than there are
-that read and interpret them. If a query language is defined in terms of data
-structures it's very easy to generate queries, even dynamically. The same is
-not true for text-based query languages.
+* Data structures are easy to use for clients
+** No special library support necessary
+** No need for string concatenation or manual escaping
+* Data structures are easy to use for servers
+** No special parsers needed
+** Semantics are attached to data structures rather than text tokens
+* Consistent with other HTTP methods
+* API is defined in OpenAPI completely
+** No external documents or grammars needed
+** Existing means are familiar to everyone
 
-Complex query languages should be specified within the OpenAPI specification
-of the API itself. The query is then obviously subject to all
 <<json-guidelines, JSON-specific rules>> and most certainly needs to make use of the
 <<get-with-body, `GET`-with-body>> pattern.
+
+=== Example
+
+The following JSON document should serve as an idea how a structured query might look like. Feel free to also get some inspiration from https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl.html[Elastic Search: Query DSL].
+
+[source,json]
+----
+{
+  "and": {
+    "name": {
+      "match": "Alice"
+    },
+    "age": {
+      "or": {
+        "range": {
+          ">": 25,
+          "<=": 50
+        },
+        "=": 65
+      }
+    }
+  }
+}
+----
 
 
 [#226]
--- a/chapters/http-requests.adoc
+++ b/chapters/http-requests.adoc
@@ -370,6 +370,67 @@ explicitly define the collection format to guide consumers as follows:
 When choosing the collection format, take into account the tool support,
 the escaping of special characters and the maximal URL length.
 
+[#236]
+== {SHOULD} Design simple query languages using query parameters
+
+We prefer the use of query parameters to describe resource-specific
+query languages for the majority of APIs because it's native to HTTP,
+easy to extend and has excellent implementation support in HTTP clients
+and web frameworks.
+
+Query parameters should have the following aspects specified:
+
+* Reference to corresponding property, if any
+* Value range, e.g. inclusive vs. exclusive
+* Comparison semantics (equals, less than, greater than, etc)
+* Implications when combined with other queries, e.g. _and_ vs. _or_
+
+How query parameters are named and used is up to individual API designers.
+The following examples should serve as ideas:
+
+* `name=Zalando`, to query for elements based on property equality
+* `age=5`, to query for elements based on logical properties
+**  Assumung that elements don't actually have an `age` but rather a `birthday`
+* `max_length=5`, based on upper and lower bounds (`min` and `max`)
+* `shorter_than=5`, using terminology specific e.g. to _length_
+* `created_before=2019-07-17` or `not_modified_since=2019-07-17`
+** Using terminology specific e.g. to time: _before_, _after_, _since_ and _until_
+
+We don't advocate for or against certain names because in the end
+APIs should be free to choose the terminology that fits their domain the best.
+
+
+[#237]
+== {SHOULD} Design complex query languages using JSON
+
+Minimalistic query languages based on <<236, query parameters>> are suitable for
+simple use cases with a small set of available filters that are combined in one
+way and one way only (e.g. _and_ semantics).
+
+Some APIs will have a need for sophisticated and more complex query languages.
+Dominant examples are APIs around search (incl. faceting) and product catalogs.
+
+Aspects that sets those APIs apart from the rest include but are not limited to:
+
+* Unusual high number of available filters
+* Dynamic filters, due to a dynamic and extensible resource model
+* Free choice of operators, e.g. `and`, `or` and `not`
+
+APIs that qualify for a specific, complex query language are encouraged to define
+it as a structured, potentially nested data structure in JSON.
+
+The reason for favoring data structures over text-based query languages is that
+traditionally there are more applications generating queries than there are
+that read and interpret them. If a query language is defined in terms of data
+structures it's very easy to generate queries, even dynamically. The same is
+not true for text-based query languages.
+
+Complex query languages should be specified within the OpenAPI specification
+of the API itself. The query is then obviously subject to all
+<<json-guidelines, JSON-specific rules>> and most certainly needs to make use of the
+<<get-with-body, `GET`-with-body>> pattern.
+
+
 [#226]
 == {MUST} Document Implicit Filtering
 
@@ -384,7 +445,7 @@ Implicit filtering could be done on:
 
 In such cases, the implicit filtering must be in the API specification (in its description).
 
-Consider <<227,caching considerations>> when implicitely filtering.
+Consider <<227,caching considerations>> when implicitly filtering.
 
 Example:
 
--- a/chapters/http-requests.adoc
+++ b/chapters/http-requests.adoc
@@ -403,12 +403,9 @@ APIs should be free to choose the terminology that fits their domain the best.
 [#237]
 == {SHOULD} Design complex query languages using JSON
 
-Minimalistic query languages based on <<236, query parameters>> are suitable for
-simple use cases with a small set of available filters that are combined in one
-way and one way only (e.g. _and_ semantics).
+Minimalistic query languages based on <<236, query parameters>> are suitable for simple use cases with a small set of available filters that are combined in one way and one way only (e.g. _and_ semantics). Simple query languages are generally preferred over complex ones.
 
-Some APIs will have a need for sophisticated and more complex query languages.
-Dominant examples are APIs around search (incl. faceting) and product catalogs.
+Some APIs will have a need for sophisticated and more complex query languages. Dominant examples are APIs around search (incl. faceting) and product catalogs.
 
 Aspects that set those APIs apart from the rest include but are not limited to:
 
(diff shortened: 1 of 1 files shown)


chapters/http-requests.adoc
```diff
@@ -361,6 +361,67 @@ explicitly define the collection format to guide consumers as follows:
...
+* `name=Zalando`, to query for elements based on property equality
+* `age=5`, to query for elements based on logical properties
+**  Assumung that elements don't actually have an `age` but rather a `birthday`
+* `max_length=5`, based on upper and lower bounds (`min` and `max`)
+* `shorter_than=5`, using terminology specific e.g. to _length_
+* `created_before=2019-07-17` or `not_modified_since=2019-07-17`
+** Using terminology specific e.g. to time: _before_, _after_, _since_ and _until_
+
+We don't advocate for or against certain names because in the end
+APIs should be free to choose the terminology that fits their domain the best.
```
maxim-tschumak: Do you think [Spring Data JPA repository queries](https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#repositories.query-methods.query-creation) is a good example of how to name the query parameters? Maybe we should add it as a reference. On the other hand we probably should not compare REST APIs with persistence layer interfaces.

chapters/http-requests.adoc
```diff
@@ -361,6 +361,67 @@ explicitly define the collection format to guide consumers as follows:
...
+== {SHOULD} Design complex query languages using JSON
+
+Minimalistic query languages based on <<236, query parameters>> are suitable for
+simple use cases with a small set of available filters that are combined in one
+way and one way only (e.g. _and_ semantics).
+
+Some APIs will have a need for sophisticated and more complex query languages.
+Dominant examples are APIs around search (incl. faceting) and product catalogs.
+
+Aspects that sets those APIs apart from the rest include but are not limited to:
```
maxim-tschumak: `sets` -> `set`

chapters/http-requests.adoc
```diff
@@ -370,6 +370,67 @@ explicitly define the collection format to guide consumers as follows:
...
+The reason for favoring data structures over text-based query languages is that
+traditionally there are more applications generating queries than there are
+that read and interpret them. If a query language is defined in terms of data
+structures it's very easy to generate queries, even dynamically. The same is
+not true for text-based query languages.
+
+Complex query languages should be specified within the OpenAPI specification
+of the API itself. The query is then obviously subject to all
+<<json-guidelines, JSON-specific rules>> and most certainly needs to make use of the
+<<get-with-body, `GET`-with-body>> pattern.
```
tkrop: Use macro: ```<<get-with-body, `GET`-with-body>>``` => `{GET-with-body}`

whiskeysierra:
> I do not share the opinion, that query languages should be expressed as JSON

Why not? 

- We use `POST /things` with JSON, instead of `INSERT INTO things ...`
- We use `GET /things/1`, instead of `SELECT ... FROM things WHERE id = 1`
- We use `PATCH /things/1`with JSON, instead of `UPDATE things SET ... WHERE id = 1`
- We use `DELETE /things/1`, instead of `DELETE FROM things WHERE id = 1`

Why is a non-trivial read (GET) different from our other use cases? 

There is an obvious reason why we prefer JSON of some text-based format like SQL for everything else already: It's machine readable and even more importantly it's *machine-writable*. SQL is a language that was designed to be written by humans, not machines. (It's original name is SEQUEL = _Structured **English** Query Language_.)

The whole attack vector of [SQL injection](https://en.wikipedia.org/wiki/SQL_injection) only exists because it's a text-based format without the notion of escaping/quoting like JSON or XML.

APIs are **Application** Programming Interfaces, so the primary purposes of query languages should be the same: They should be meant to be implemented/used/parsed and *written* by applications. Data structures are far superior than strings in that context.

whiskeysierra:
> 1\. there are just much better ways to specify advanced query languages.

Like what? BNF?

> 3\. Personally I must admit, that I also find creating JSON structures tiresome. A simple query-language string is constructed much easier and way more efficient.

Easier for humans maybe, but not in code. String concat would be easy but its insufficient. Everything beyond string concat might work but is no longer easy.

Have you ever dynamically constructed a SQL query string because you had a dynamic set or filters?
There is a reason why projects like [jOOQ](https://www.jooq.org/) or JPA Criteria API exist

Constructing queries based on data structure requires nothing beyond the core libraries of the language. E.g. Java:

```java
var filter = List.of(Map.of("age", Map.of("<", 5)))
```

> This is based on sloppy developers not creating theire internal SQL queries correctly.

SQL queries are just strings, so it's not sloppiness, but rather that the tool is insufficient to prevent common mistakes like that. APIs (and SQL is an API for the DB) should be hard to misuse. SQL fails at that.

> Consistency is nice, but not on the costs of comprehensibility and ease of use - also from code perspective.

Every single programming language has support for the most basic data structures (otherwise they wouldn't be able to integrate with JSON APIs). Support for constructing SQL strings either require external libraries or database drivers.

Apart from ease of use (where data structures clearly win over string concatenation) the fundamental issue to me is that a query string requires two new things from engineers:

1. **What** the language is trying to represent, in this case a set of filters: *semantic*
2. **How** the language represents that in technical terms: *format*

Custom, string-based query languages require both. Query languages defined in terms of data structures are already familiar in terms of format and technology support. So engineers only have to understand the semantics of it.

tkrop:
@whiskeysierra  You really find `var filter = List.of(Map.of("age", Map.of("<", 5)))` more readable than `age < 5` - and this example is still trivial?

whiskeysierra:
> and this example is still trivial?

Agreed, that is way too oversimplified. You'd need to compare e.g. this:

```java
// better not forget the escaping here!
// int, so probably safe?!
return "name = \"" + escape(name) + "\" or age < " + age; 

// just an example, no clue what to escape actually
private String escape(String raw) {
    String escaped = raw;
    escaped = escaped.replace("\\", "\\\\");
    escaped = escaped.replace("\"", "\\\"");
    escaped = escaped.replace("\b", "\\b");
    escaped = escaped.replace("\f", "\\f");
    escaped = escaped.replace("\n", "\\n");
    escaped = escaped.replace("\r", "\\r");
    escaped = escaped.replace("\t", "\\t");
    // TODO optimize!
    return escaped;
}
```

to

```java
return Map.of(
    "and", Map.of(
        "name", Map.of("=", name),
        "age", Map.of("<", age)));
```

tkrop:
@whiskeysierra But the first code snipped neither reflect the customer nor the service perspective of the complex query language. It only reflects the implementation perspective of the simple query language interface.

whiskeysierra:
> But the first code snipped neither reflect the customer nor the service perspective of the complex query language. It only reflects the implementation perspective of the simple query language interface.

No, it directly represents what the client (customer?) needs to prepare in order to send it to the server. Or what kind of customer are you talking about? End user with the ability to manually use this?

whiskeysierra:
Relevant part of an interesting talk that mentions this issue: https://youtu.be/ROor6_NGIWU?t=3428

chapters/http-requests.adoc
```diff
@@ -370,6 +370,67 @@ explicitly define the collection format to guide consumers as follows:
...
+* Value range, e.g. inclusive vs. exclusive
+* Comparison semantics (equals, less than, greater than, etc)
+* Implications when combined with other queries, e.g. _and_ vs. _or_
+
+How query parameters are named and used is up to individual API designers.
+The following examples should serve as ideas:
+
+* `name=Zalando`, to query for elements based on property equality
+* `age=5`, to query for elements based on logical properties
+**  Assumung that elements don't actually have an `age` but rather a `birthday`
```
nicklasfrahm: Small typo in `line 393`:

```diff
- **  Assumung that elements don't actually have an `age` but rather a `birthday`
+ **  Assuming that elements don't actually have an `age` but rather a `birthday`
```

ePaul:
*Update from the guild meeting*

* we might need to motivate the reasoning for the JSON rule (compared to other text based query languages) more.
   * advantage mentioned by @tfrauenstein: The API spec itself describes how the query looks like, and what it means, instead of having to put this into a separate document.
* What is the relation to GraphQL? (Possibly postpone this part until we got the GraphQL rules done).
* Link to Elastic search API as an example

nicklasfrahm:
Just throwing my opinion in here. I think before one can decide on the format of a query language, one should define the use-cases of such a query language and develop the requirements from this.

Some use-cases were already mentioned:
- **Dynamic generation of complex queries** through code, e.g. based on search input fields an algorithm generates a query.
- **Reduction of multiple requests** to different resources for complex queries.
- **Increased coverage of edge-cases** which are not supported by standard means and interfaces provided by resources.

From this, you could derive the following requirements:
- **Understandable for developers.** This does not have to mean that it has to be human-readable, but requires the language format to be well-documented.
- **Secure by design** with an easy integration of additional restrictions.
- **Machine-native** by being easily parse-able and processable by machines.

Speaking about something that can be serialized as JSON, an example **could** be the [query language of MongoDB](https://docs.mongodb.com/manual/tutorial/query-documents/). It can even be serialized to JSON.

whiskeysierra:
> I think before one can decide on the format of a query language, one should define the use-cases of such a query language and develop the requirements from this.

That's why I would leave it at the current set of guidelines. We can't know of specific requirements, but I can't think of any concrete requirement that would make a text-based query language more appealing than a JSON-based one.

> * **Dynamic generation of complex queries** through code, e.g. based on search input fields an algorithm generates a query.

That will always be a use case, otherwise it wouldn't have the need for a complex query language in the first place.

> * **Reduction of multiple requests** to different resources for complex queries.

Do you mean Bulk requests?

> * **Increased coverage of edge-cases** which are not supported by standard means and interfaces provided by resources.

I don't get that, tbh.

> * **Understandable for developers.** This does not have to mean that it has to be human-readable, but requires the language format to be well-documented.

Hence the guidance to document it within the OpenAPI spec.

> * **Secure by design** with an easy integration of additional restrictions.

In terms of escaping/injection/... or do you have something else in mind?

> * **Machine-native** by being easily parse-able and processable by machines.

Again this to me is always a use case. I can't imagine a case where someone wouldn't want this.

> Speaking about something that can be serialized as JSON, an example **could** be the [query language of MongoDB](https://docs.mongodb.com/manual/tutorial/query-documents/). It can even be serialized to JSON.

Good idea. Might be worth adding as an alternative example.

nicklasfrahm:
> That's why I would leave it at the current set of guidelines. We can't know of specific requirements, but I can't think of any concrete requirement that would make a text-based query language more appealing than a JSON-based one.

I do agree, I was just trying to outline what needs to be considered, if a more specific recommendation should be made, be it either for or against JSON.

> Do you mean Bulk requests?

In this case I was actually thinking more of complex queries, that require operators, such as OR, AND or IN. These can in some cases only be done by help of multiple requests to different resources. Hence, a query language can reduce the amount of requests and make the communication more efficient.

> I don't get that, tbh.

In this case I was thinking of unsupported query parameters or other filters on resources. This way a query language can be a temporary workaround for parameters that are not yet or not at all supported.

> In terms of escaping/injection/... or do you have something else in mind?

For the MongoDB JSON queries, the developer can easily look for certain patterns in the _query object_ that potentially look malicious. Thus it is easy to identify potential security caveats.

**Final note:** @whiskeysierra I was not trying to contradict, but rather to sum up what needs to be taken into consideration for a more specific recommendation.

whiskeysierra:
> In this case I was actually thinking more of complex queries, that require operators, such as OR, AND or IN. These can in some cases only be done by help of multiple requests to different resources. Hence, a query language can reduce the amount of requests and make the communication more efficient.

That's true. Haven't thought about that but the use of `or` filters definitely would allow for bulk-style requests. 

> Here I was thinking of unsupported query parameters or other filters on resources. This way a query language can be a temporary workaround for parameters that are not yet or not at all supported.

Can you give me an example?

> I was not trying to contradict, but rather to sum up what needs to be taken into consideration for a more specific recommendation.

After re-reading my response I have to admit the wording wasn't really in line with my intended tone. Didn't meant to be harsh or defensive.

nicklasfrahm:
> After re-reading my response I have to admit the wording wasn't really in line with my intended tone. Didn't meant to be harsh or defensive.

This is not at all how I took your response. Just seemed to that I caused more confusion with my comment than I intended. 😄 

> Can you give me an example?

The simplest would be the `max_age` query parameter again, that is **supported by your docs and specifications**, but might not be supported by the implementation from the start in the development process of a RESTful service.

My point is that a query language can improve data accessibility significantly for a semi-feature-complete API. An API query language can then help to implement this temporarily until these query parameters are in place. An example with a MongoDB query could then be:
`GET /users?q={"age":{"$lte":20}}`

I am however not sure if this is save, thus I think that the query JSON should be URI encoded. This is for example something that would need to be discussed as the query would then yield to be this, which is really hard to read for humans:
`GET /users?q=%7B%22age%22%3A%7B%22%24lte%22%3A20%7D%7D%0A`

whiskeysierra:
> I am however not sure if this is save, thus I think that the query JSON should be URI encoded.

Preferred way would be https://opensource.zalando.com/restful-api-guidelines/#get-with-body

nicklasfrahm:
Makes sense. One question on the mentioned **hint**. Would that be an HTTP header?

whiskeysierra:
No, just a sentence in the operation's description in OpenAPI. `X-HTTP-Method-Override` would be a good candidate but it requires `Content-Type: application/x-www-form-urlencoded` unfortunately.

tkrop:
@whiskeysierra @nicklasfrahm In case of the *complex query language rule*, we are talking about an absolutely rare case for API design only relevant for *database management systems*, e.g elastic search, mongoDB etc. I do not see much value in giving advise about these here, as they have their special problems, tools, and requirements, e.g. for distributed query execution planning and optimization, we cannot foresee.

Please let us separate this discussion from the suggested *simple query language* rule, and keep the advise as simple as possible. E.g. *do not write complex query languages - unless you have to! Make sure that you know what you do and that your query language is simple to understand and use.*

whiskeysierra:
> we are talking about an absolutely rare case for API design only relevant for _database management systems_, e.g elastic search, mongoDB etc.

Not really. Any e-commerce platform that has a catalog/assortment/search of some sort will have this issue and unless you're willing to expose your implementation you'd need to have some sort of query language for those cases.

whiskeysierra:
> Please let us separate this discussion from the suggested _simple query language_ rule

I can do that.

> and keep the advise as simple as possible. E.g. _do not write complex query languages - unless you have to! Make sure that you know what you do and that your query language is simple to understand and use._

That's a bit to simplistic. Our advice in the guideline should include what it means to be easy to use for clients.

tfrauenstein:
👍

tkrop:
:+1:

PR #519, opened 2019-07-29
Made using the money object a MUST rule

whiskeysierra:
I don't see a good reason why someone wouldn't use the `Money` object as a representation of monetary amounts.


ePaul:
:+1:

whiskeysierra:
:+1:

Issue #513, opened 2019-06-25
Make consistent use of must/should/may keyword headings.


tkrop:
Currently, not all headings make consistent use of integrating the initial must/should/may keyword into the sentence. In addition Headings follow different patterns of upper and lower cases. Task of this sentence is to

* [x] make consistent use of the keyword within the sentence, and
* [x] use a consistent pattern for upper and lower cases.
* [x] maybe to remove the `:` between the keyword and the heading to make the integration clearer.

Since we refer to rules by numbers, this task should have no impact on Zally and direct linking.

tkrop:
Please vote in this comment whether to remove the `:` between keyword and heading.

tkrop:
Please vote in this comment whether to use **lower** letter headings using an initial big letter.

zeitlinger:
vote: lower, without ":"

whiskeysierra:
https://css-tricks.com/almanac/properties/t/text-transform/ Might help

whiskeysierra:
This should do the trick:

```css
h3 {
  text-transform: lowercase;
}

h3 strong {
  text-transform: capitalize;
}

h3 code {
  text-transform: none;
}
```

And removing the colons from `index.adoc`.

vadeg:
How about uppercase?

### TOC

![api-guidelines-uppercase-sample](https://user-images.githubusercontent.com/5587834/63009393-b0a3dc00-be84-11e9-9b8d-6b272214d7ab.png)

### Page
![api-uppercase-details-sample](https://user-images.githubusercontent.com/5587834/63009436-c74a3300-be84-11e9-96e4-c1c242b40cf2.png)

tkrop:
@vadeg Do you suggest to vote on *uppercase* Keywords? Than add a comment to vote on as @zeitlinger did.

vadeg:
@tkrop updated

ePaul:
Discussion in the meeting:

* no `:`
* we'll reword all titles so the text is a proper sentence (including the keyword)
* we'll use title case
* keywords for now stay as-is, but can easily be changed later

→ @tkrop volunteered to do the necessary text changes.

zeitlinger:
:+1:

PR #525, opened 2019-07-31
GH-513 Changed casing of headlines

whiskeysierra:
Partially solves #513 

Unfortunately this method lower cases words like:
- API
- ISO
- OAuth
- OpenAPI
- RFC
- URL


tkrop:
@whiskeysierra good try, but unfortunately not working that good. We have a number of keywords, that must be capitalized independent of title style, e.g. HTTP, JSON, ...

May be we have to do it manually.

Issue #234, opened 2017-05-12
Improve HTTP method guidance


dehora:
The guidance ["Use HTTP Methods Correctly"](https://github.com/zalando/restful-api-guidelines/blob/master/http/Http.md#-bookmust--use-http-methods-correctly) could be improved and needs some editing:

- In general: the guidance tries to summarise what HTTP specifies, but doesn't explain how to map methods to common concerns (such as creating a resource or updating). It's not optimised for people trying to solve problems. See [Google's guidelines](https://cloud.google.com/apis/design/standard_methods) for an alternate layout that works from common API needs to HTTP.

- GET. There's no need to say anything about 404 or collections. This section can be shortened.

- GET with Body. Should be a top level guideline. It comes up regularly and should be made easier to find in the guidelines for API designers. The "don't use headers for queries" statement should also be a top level guideline.

- POST. The post to collection pattern is described but could use some rewording.

- PATCH and updating. The wording should be rephrased to not have one approach per "endpoint" to having one approach per API, or at least trying to reduce the variability per API to make customers jobs easier.

tfrauenstein:
@dehora As discussed, let us proceed like follows here:

a) Yes, an API developer use case oriented presentation could be helpful, abstracting, respectively referencing to HTTP and REST details. Typical use cases could be "c/r/u/d resource object", "retrieval from data collection", "batching" etc. Let us also evaluate #guild-api chat to identify most relevant use cases. I would propose additional top-level chapter for here -- it does not define guidelines, but helps applying the guidelines for specific use cases. Let us open in this sense a separate issue for further tracking. 

b) The other topics are more editorial -- partially already addressed by latest version. Let us create here a concrete PR to make progress and close this issue.

Thank you!

ePaul:
**From the meeting**:

Closing this for now. @dehora Feel free to reopen when you have more detailed proposals.

PR #532, opened 2019-09-02
feat: repair money object link (#531)

### LINKED ISSUE #531: Link to mony object not working
Link to money object not working. In addition, I would like to have semantic version on models we create as best practice to keep this more stable.

tkrop:
fixes #531.


chapters/common-data-types.adoc
```diff
@@ -20,7 +20,7 @@ APIs are encouraged to include a reference to the global schema for Money.
 SalesOrder:
   properties:
     grand_total:
-      $ref: 'https://opensource.zalando.com/restful-api-guidelines/money.yaml#/Money'
```
whiskeysierra: We're not doing this for problem. Should we do that as well?

whiskeysierra: @tfrauenstein I vaguely remember that you had the idea of only referencing immutable schemas which would remove the need for versioning. (With the downside of not being able to evolve.)

whiskeysierra: That's technically a breaking change. Practically I couldn't find a single usage in github right now.

tkrop: No, practically it has never worked ;), so it cannot be a breaking change.

If we see it from a semantic perspective, it is a technical non-breaking change, because the reference resource stays the same.

tkrop: [Rule #234](https://opensource.zalando.com/restful-api-guidelines/#234) is open with respect to link format - it only defines the requirement for immutability and durability.

It is debatable whether a resource file should a semantic version in the file name.

* The Open API specs follow this approach to allow for immutability in case of specification updates.
* Our API repository is following the same schema requiring version and revision numbers in the path.
* Many packaging systems are following also an enforced semantic version approach.

I would say it is good practice for file names too. It prevent surprising consumers with an unclear semantic version number on a file if a change is later required/proposed.

whiskeysierra: I agree. Still leaves the question whether we should do the same for Problem.

tkrop: I see no immediate requirement for action for Problem, as it is already heavily used in the way it is. This was different for money - there was time for adjustment.

whiskeysierra:
:+1:

zeitlinger:
:+1:

PR #539, opened 2019-09-10
feat: reorder chapters and remove duplicate rule (#534)

### LINKED ISSUE #534: Duplicate rule for currency/country codes?
We love the work done on these.  While reviewing we noticed that 
[May: Use Standards for Country, Language and Currency Codes [170]](https://opensource.zalando.com/restful-api-guidelines/#170) and 
[May: Standards could be used for Language, Country and Currency [128]](https://opensource.zalando.com/restful-api-guidelines/#128) seem to be duplicates of each other.

Was wondering if this was on purpose?

tkrop:
fixes #534.

In addition, I suggest to advance this rule from `MAY` to `SHOULD`, as I think it is good practice to use these standard format instead of inventing own formats.


whiskeysierra:
:+1:

tfrauenstein:
👍

Issue #494, opened 2019-02-26
#206 is outdated (there are no permissions for events)


ePaul:
[Rule 206](https://opensource.zalando.com/restful-api-guidelines/#206) says:

> **Must: Permissions on events must correspond to API permissions [206]**
>
> If a resource can be read synchronously via a REST API and read asynchronously
> via an event, the same read-permission must apply: We want to protect access to
> data, not the way data is accessed.

Nakadi never supported permissions/scopes for event types, and instead of adding it (what we intended when this rule was written) they instead added a whitelisting by service-name. We should update the guideline to reflect this.

(This rule was added in #171/#189.)

ePaul:
I will open a PR to remove this rule.

PR #542, opened 2019-09-23
reword Rule 206 (issue #494)

ePaul:
This rewords rule 206, making it a rather soft rule.
Alternatively, we could also just remove it.

### DIFF
--- a/chapters/events.adoc
+++ b/chapters/events.adoc
@@ -679,11 +679,19 @@ algorithm, or the generation of enriched data, and which might not be
 stored as entity by the service.
 
 [#206]
-== {MUST} Permissions on events must correspond to API permissions
+== {MUST} Carefully manage read and write permissions for events
 
-If a resource can be read synchronously via a REST API and read
-asynchronously via an event, the same read-permission must apply: We
-want to protect access to data, not the way data is accessed.
+Our event bus system (Nakadi) currently doesn't support permission-based
+access control (which would allow to use the same permissions as for the
+corresponding APIs), but requires explicitly listing the services and
+employees which have access.
+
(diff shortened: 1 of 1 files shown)


chapters/events.adoc
```diff
@@ -679,11 +679,19 @@ algorithm, or the generation of enriched data, and which might not be
...
-
-If a resource can be read synchronously via a REST API and read
-asynchronously via an event, the same read-permission must apply: We
-want to protect access to data, not the way data is accessed.
+== {MUST} Carefully manage read and write permissions for events
+
+Our event bus system (Nakadi) currently doesn't support permission-based
+access control (which would allow to use the same permissions as for the
+corresponding APIs), but requires explicitly listing the services and
+employees which have access.
```
ePaul: Do we need this paragraph? Should we word it more neutrally?

chapters/events.adoc
```diff
@@ -679,11 +679,19 @@ algorithm, or the generation of enriched data, and which might not be
...
+
+Event type owners need to think about who needs access to their
+events' data, and give *read access* to those services (+ usually the
+developers of those services, for ease of debugging, unless it's
+sensitive data).
+*Write access* should only be given to the (normally only one) service
+which is supposed to publish events on this event type.
```
ePaul: I wonder whether I should separately mention the (less often) use case, where an event type is meant as input for a service, not as an output, i.e. where there is only one consumer (which is the event type owner and API provider), but multiple publishers (the clients).

tkrop:
:+1:  for removal.

ePaul:
**Discussion in the meeting**: It's better to just remove the rule, as we can't enforce it anyways.

PR #543, opened 2019-09-24
remove rule 206

### LINKED ISSUE #494: #206 is outdated (there are no permissions for events)

ePaul:
As discussed in today's guild meeting, and documented in #542,
we want to remove this rule completely instead of trying to rewrite it.

Fixes #494.

### DIFF
--- a/chapters/events.adoc
+++ b/chapters/events.adoc
@@ -678,13 +678,6 @@ representation than that defined for an API
 algorithm, or the generation of enriched data, and which might not be
 stored as entity by the service.
 
-[#206]
-== {MUST} Permissions on events must correspond to API permissions
-
-If a resource can be read synchronously via a REST API and read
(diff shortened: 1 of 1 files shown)


vadeg:
LGTM

ePaul:
:+1:

whiskeysierra:
:+1:

PR #546, opened 2019-10-23
feat: clarified default on fields rule; cleaned up code; fixed bugs (#533)

### LINKED ISSUE #533: Clarify interaction of `fields` and `default`
It's not clear whether the [`fields` parameter](https://opensource.zalando.com/restful-api-guidelines/index.html#157) is intended to have a `default` (at the level of design), or if it's intended to be used only to create partials from a whole. As things stand in the guidelines, it seems to not be an incorrect approach technically speaking. to combine the two so as to allow a resource using fields to return a partial object by default.The follow on implication is an API designer using these guidelines can require a client that wants to obtain a whole object to ask for every field at every level of depth.

tkrop:
fixes #533.

### DIFF
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -1,14 +1,14 @@
 Bill de Hóra <bill.dehora@zalando.ie>
-Michael Duergner <michael.duergner@zalando.fi>
 Paul Ebermann <paul.ebermann@zalando.de>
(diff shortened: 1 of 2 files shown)


vadeg:
👍

whiskeysierra:
:+1:

PR #561, opened 2020-01-21
Added problem type examples

### LINKED ISSUE #550: Examples for good problem types
In [Rule 176](https://opensource.zalando.com/restful-api-guidelines/#176) we advocate for `application/problem+json` but we don't give good examples and guidance for:

- when to use a custom problem type
- good values for problem types

While the RFC encourages a URI reference that can be dereferenced:

> This specification encourages that, when
      dereferenced, it provide human-readable documentation for the
      problem type (e.g., using HTML [W3C.REC-html5-20141028]).
>
> https://tools.ietf.org/html/rfc7807#page-5

I'd suggest to pick non-URL types, e.g. `my-custom-problem-type`. It frees us from situation where locations URLs change and one would need to server a proper HTML page there. We have OpenAPI for that, imho.

whiskeysierra:
Fixes #550

### DIFF
--- a/chapters/http-status-codes-and-errors.adoc
+++ b/chapters/http-status-codes-and-errors.adoc
@@ -327,7 +327,8 @@ responses:
 You may define custom problem types as extensions of the Problem JSON object
 if your API needs to return specific, additional and detailed error information.
 
-The specification encourages that custom problem types are URI references that point to human-readable documentation, **but** we deliberately decided against that. URLs tend to be fragile and not very stable over a longer period. Hosting documentation often requires to bind to a specific tool or have DNS records that contain volatile organization structures, e.g. team names. Another reason is that all the important parts of an API must be documented using <<101, OpenAPI>> anyway.
+Problem `type` identifiers in our APIs are not meant to be resolved.
+The RFC encourages that custom problem types are URI references that point to human-readable documentation, **but** we deliberately decided against that. URLs tend to be fragile and not very stable over a longer period. Hosting documentation often requires to bind to a specific tool or have DNS records that contain volatile organization structures, e.g. team names. Another reason is that all the important parts of an API must be documented using <<101, OpenAPI>> anyway.
 
 In order to stay compatible the proposed pattern for custom problem types is to use https://tools.ietf.org/html/rfc3986#section-4.1[relative URI references]:
 
@@ -358,4 +359,3 @@ Stack traces contain implementation details that are not part of an API,
 and on which clients should never rely. Moreover, stack traces can leak
 sensitive information that partners and third parties are not allowed to
 receive and may disclose insights about vulnerabilities to attackers.
-
--- a/chapters/http-status-codes-and-errors.adoc
+++ b/chapters/http-status-codes-and-errors.adoc
@@ -324,8 +324,22 @@ responses:
           $ref: 'https://opensource.zalando.com/problem/schema.yaml#/Problem'
 ----
 
-You may define custom problem types as extension of the Problem JSON object 
-if your API need to return specific additional error detail information.
+You may define custom problem types as extensions of the Problem JSON object
+if your API needs to return specific, additional and detailed error information.
+
+The specification encourages that custom problem types are URI references that point to human-readable documentation, **but** we deliberately decided against that. URLs tend to be fragile and not very stable over a longer period. Hosting documentation often requires to bind to a specific tool or have DNS records that contain volatile organization structures, e.g. team names. Another reason is that all the important parts of an API must be documented using <<101, OpenAPI>> anyway.
+
+In order to stay compatible the proposed pattern for custom problem types is to use https://tools.ietf.org/html/rfc3986#section-4.1[relative URI references]:
+
+* `/out-of-stock`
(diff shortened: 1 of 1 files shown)


chapters/http-status-codes-and-errors.adoc
```diff
@@ -324,8 +324,22 @@ responses:
...
-You may define custom problem types as extension of the Problem JSON object 
-if your API need to return specific additional error detail information.
+You may define custom problem types as extensions of the Problem JSON object
+if your API needs to return specific, additional and detailed error information.
+
+The specification encourages that custom problem types are URI references that point to human-readable documentation, **but** we deliberately decided against that. URLs tend to be fragile and not very stable over a longer period. Hosting documentation often requires to bind to a specific tool or have DNS records that contain volatile organization structures, e.g. team names. Another reason is that all the important parts of an API must be documented using <<101, OpenAPI>> anyway.
+
+In order to stay compatible the proposed pattern for custom problem types is to use https://tools.ietf.org/html/rfc3986#section-4.1[relative URI references]:
+
+* `/out-of-stock`
```
tkrop: How about prefixing/name-spacing them, so they could be potentially resolved from the API implementing server?

whiskeysierra: I'd let go of the *"feature"* to resolve them and get some HTML page in return.

tkrop: Yes, and that is exactly why I would suggest to use a coordinated relative address that makes clear it is different from the APIs resources. How about using: `/problems/{type}`

whiskeysierra: But would servers need to support those endpoints then?

tkrop: They do not need, however, they may if they want. I just suggest not to pollute the standard resource space by this option.

whiskeysierra: I wouldn't allow it. That would remove the whole problem of polluting the resource space. Also, I believe if we would allow it it would weaken our commitment to OpenAPI and API first.

tkrop: I must admit, I do not understand this argument.

ePaul: We need to make clear that those problem type URI references (contrary to what the RFC says) are **not meant for resolving them** – this currently is missing in the text. They are just "fake URI references" (look like URI references, but are not actually URI references).

With a separate problem namespace (as proposed by Tronje) we avoid confusion if by chance the same names as in the problem types are also used for other resources in this service, and someone still tries to resolve them. A service could also send a canned response like "problem types are not meant for resolution, please check the API documentation for their meaning", if wished.

chapters/http-status-codes-and-errors.adoc
```diff
@@ -324,8 +324,22 @@ responses:
           $ref: 'https://opensource.zalando.com/problem/schema.yaml#/Problem'
 ----
 
-You may define custom problem types as extension of the Problem JSON object 
-if your API need to return specific additional error detail information.
+You may define custom problem types as extensions of the Problem JSON object
+if your API needs to return specific, additional and detailed error information.
+
+The specification encourages that custom problem types are URI references that point to human-readable documentation, **but** we deliberately decided against that. URLs tend to be fragile and not very stable over a longer period. Hosting documentation often requires to bind to a specific tool or have DNS records that contain volatile organization structures, e.g. team names. Another reason is that all the important parts of an API must be documented using <<101, OpenAPI>> anyway.
```
ePaul: This might make it clearer that we don't want to resolve these:
```suggestion
Problem `type` identifiers in our APIs are not meant to be resolved.
The RFC encourages that custom problem types are URI references that point to human-readable documentation, **but** we deliberately decided against that. URLs tend to be fragile and not very stable over a longer period. Hosting documentation often requires to bind to a specific tool or have DNS records that contain volatile organization structures, e.g. team names. Another reason is that all the important parts of an API must be documented using <<101, OpenAPI>> anyway.
```

whiskeysierra: That is lacking a reason which is important because we go against the spec.

ePaul: I didn't remove the reason (it's just shown a bit strange in the diff view).

chapters/http-status-codes-and-errors.adoc
whiskeysierra: > if by chance the same names as in the problem types are also used for other resources in this service, and someone still tries to resolve them

Is that really an issue? The chance the the name of a problem matches one of the resources is pretty slim, imho. But even then I'd rather make a strong statement **once** in our guidelines than forcing a prefix upon **all** problem definitions in all APIs and implementations.

whiskeysierra: It's cheaper to let 1% of engineers run into the (unlikely) misunderstanding of problem types and resources compared to 100% of engineers who need to specify and implement APIs.

tkrop:
:+1:

ePaul:
:+1:

PR #563, opened 2020-01-27
feat: clarify batch/bulk response using 207 (#552)

### LINKED ISSUE #552: Add clarification to rule 152
We noted that rule 152 only says to always use 207 (not only in the partial failure case, but also in the success case) and always return batch responses, but doesn't have a rationale for this.

The reason was that we want to force the client to always look at the result, in order to not miss any problems. Also this way is less logic to implement in both client and server.

tkrop:
fixes #552.

### DIFF
--- a/chapters/http-status-codes-and-errors.adoc
+++ b/chapters/http-status-codes-and-errors.adoc
@@ -244,7 +244,7 @@ define the following approach:
   failure cannot be related to individual items of the batch or bulk request,
   e.g. in case of a general overload situation.
 
-* A batch or bulk response with status code {207} *always* returns as payload
+* A batch or bulk response with status code {207} as payload *always* returns
   a multi-status object containing sufficient status and/or monitoring
   information for each part of the batch or bulk request.
 
@@ -259,12 +259,12 @@ Nakadi's https://nakadi.io/manual.html#/event-types/name/events_post[`POST
 want to avoid  risks and expect clients to handle partial
 batch failures anyway.
 
-A proper bulk or batch response may look as follows:
+The bulk or batch response may look as follows:
 
 [source,yaml]
 ----
-Response:
-  description: standard batch response object.
+BatchOrBulkResponse:
+  description: batch response object.
   type: object
   properties:
     items:
@@ -284,7 +284,7 @@ Response:
           description:
             description: >
               Human readable status description and containing additional
-              failure context information.
+              context information about failures etc.
             type: string
         required: [id, status]
 ----
--- a/chapters/http-status-codes-and-errors.adoc
+++ b/chapters/http-status-codes-and-errors.adoc
@@ -231,12 +231,12 @@ about your request processing status or error situations.
 [#152]
 == {MUST} use code 207 for batch or bulk requests
 
-Some APIs are required to provide either _batch_ or _bulk_ requests
-using {POST} for performance reasons, i.e. for communication and
-processing efficiency. In this case services may be in need to signal
-multiple response codes for each part of an batch or bulk request. As
-HTTP does not provide proper guidance for handling batch/bulk requests
-and responses, we herewith define the following approach:
+Some APIs are required to provide either _batch_ or _bulk_ requests using
+{POST} for performance reasons, i.e. for communication and processing
+efficiency. In this case services may be in need to signal multiple response
+codes for each part of an batch or bulk request. As HTTP does not provide
+proper guidance for handling batch/bulk requests and responses, we herewith
+define the following approach:
 
 * A batch or bulk request *always* has to respond with HTTP status code
   {207}, unless it encounters a generic or unexpected failure before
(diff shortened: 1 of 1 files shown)


chapters/http-status-codes-and-errors.adoc
```diff
@@ -248,10 +248,44 @@ and responses, we herewith define the following approach:
   the service encounters a failure before looking at individual parts or,
   if an unanticipated failure occurs.
 
-The before rules apply _even in the case_ that processing of all
-individual part _fail_ or each part is executed _asynchronously_! They
-are intended to allow clients to act on batch and bulk responses by
-inspecting the individual results in a consistent way.
+The before rules apply _even in the case_ that processing of all individual
+parts _fail_ or each part is executed _asynchronously_! They are intended
```
whiskeysierra: > They are intended to

> This rules are intended to

chapters/http-status-codes-and-errors.adoc
```diff
@@ -248,10 +248,44 @@ and responses, we herewith define the following approach:
   the service encounters a failure before looking at individual parts or,
   if an unanticipated failure occurs.
```
tfrauenstein: editorial: propose to make it a bit more crispy:

* A batch request may return {4xx}/{5xx} status code, *only* if it is not related to the individual items of the batch (e.g. general server overload situation). Otherwise the batch request should *always* return HTTP status code {207}.

* A batch or bulk response with status code {207} *always* returns as payload a  multi-status object containing sufficient status and/or monitoring information for each item of the batch request.

chapters/http-status-codes-and-errors.adoc
```diff
@@ -248,10 +248,44 @@ and responses, we herewith define the following approach:
...
 
-The before rules apply _even in the case_ that processing of all
-individual part _fail_ or each part is executed _asynchronously_! They
-are intended to allow clients to act on batch and bulk responses by
-inspecting the individual results in a consistent way.
+The before rules apply _even in the case_ that processing of all individual
```
tfrauenstein: The before rules --> These rules

chapters/http-status-codes-and-errors.adoc
```diff
@@ -248,10 +248,44 @@ and responses, we herewith define the following approach:
...
+parts _fail_ or each part is executed _asynchronously_! They are intended
+to allow clients to act on batch and bulk responses in a consistent way by
+inspecting the individual results. This rules are intended to prevent
+misinterpretation base on global response codes, i.e. we explicitly reject
+the option to apply {200} for a completely successful batch as proposed in
```
tfrauenstein: This rules are intended to prevent misinterpretation base on global response codes, i.e. we ....
--> 
We ...

chapters/http-status-codes-and-errors.adoc
```diff
@@ -248,10 +248,44 @@ and responses, we herewith define the following approach:
...
+Nakadi's https://nakadi.io/manual.html#/event-types/name/events_post[`POST
+/event-types/{name}/events`] as short cut without inspecting the result.
```
tfrauenstein: ... because, we want to avoid misinterpretation risks and expect clients to anyway handle partial batch failures,

chapters/http-status-codes-and-errors.adoc
tkrop: I have tried to incorporate your improvements, except I'm not convinced that the general `207` rule should be subsequent to the `4xx/5xx` rule. Thus I have pushed it to the top.

tfrauenstein:
👍

chapters/http-status-codes-and-errors.adoc
```diff
@@ -231,27 +231,63 @@ about your request processing status or error situations.
...
+  properties:
+    items:
+      type: array
+      items:
+        type: object
+        properties:
+          id:
+            description: Identifier of batch or bulk request item.
+            type: string
+          status:
```
vadeg: Should we restrict it to HTTP status codes or custom status can be provided?

chapters/http-status-codes-and-errors.adoc
```diff
@@ -231,27 +231,63 @@ about your request processing status or error situations.
...
+Nakadi's https://nakadi.io/manual.html#/event-types/name/events_post[`POST
+/event-types/{name}/events`] as short cut without inspecting the result, as we
+want to avoid  risks and expect clients to handle partial
+batch failures anyway.
+
+A proper bulk or batch response may look as follows:
+
+[source,yaml]
+----
+Response:
```
vadeg: I am not sure up to which level we want to support `207` but [specification](https://tools.ietf.org/html/rfc4918#section-13) says
> Each 'response' element MUST have an 'href' element to identify the resource.

I checked some of our API's and have not found any that support this.

chapters/http-status-codes-and-errors.adoc
tkrop: I vote for custom status codes. That's why I suggested to represent the status with an (x-)enum.

chapters/http-status-codes-and-errors.adoc
tkrop: We reuse 207, however, we do not implement WebDAV. So we took the freedom to redefine the format to provide an identifier pointing to the request. This is how we can observe the current practice. A `href` would not be practically, I think.

tfrauenstein:
👍

tkrop:
@tfrauenstein @vadeg @ePaul Thanks for the feedback and discussion. I hope, I have addressed all your great suggestions. Please have a final look and give you consent.

tfrauenstein:
👍

tkrop:
:+1:

whiskeysierra:
:+1:

PR #564, opened 2020-01-27
feat: allow compound keys in identifiers (#557)

### LINKED ISSUE #557: Resources with compound keys are different than sub-resources
There are use cases where resources are identified via compound-keys, e.g. `article-size-advice/{sku}/{sales-channel}`. [Rule 143](https://opensource.zalando.com/restful-api-guidelines/#143) requires that a 'artificial sub-resource' is created, like `article-size-advice/{sku}/sales_channel/{sales-channel}`. This does not feel right: There is no `article-size-advice/{sku}` as an own resource with distinct life cycle. There are a couple of APIs that are not compliant to the rule or work around the rule (like above, or string concat). I vote for being less restrictive and allow `article-size-advice/{sku}/{sales-channel}` for resources with compound keys.

tkrop:
fixes #557.

I have written a suggestion how to open up the rules a bit with respect to _compound keys_.

### DIFF
--- a/chapters/resources.adoc
+++ b/chapters/resources.adoc
@@ -141,15 +141,27 @@ evolve the structure of the resource identifier as it is no longer opaque.
 
 To compensate for this drawback, APIs must apply a compound key abstraction
 consistently in all requests and responses parameters and attributes allowing
-consumer to treat these as _technical resource identifier_ replacement. The
+consumers to treat these as _technical resource identifier_ replacement. The
 use of independent compound key components must be limited to search and
-creation requests, e.g.:
+creation requests, as follows:
 
 [source,http]
 ----
-GET /article-size-advices?skus=id-1,id-2,id-3&sales_channel_id=sid-1
-POST /shopping-carts { "country": "DE", "session_id": "id-1" }
+# compound key components passed as independent search query parameters
+GET /article-size-advices?skus=sku-1,sku-2&sales_channel_id=sid-1
+=> { "items": [{ "id": "id-1", ...  },{ "id": "id-2", ...  }] }
+
+# opaque technical resource identifier passed as path parameter
+GET /article-size-advices/id-1
+=> { "id": "id-1", "sku": "sku-1", "sales_channel_id": "sid-1", "size": ... }
+
+# compound key components passed as mandatory request fields
+POST /article-size-advices { "sku": "sku-1", "sales_channel_id": "sid-1", "size": ... }
+=> { "id": "id-1", "sku": "sku-1", "sales_channel_id": "sid-1", "size": ... }
 ----
+
+Where `id-1` is representing the opaque provision of the compound key
+`sku-1/sid-1` as technical resource identifier.
 
 **Remark:** A compound key component may itself be used as another resource
 identifier providing another resource endpoint, e.g `/article-size-advices/{sku}`.
--- a/chapters/resources.adoc
+++ b/chapters/resources.adoc
@@ -72,9 +72,11 @@ which business object it represents. Along these lines, "items" is too general.
 == {MUST} use URL-friendly resource identifiers: [a-zA-Z0-9:._\-/]*
 
 To simplify encoding of resource IDs in URLs, their representation must only
-consist of ASCII strings of letters, numbers, underscore, minus, colon, and
-period. **Note:** slashes are only allowed to build and signal resource
-identifiers consisting of <<241, compound keys>>.
+consist of ASCII strings using letters, numbers, underscore, minus, colon,
+period, and - on rare occasions - slash.
+
+**Note:** slashes are only allowed to build and signal resource identifiers
+consisting of <<241, compound keys>>.
 
 
 [#143]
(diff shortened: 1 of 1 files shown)


tfrauenstein:
Thank you for the proposal -- I like it. 
I think it makes sense to also mention an alternative design approach via additional 'technical key'.
Here an initial draft: 

You may consider to abstract from compound keys as resource id. Exposing the compound key as resource id has the disadvantage that the id structure is not opaque and all clients must know the id structure, even if they e.g. only want to pass the id information.  The abstraction can be done via a 'technical key' that is provided as additional resource attribute and used in the resource paths (as primary id), whereas the compound key elements are handled as secondary (compound) key resource attributes provided via resource payload for POST or via query parameters for GET.

tkrop:
@tfrauenstein I have a look how to address this idea - seems as I have formulated this concept too implicit.

chapters/resources.adoc
```diff
@@ -82,36 +83,98 @@ period.
...
+If a resource is best described by a _compound key_ consisting of multiple
+other resource identifiers, it is allowed to reuse the compound key in its
+natural form containing also slashes as resource identifier, e.g. for Docker
+images the compound key may look like `{repository-name}/{artifact-name}:{tag}`.
+
+However, resources using a compound key are expected to apply it consistently
+for all requests and responses as primary identifier. The stand alone usage
+of the compound key components must be secondary and limited to search
+requests only. For a consumer it must be - in general - indistinguishable
+whether it is invoking an endpoint with compound key normal resource
```
whiskeysierra:
```suggestion
whether it is invoking an endpoint with compound key or a normal resource
```

chapters/resources.adoc
```diff
@@ -82,36 +83,98 @@ period.
...
+----
+/shopping-carts/de/1681e6b88ec1/items/1
+/shopping-carts/de/1681e6b88ec1
+/api-specifications/{repository-name}/{artifact-name}:{tag}/apis/{path}/{file-name}
+/api-specifications/{repository-name}/{artifact-name}:{tag}
+/article-size-advices/{sku}/{sales-channel}
+/article-size-advices/{sku}
+----
+
+**Remark:** A compound key component may itself be used as another resource
```
whiskeysierra: I'd prefer to have an individual remark/paragraph highlighting that exposing a compound key will limit an API provider's ability to evolve the key structure in the future.

chapters/resources.adoc
```diff
@@ -82,36 +83,98 @@ period.
...
+/resources/{compound-key-1}[delim-1]...[delim-n-1]{compound-key-n}
+----
+
+Paths may look as follows:
+
+[source,http]
+----
+/shopping-carts/de/1681e6b88ec1/items/1
+/shopping-carts/de/1681e6b88ec1
+/api-specifications/{repository-name}/{artifact-name}:{tag}/apis/{path}/{file-name}
```
ePaul: I would expect here (from reading the text above): 
```suggestion
/api-specifications/{docker-image-id}/apis/{path}/{file-name}
```
where docker-image is composed.

chapters/resources.adoc
tkrop: Added to the begin.

tkrop:
@tfrauenstein @whiskeysierra @ePaul @vadeg Thanks for the good feedback and fruitful discussion. Please have a final look and give your consent - if possible.

chapters/resources.adoc
```diff
@@ -82,36 +85,102 @@ period.
...
+
+**Note:** resource identifiers may be build of multiple other resource
+identifiers (see <<241>>).
+
+
+[#241]
+== {MAY} expose compound keys as resource identifiers
+
+**Warning:** Exposing a compound key as described below limits ability to
+evolve the structure of the resource identifier as it is not any longer
```
whiskeysierra: I'd say either 

> as it is not opaque any longer

or

> as it is no longer opaque

Maybe some native speaker can give a hint.

edchapel: > as it is no longer opaque.

This is the best and most correct option, IMO.

chapters/resources.adoc
```diff
@@ -82,36 +85,101 @@ period.
...
+To compensate for this drawback, APIs must apply a compound key abstraction
+consistently in all requests and responses parameters and attributes allowing
+consumer to treat these as _technical resource identifier_ replacement. The
+use of independent compound key components must be limited to search and
+creation requests, e.g.:
+
+[source,http]
+----
+GET /article-size-advices?skus=id-1,id-2,id-3&sales_channel_id=sid-1
+POST /shopping-carts { "country": "DE", "session_id": "id-1" }
```
tfrauenstein: I like the update. To make it a bit clearer, I propose to use the same resource for the method examples and explicitly mention the technical resource identifier:

GET /article-size-advices?sku=sku1&sales_channel_id=sid1  -- with compound key passed as query parameter
GET /article-size-advices/id1  -- with opaque technical resource identifier passed as path parameter

POST  /article-size-advices   { "sku": "sku1", "sales_channel_id": "sid1" }  -- with compound key passed as mandatory read-only fields

And POST returns the technical resource identifier as output field of the payload { "id": "id1", "sku": "sku1", "sales_channel_id": "sid1" }

tkrop: Good suggestion. I tried to improve on this approach.

tfrauenstein:
👍

tkrop:
:+1:

whiskeysierra:
:+1:

PR #565, opened 2020-01-27
feat: clarify deprecation and sunset header (#562)

### LINKED ISSUE #562: Clarify Deprecation Header usage
copied from API-Guild chat: Daniel Konczyk:

1) Hi, was actually going through these changes today and it would be great to get some clarification. https://opensource.zalando.com/restful-api-guidelines/#189 starts with During deprecation phase, the producer should add a Deprecation header [...] with the sunset date value. However https://tools.ietf.org/html/draft-dalal-deprecation-header-02 states that Deprecation header carries a deprecation date (past/future, aka 'deprecated since') or "true" value, while the actual sunset date should be sent through the Sunset header. Any reason we have it differently?

2) Additionally, what should happen if there are e.g. two fields deprecated at different times under the same uri - what date would we return in the Deprecation header?

3) Oh and one last thing - have you considered providing a specification extension for the actual date that should be set in the header, since there's nothing in the OpenAPI at the moment to support that and seems that proposals for additional fields (deprecation-related) were not accepted. We already support extensions (e.g. x-extensible-enum), so it could be useful to be able to set the deprecation date this way too?

tkrop:
fixes #562.

### DIFF
--- a/chapters/deprecation.adoc
+++ b/chapters/deprecation.adoc
@@ -59,11 +59,12 @@ During the deprecation phase, the producer should add a `Deprecation: <date-time
 Deprecation HTTP Header]) and a `Sunset: <date-time>` (see {RFC-8594}#section-3[RFC
 8594]) header on each response affected by a deprecated element (see <<187>>).
 
-The {Deprecation} header can either be set to `true` or carry a deprecation
-time stamp, at which a replacement will become/became available and further
-on-boarding is discouraged. The optional {Sunset} time stamp carries the
-information when consumers latest have to switch to the replacement solution,
-and should always offer an eligible time interval for switching.
+The {Deprecation} header can either be set to `true` - if a feature is retired
+-, or carry a deprecation time stamp, at which a replacement will become/became
+available and consumers must not on-board any longer (see <<191>>). The optional
+{Sunset} time stamp carries the information when consumers latest have to stop
+using a feature and switch to a replacement solution. The sunset date should
+always offer an eligible time interval for switching.
 
 [source,txt]
 ----
@@ -89,8 +90,10 @@ Caching], and our syntax was not compliant with {RFC-7234}#page-29[RFC 7234
 [#190]
 == {SHOULD} add monitoring for `Deprecation` and `Sunset` header
 
-Clients should monitor the {Deprecation} and {Sunset} headers in HTTP
-responses to get information about future sunset of API features (see <<189>>).
+Clients should monitor the {Deprecation} and {Sunset} headers in HTTP responses
+to get information about future sunset of APIs and API features (see <<189>>).
+Based collected dates client owners should build alerts to ensure alignment
+with service owners on required migration task.
 
 *Hint:* In earlier guideline versions, we used the `Warning` header to provide
 the deprecation info (see hint in <<189>>).
--- a/chapters/deprecation.adoc
+++ b/chapters/deprecation.adoc
@@ -46,33 +46,41 @@ supported by links or references to more details.
 [#188]
 == {MUST} monitor usage of deprecated APIs and API endpoints
 
-Owners of APIs used in production must monitor usage of deprecated APIs
-until the API can be shut down in order to align deprecation and avoid
-uncontrolled breaking effects. See also the <<193>>.
+Owners of APIs used in production must monitor usage of deprecated APIs until
+the API can be shut down in order to align deprecation and avoid uncontrolled
+breaking effects. See also <<193>>.
 
 
 [#189]
 == {SHOULD} add `Deprecation` and `Sunset` header to responses
 
-During the deprecation phase, the producer should add a `Deprecation: true | <date-time>`
+During the deprecation phase, the producer should add a `Deprecation: <date-time>`
 (see https://tools.ietf.org/html/draft-dalal-deprecation-header[draft: RFC
-Deprecation HTTP Header]) and a `Sunset: <date-time>` (See {RFC-8594}#section-3[RFC
+Deprecation HTTP Header]) and a `Sunset: <date-time>` (see {RFC-8594}#section-3[RFC
 8594]) header on each response affected by a deprecated element (see <<187>>).
-If multiple elements are deprecated the {Sunset} date is expected to be set to
-the earliest involved time stamp.
+
+The {Deprecation} header can either be set to `true` or carry a deprecation
+time stamp, at which a replacement will become/became available and further
+on-boarding is discouraged. The optional {Sunset} time stamp carries the
+information when consumers latest have to switch to the replacement solution,
+and should always offer an eligible time interval for switching.
 
 [source,txt]
 ----
-Deprecation: true
+Deprecation: Sun, 31 Dec 2024 23:59:59 GMT
 Sunset: Sun, 31 Dec 2025 23:59:59 GMT
 ----
 
-*Note:* adding the {Deprecation} and {Sunset} header is not sufficient to
-gain client consent to shut down an API.
+If multiple elements are deprecated the {Deprecation} and {Sunset} headers are
+expected to be set to the earliest time stamp to reflect the shortest interval
+consumers are expected to get active.
 
-*Hint:* In earlier guideline versions, we used the `Warning` header to
-provide the deprecation info to clients. However, `Warning` header has a less
-specific semantics, will be obsolete with 
+*Note:* adding the {Deprecation} and {Sunset} header is not sufficient to gain
+client consent to shut down an API or feature.
+
+*Hint:* In earlier guideline versions, we used the `Warning` header to provide
+the deprecation info to clients. However, `Warning` header has a less specific
+semantics, will be obsolete with 
 https://tools.ietf.org/html/draft-ietf-httpbis-cache-06[draft: RFC HTTP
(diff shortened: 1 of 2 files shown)


chapters/deprecation.adoc
```diff
@@ -54,32 +52,38 @@ uncontrolled breaking effects. See also the <<193>>.
...
-== {SHOULD} add a deprecation header to responses
+== {SHOULD} add `Deprecation` and `Sunset` header to responses
 
-During deprecation phase, the producer should add a `Deprecation` header 
-(see
-https://tools.ietf.org/html/draft-dalal-deprecation-header[draft: RFC Deprecation HTTP Header])
-with the sunset date value (in {RFC-7231}#section-7.1.1.1[HTTP Date/Time] form): 
+During the deprecation phase, the producer should add a `Deprecation: true`
+(see https://tools.ietf.org/html/draft-dalal-deprecation-header[draft: RFC
+Deprecation HTTP Header]) and a `Sunset: <data>` (See {RFC-8594}#section-3[RFC
```
konczyk: Probably a typo in <data>?

chapters/deprecation.adoc
```diff
@@ -54,32 +52,38 @@ uncontrolled breaking effects. See also the <<193>>.
...
+8594]) header on each response affected by a deprecated element (see <<187>>).
 
 [source,txt]
 ----
-Deprecation: Sun, 11 Nov 2018 23:59:59 GMT
+Deprecation: true
+Sunset: Sun, 31 Dec 2025 23:59:59 GMT
```
konczyk: So this change would mean we can't deprecate in advance. Any reason we can't accept true/date for deprecation, same as the RFC?

index.adoc
```diff
@@ -187,6 +188,9 @@
 :Prefer: pass:[<a href="https://tools.ietf.org/html/rfc7240"><code>Prefer</code></a>]
 :Preference-Applied: pass:[<a href="https://tools.ietf.org/html/rfc7240#section-3"><code>Preference-Applied</code></a>]
 
+:Sunset: pass:[<a href="https://tools.ietf.org/html/rfc8594"><code>Sunset</code></a>]
+:Deprecation: pass:[<a href="https://tools.ietf.org/html/draft-dalal-deprecation-header"><code>Deprectated</code></a>]
```
konczyk: Should be Deprecation int the <code>?

konczyk:
I think there is still an issue when we have multiple deprecations (added over time) under the same endpoint. If subsequent deprecations have the same or later sunset date than the previous one(s), client wouldn't see any difference in the deprecation response and may wrongly assume, that nothing has changed wrt deprecations since he checked last time.
Not sure what's the best solution to that except limiting deprecations to one per endpoint at a time. Surely we don't want clients to inspect the API spec all the time?

In any case, I think could be addressed in the guidelines as well.

chapters/deprecation.adoc
tkrop: Just my idea to simplify life. We can go for date too, if we have a real use case.

chapters/deprecation.adoc
```diff
@@ -54,32 +52,38 @@ uncontrolled breaking effects. See also the <<193>>.
...
 [#189]
-== {SHOULD} add a deprecation header to responses
+== {SHOULD} add `Deprecation` and `Sunset` header to responses
 
-During deprecation phase, the producer should add a `Deprecation` header 
-(see
-https://tools.ietf.org/html/draft-dalal-deprecation-header[draft: RFC Deprecation HTTP Header])
-with the sunset date value (in {RFC-7231}#section-7.1.1.1[HTTP Date/Time] form): 
+During the deprecation phase, the producer should add a `Deprecation: true`
+(see https://tools.ietf.org/html/draft-dalal-deprecation-header[draft: RFC
```
vadeg: Link does not work. I guess it should be https://tools.ietf.org/id/draft-dalal-deprecation-header-01.html

index.adoc
vadeg: Link should be https://tools.ietf.org/id/draft-dalal-deprecation-header-01.html

chapters/deprecation.adoc
tfrauenstein: I think we clearly should differ between deprecation (with true or deprecation-date values) and sunset -- as done by the two RFCs.  And we should mention that sunset should go with deprecation (but not vice versa) with  deprecation-date <= sunset-date.

chapters/deprecation.adoc
```diff
@@ -54,32 +52,40 @@ uncontrolled breaking effects. See also the <<193>>.
...
 [source,txt]
 ----
-Deprecation: Sun, 11 Nov 2018 23:59:59 GMT
+Deprecation: true
+Sunset: Sun, 31 Dec 2025 23:59:59 GMT
 ----
 
-Note, adding the `Deprecation` header is not sufficient to gain client consent to shut down an API.
+*Note:* adding the {Deprecation} and {Sunset} header is not sufficient to
+gain client consent to shut down an API.
```
tfrauenstein: Let as also provide a hint, how these headers can be used to fulfill the monitoring requirement above with our OpenTracing infrastructure service.

chapters/deprecation.adoc
tkrop: for me it is working and redirecting to the current version https://tools.ietf.org/html/draft-dalal-deprecation-header-02.

index.adoc
tkrop: see above.

chapters/deprecation.adoc
tkrop: Difficult aspect. I'm not sure how this should best look like:

* Suggest to add a deprecation-tag to OpenTracing.
* Suggest to install filters in our routing infrastructure to count `Deprecation: *` headers.

We may even provide this within our api-monitoring infrastructure, so that owner only have to create alerts on this.

chapters/deprecation.adoc
tkrop: Did a slightly different approach to address this.

chapters/deprecation.adoc
```diff
@@ -46,33 +46,41 @@ supported by links or references to more details.
...
 (see https://tools.ietf.org/html/draft-dalal-deprecation-header[draft: RFC
-Deprecation HTTP Header]) and a `Sunset: <date-time>` (See {RFC-8594}#section-3[RFC
+Deprecation HTTP Header]) and a `Sunset: <date-time>` (see {RFC-8594}#section-3[RFC
 8594]) header on each response affected by a deprecated element (see <<187>>).
-If multiple elements are deprecated the {Sunset} date is expected to be set to
-the earliest involved time stamp.
+
+The {Deprecation} header can either be set to `true` or carry a deprecation
+time stamp, at which a replacement will become/became available and further
+on-boarding is discouraged. The optional {Sunset} time stamp carries the
```
conorclifford: "discouraged" - elsewhere in the API Guidelines there is specific mention that [clients *must* not start using `deprecated` features](https://opensource.zalando.com/restful-api-guidelines/#191) - this should be aligned with that IMO.

chapters/deprecation.adoc
konczyk: This last changed removed `true` from the Deprecation header possible values, but still mentions it in the description above. Also, I think it was decided earlier that `Sunset` is not optional if something is deprecated? Previous paragraph mentions both headers the producer should add, new paragraph says optional".

IMHO I think this new description paragraph is a bit unclear regarding the intent of both headers. Say, `Sunset` date is defined in the RFC as a date after which deprecated resource is expected to become unresponsive, but we don't point it out explicitly in the description.

chapters/deprecation.adoc
tkrop: Decided to not address in this issue. Needs further discussion in the guild.

chapters/deprecation.adoc
tkrop: @konczyk thanks for the great feedback.

I described `Sunset` to be optional because it might not have been decided when to sunset a deprecated feature - majorly because this requires consent and alignment with consumers according to [Rule 185](https://opensource.zalando.com/restful-api-guidelines/#185).

I removed `true`, to not overemphasize this option. I see `true` as an exception, that should only be used, if no replacement will be provided. In this case there will be no availability date for the replacement and it, thus, does not make sense to set one in the `Deprecation` header. I hope the text now clearly explains this as short as possible.

tkrop:
@konczyk @tfrauenstein @vadeg and @conorclifford thanks for the great feedback. I hope, I have now addressed all aspects in the PR. Please have a final look and - if possible - approve.

chapters/deprecation.adoc
konczyk: > I described `Sunset` to be optional because it might not have been decided when to sunset a deprecated feature - majorly because this requires consent and alignment with consumers according to [Rule 185](https://opensource.zalando.com/restful-api-guidelines/#185).

I see, that's a fair point (I think we have gone full circle on that though 😄). Was just thinking that `{SHOULD} add `Deprecation` and `Sunset` header to responses` might suggest that we should use both (while optionality of `Sunset` is "hidden" in the details), but maybe I'm just too picky.

tkrop:
After a thorough proof read, by @tfrauenstein I extended this PR to include further consistency changes of the deprecation guideline.

tfrauenstein:
👍

tkrop:
:+1:

PR #586, opened 2020-06-02
Aligned permissions examples with rule 255

fatroom:
The rule 255 ({MUST} follow naming convention for permissions) states that pattern for resource part of permission should follow `[a-z][a-z0-9-]*` pattern, but resource scopes in example are not aligned to this pattern.


tfrauenstein:
👍

tkrop:
:+1:

PR #600, opened 2020-08-17
feat: add guidance on path normalization (#571)

### LINKED ISSUE #571: Improve guidance on path acceptance/normalization
When looking on how services usually handle variations of paths, we can see big differences between different frameworks and service implementations, which is not reflected in the API specification but has a negative impact on predictability of the behavior of an API and the way how it is or can be used. The keyword here is request path normalization, i.e.

1. How are adjacent slashes in a request are handled, and
2. How are trailing and leading slashes in a request are handled.

While we already have a rule for trailing slashes [#136](https://opensource.zalando.com/restful-api-guidelines/#136), we are missing more guidance how adjacent slashes should be handled - and subsequently, whether empty path variable values should be permitted.

**Note:** leading slashes should not be an issue as they are taken care by the HTTP request specification.

To create a consistent behavior of APIs and allow for efficient high quality API monitoring we should improve our guidance on path acceptance/normalization as follows:

1. Services MUST normalize request paths before matching, i.e.
1.1. Multiple adjacent slashes do not change the semantic and must be treated as a single slash,
1.2. Trailing slashes do not change the semantic and must be treated as if absent.
2. Consequently, empty strings as path parameters are not allowed, as these would conflict with the normalization requirement.

tkrop:
fixes #571.


chapters/naming.adoc
```diff
@@ -151,10 +151,24 @@ https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#serve
 
 
 [#136]
-== {MUST} avoid trailing slashes
+== {MUST} normalize paths by removing duplicate and trailing slashes
```
tfrauenstein, line 154: As discussed, we should strictly separate specification from implementation aspects.
Here a concrete PR update proposal:
 
== {MUST} use standard paths without empty path segments
You must not use paths with duplicate or trailing slashes.
As a consequence, you must not use path variables with possibly empty values. 
Examples for _invalid_ path definitions:
 
[source,http]
----
GET /customers/
GET /customers//addresses
GET /customers/{customer-id}  -- if empty {customer-id} value is allowed
---

Reasoning: Non standard paths have no clear semantics. They are not supported by all HTTP infrastructure components or libraries and create ambiguities, which may lead, for instance, to uncertainties in API monitoring since the association of traffic to the specified endpoints is not unique. 

With respect to service implementation we recommend to provide the following robust behavior once clients do not follow the rule: Services SHOULD 'normalize' invalid paths and ignore trailing or
duplicate slashes in the request path. Hence, the following requests should refer to the same resource:
 
[source,http]
----
GET /orders/{order-id}
GET /orders/{order-id}/
GET /orders//{order-id}
----
 
**Note:** path normalization is not supported by all frameworks out-of-the-box.
API implementations are required to support at least the normalized path while
rejecting all alternatives requests, if failing to deliver the same resource.

tkrop: Yes, I can see this point of separating specification from implementation aspects. However, I'm not a fan of giving counter examples first. Readers tend to overlook the warning in front of the counter example and take it as best practice.

chapters/resources.adoc
```diff
@@ -78,6 +78,10 @@ period, and - on rare occasions - slash.
 **Note:** slashes are only allowed to build and signal resource identifiers
 consisting of <<241, compound keys>>.
 
+**Note:** to prevent ambiguities of <<136, unnormalized paths>> resource
```
vadeg, line 81: I think it make sense to mention this in 136 itself for visibility. Otherwise in 136 it is not clear whether duplicated trailing slashes are accepted or not.
 I like how @tfrauenstein formulated it above:
> You must not use paths with duplicate or trailing slashes.
As a consequence, you must not use path variables with possibly empty values.

chapters/naming.adoc
tfrauenstein: @tkrop ok, updated the proposal to avoid this risk.

chapters/resources.adoc
tkrop: Does this proposal include to remove this mentioning here?

tkrop:
@tfrauenstein @vadeg thanks for the helpful comments. I have update the PR based on your suggestions.

otaciliolacerda:
👍

tfrauenstein:
👍

chapters/resources.adoc
tfrauenstein: no, it is worth to be also mentioned here.

tfrauenstein:
👍

chapters/resources.adoc
vadeg: I think having a reference here to 136 with a comment, e.g. "resource identifiers should follow rules defined in 136` would be enough.

tkrop:
:+1:

vadeg:
👍

PR #601, opened 2020-08-18
feat: adapt problem specification to guideline (#581)

### LINKED ISSUE #581: Adapt OpenAPI specification of application/problem+json
Since #561 we have changed our advice how to use [application/problem+json](https://opensource.zalando.com/restful-api-guidelines/#176) (now the type is not supposed to be dereference-able), however, we have not adapted our [global specification](https://zalando.github.io/problem/schema.yaml).

Should we do compatible description update?

tkrop:
fixes #581.

To complete the guideline change, I also propose that the `instance` field is changed to support a relative URI. In addition, I suggest to turn around and shorten the argumentation a bit.

Finally, I also suggest to remove the compatibility remark, that is now referencing the outdated state for more than 4 years and should not be relevant any longer.

I think all changes are somehow compatible or editorial, however, do we need to add a note, that absolute URIs are not forbidden?


tfrauenstein:
thanks for the PR!

> To complete the guideline change, I also propose that the `instance` field is changed to support a relative URI. 

yes, makes sense. But pls. document relative URI usage and relative to what.

> Finally, I also suggest to remove the compatibility remark, that is now referencing the outdated state for more than 4 years and should not be relevant any longer.

makes sense

> I think all changes are somehow compatible or editorial, however, do we need to add a note, that absolute URIs are not forbidden?

+1

tkrop:
`relative (URI) reference` is a defined term in the RFC. To make this more clear I have added the definition part, that we use here. As we do not want to encourage users to make use of the reference resolution strategy, I think it is not wise to refer here to the base URI of the service that build the anchor of this "relative" reference.

Adding a note that we do not forbid the usage of `absolute URIs` I felt it necessary to remove the now contradicting counter examples to keep the guideline-change really compatible to our original guidance.

models/problem-1.0.1.yaml
```diff
@@ -0,0 +1,46 @@
+Problem:
+  type: object
```
vadeg, line 2: I would add here a description which fields are required, e.g.
```
required: [type]
```
because according [to RFC](https://tools.ietf.org/html/rfc7807#section-3.1)  only `type` field _MUST_ present.

models/problem-1.0.1.yaml
```diff
@@ -0,0 +1,46 @@
...
+  properties:
+    type:
+      type: string
+      format: uri
+      description: >
+        A relative URI reference that uniquely identifies the problem type.
```
tfrauenstein, line 8: a) A relative URI reference that ... --> A URI reference (either URI or relative reference) that...
It now reads that we only allow 'relative references' as type:uri values which is not intended, is it?

b) problem type --> problem type (in context of the problem definition).
In contrast to the always unique URI, the relative reference is only unique in the definition context.

models/problem-1.0.1.yaml
```diff
@@ -0,0 +1,46 @@
...
+        Opposed to the specification in RFC-7807, it is not required to be
```
tfrauenstein, line 9: it is not required --> it is recommended, but not required ...

chapters/http-status-codes-and-errors.adoc
```diff
@@ -359,35 +359,32 @@ responses:
     content:
       "application/problem+json":
         schema:
-          $ref: 'https://opensource.zalando.com/problem/schema.yaml#/Problem'
+          $ref: ''https://opensource.zalando.com/restful-api-guidelines/problem-1.0.1.yaml#/Problem'
```
otaciliolacerda, line 362: Typo (additional `'`)?

models/problem-1.0.1.yaml
tkrop: To my knowledge even `type` is optional. We would have loved to have such a thing to define a minimal constraint on manual problem definitions, but there are no required fields (see https://tools.ietf.org/html/rfc7807#section-6.1).

models/problem-1.0.1.yaml
tkrop: The absolute URI with scheme is exactly what we want discouraged and reason for this PR. I would not bring it into this standard object specification. If one intends to use absolute URIs (s)he should stick to the old model description which is uploaded as `models/problem-1.0.0.yaml`.

tkrop: I have incorporated (b) and added a sentence for (a) that refers to the old specification.

models/problem-1.0.1.yaml
tkrop: Still, I must admit, that I'm open to require even more fields. In my opinion each problem should at least make use of `type`, `title`. If you really, want to use problems for error analysis, you even should make use of `instance` and `description` to transport some context that allows to locate the root cause of the problem and gives advice on how to proceed.

tkrop:
@otaciliolacerda @tfrauenstein @vadeg thanks for the comments. Please have a second look.

tfrauenstein:
👍

otaciliolacerda:
👍

tkrop:
:+1:

Issue #595, opened 2020-07-29
Allow `self` in Urls /customers/self


raiskumar:
The [API guideline](https://opensource.zalando.com/restful-api-guidelines/#134) has a MUST rule to always use plural resources. In some cases, we use adjectives like `self` to refer to the customer whose `Authorization` token is passed in the request (header).

```
GET /customers/self/personal-details
```
Pluralizing `self` to comply with guidelines or using `{customer-number}` doesn't look good. The `self` version is more readable and many customer-facing services are already using it. 
Can we allow `self` for cases like above in `zally` and add it to API guidelines. 

Discussion thread: https://chat.google.com/room/AAAAAqZDJ28/WSrq8o4lHTE

ePaul:
*(Discussion in the guild meeting.)*

That's what we have currently:

* `/customers` → all customers
* `/customers/{id}` → one specific customer with this id
* `/customers/4711` → the customer with id 4711
* `/customers/{id}/personal-details` → personal details of this one customer
* `/customers/4711/personal-details` → personal details of the customer with id 4711.

Use-case: we want to fetch details for one specific customer (identified by context information, e.g. the authorization token, not the URL). So we don't have an id to provide, and providing one would also not be supported.

Proposed in this issue (and already used by some teams):

* `/customers/self/personal-details`, which is currently not allowed by rule 143 (alternate const + param segments) + 134 (plural), and gives Zally-violations.

Viewed as sequence of sub-resources:

* `/customers` – all customers
* `/customers/self` – the customer identified by the current token
* `/customers/self/personal-details` the personal details of the customer identified by the current token

`self` is here a pseudo-identifier, it takes the syntactic position where in rule 143 an identifier would be expected.
Alternative proposed names: `me`, `from-token`, ...

Workaround which is currently (by letter) allowed by Rule 143:

* Have an endpoint `/customers/{self_enum}/personal-details`, and define `self_enum` as a path parameter with `enum: ["self"]` or similar.
* This doesn't give any violations
* fits to the previous implementation
* if used with code generations, you get an additional parameter here, which needs to be provided.
* It is confusing when reading the API.

**Decision**:
We want to extend rule 143 to allow the replacement of up to one occurrence of the `{resource-id}` parameter with the direct path component `/self/`.

PR #603, opened 2020-08-21
'self' as pseudo-identifier in resource paths

tfrauenstein:
fixes https://github.com/zalando/restful-api-guidelines/issues/595

### DIFF
--- a/chapters/naming.adoc
+++ b/chapters/naming.adoc
@@ -129,9 +129,11 @@ on HTTP headers.
 == {MUST} pluralize resource names
 
 Usually, a collection of resource instances is provided (at least API
-should be ready here). The special case of a resource singleton is a
-collection with cardinality 1.
+should be ready here). The special case of a _resource singleton_ is a
+collection with cardinality 1 --  to be highlighted in the resource 
+definition description.
 
+**Exception:** The 'pseudo identifier' `self` is the only exception --  see <<143>>.
 
 [#135]
 == {SHOULD} not use /api as base path
--- a/chapters/naming.adoc
+++ b/chapters/naming.adoc
@@ -135,7 +135,7 @@ be modeled as a collection with cardinality 1 including definition of
 to make the cardinality constraint explicit.
 
 **Exception:** the _pseudo identifier_ `self` used to specify a resource endpoint 
-where the resource identifier provided by authorization information (see <<143>>).
+where the resource identifier is provided by authorization information (see <<143>>).
 
 
 [#135]
--- a/chapters/naming.adoc
+++ b/chapters/naming.adoc
@@ -128,12 +128,15 @@ on HTTP headers.
 [#134]
 == {MUST} pluralize resource names
 
-Usually, a collection of resource instances is provided (at least API
-should be ready here). The special case of a _resource singleton_ is a
-collection with cardinality 1 --  to be highlighted in the resource 
-definition description.
+Usually, a collection of resource instances is provided (at least the API 
+should be ready here). The special case of a _resource singleton_ must 
+be modeled as a collection with cardinality 1 including definition of 
+`maxItems` = `minItems` = 1 for the returned `array` structure 
+to make the cardinality constraint explicit.
 
-**Exception:** The 'pseudo identifier' `self` is the only exception --  see <<143>>.
+**Exception:** the _pseudo identifier_ `self` used to specify a resource endpoint 
+where the resource identifier provided by authorization information (see <<143>>).
+
 
 [#135]
 == {SHOULD} not use /api as base path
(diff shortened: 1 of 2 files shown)


chapters/resources.adoc
```diff
@@ -85,7 +85,7 @@ consisting of <<241, compound keys>>.
 Some API resources may contain or reference sub-resources. Embedded
 sub-resources, which are not top-level resources, are parts of a higher-level
 resource and cannot be used outside of its scope. Sub-resources should be
-referenced by their name and identifier in the path segments as follows:
+referenced by their identifier and name in the path segments as follows:
```
tkrop, line 88: Not sure about this change: Logical order (of appearance)  is sub-resource name and identifier, not reverse... but may be I'm to literal.

chapters/resources.adoc
```diff
@@ -94,22 +94,28 @@ referenced by their name and identifier in the path segments as follows:
...
-/customers/12ev123bv12v
 /shopping-carts/de:1681e6b88ec1/items/1
 /shopping-carts/de:1681e6b88ec1
 /content/images/9cacb4d8
 /content/images
 ----
 
 **Note:** resource identifiers may be build of multiple other resource
-identifiers (see <<241>>).
+identifiers -- see <<241>>.
```
tkrop, line 110: We should not inconsistently start to change reference patterns from `(see ...)` to `-- see ...`.

chapters/resources.adoc
```diff
@@ -94,22 +94,28 @@ referenced by their name and identifier in the path segments as follows:
...
+
+**Exception:** In some situations the id information is not passed
```
tkrop, line 112: May I suggest to substitute `the id information` against `the resource identifier`? Else I think it should be `the ID information`.

chapters/resources.adoc
```diff
@@ -94,22 +94,28 @@ referenced by their name and identifier in the path segments as follows:
...
 /content/images/9cacb4d8
 /content/images
 ----
 
 **Note:** resource identifiers may be build of multiple other resource
-identifiers (see <<241>>).
+identifiers -- see <<241>>.
+
+**Exception:** In some situations the id information is not passed 
+as a path segment but via context information, e.g. the authorization token.
```
tkrop, line 113: Should we really be open to all context information? `self` in my opinion is restricted to `authorization information`. So may be we should just write: `but via authorization information, e.g. an authorization token or session cookie.`

chapters/naming.adoc
```diff
@@ -129,9 +129,11 @@ on HTTP headers.
 == {MUST} pluralize resource names
 
 Usually, a collection of resource instances is provided (at least API
-should be ready here). The special case of a resource singleton is a
-collection with cardinality 1.
+should be ready here). The special case of a _resource singleton_ is a
+collection with cardinality 1 --  to be highlighted in the resource
```
tkrop, line 133: Not sure what to be highlighted in the specification here? I think what we should say here is:

```
The special case of a _resource singleton_ must be modeled as a collection with cardinality 1.

**Exception:** the _pseudo identifier_ `self` used to specify a resource endpoint where the
resource identifier provided by authorization information (see <<143>>).
```

chapters/resources.adoc
```diff
@@ -94,22 +94,28 @@ referenced by their name and identifier in the path segments as follows:
...
 ----
 
 **Note:** resource identifiers may be build of multiple other resource
-identifiers (see <<241>>).
+identifiers -- see <<241>>.
+
+**Exception:** In some situations the id information is not passed 
+as a path segment but via context information, e.g. the authorization token. 
+Here, it is reasonable to use **`self`** as _pseudo-identifier_ path segment. 
+For instance, you may define `/employee/self` or `/employee/self/personal-details`
```
tkrop, line 115: I think this should not be `/employee/self` but `/employees/self` to be consistent with rules and allow for a compatible update to `employees/{id}`.

chapters/naming.adoc
tfrauenstein: ok, and will add hint to explicitly define the cardinality constraint.

chapters/naming.adoc
```diff
@@ -128,9 +128,14 @@ on HTTP headers.
...
-should be ready here). The special case of a resource singleton is a
-collection with cardinality 1.
+Usually, a collection of resource instances is provided (at least the API 
+should be ready here). The special case of a _resource singleton_ must 
+be modeled as a collection with cardinality 1 including definition of 
+`maxItems` = `minItems` = 1 for the returned `array` structure 
+to make the cardinality constraint explicit.
+
+**Exception:** the _pseudo identifier_ `self` used to specify a resource endpoint 
+where the resource identifier provided by authorization information (see <<143>>).
```
tkrop, line 138: `provided` -> `is provided`

tfrauenstein: 👍

tkrop:
:+1:

ePaul:
:+1:

PR #606, opened 2020-08-25
clarify the UPPER_SNAKE_CASE rule for enums (#527)

ePaul:
* move rules 125 and 240 together (as they both are about enums)
* MUST → SHOULD for 240
* explain where the rule doesn't apply.


tkrop:
:+1:

ePaul:
:+1:

Issue #593, opened 2020-07-14
Add exception to rule 224 for merchant gateway APIs


ePaul:
As discussed in the guild meeting on 2020-07-14, the host name `api.merchants.zalando.com` (and `api-sandbox.merchants.zalando.com`) is used by several APIs exposed to merchant partners via the merchant gateway.
We want to extend rule 224 to allow this as an exception to the general *.zalandoapis.com rule (for external-partner APIs)

When the same services are used inside the company (not through the gateway), they should use "normal" naming schemes for their API.

Zally should then also be updated to allow those domain names.

rbarilani:
@ePaul is `api.merchants.zalando.com` and `api-sandbox.merchants.zalando.com` that should be whitelisted when the audience is `external-partner`

vadeg:
@rbarilani is it just a host name or you use functional names also? Do you have something like `<functional-name>.api.merchants.zalando.com`?

rbarilani:
@vadeg is just a host name

vadeg:
We missed the [Rule 223](https://opensource.zalando.com/restful-api-guidelines/index.html#223) where `external-partner` API's **MUST** use functional naming schema which is not true in this case.

rbarilani:
@vadeg sorry I don't get completely your comment, what is the action item that you are suggesting?

vadeg:
@rbarilani  Rule 223 says that `external-partner` must follow functional naming. Since this is no longer true and `external-partner` audience has an exception by only using two hostnames `api.merchants.zalando.com` and `api-sandbox.merchants.zalando.com` which violate rule 223 also my question is should we update rule 223 also by, for example, moving `external-partner` from _must_ to _should_?

rbarilani:
@vadeg IMHO we should not change that rule to a SHOULD but state the possibility to have exceptions which need to be backed up by business reasons and approved by the api-guild (If I remember correctly that was also the preferred solution discussed during the JF). 
[...]

PR #608, opened 2020-08-25
Functional naming exception for legacy hostnames

tfrauenstein:
fixes https://github.com/zalando/restful-api-guidelines/issues/593

Also adding support for `-` in functional domain names.


tkrop:
:+1:

tfrauenstein:
👍

PR #613, opened 2020-10-12
Don't put personal data into URLs

msooszalando:
To respect GDPR regulations, URLs must not contain any personal data, as URLs could potentially be logged.

### DIFF
--- a/chapters/security.adoc
+++ b/chapters/security.adoc
@@ -1,5 +1,5 @@
-[[security]]
-= Security
+[[security and privacy]]
+= Security and Privacy
 
 
 [#104]
@@ -170,5 +170,14 @@ permissions names of <<223, internal>> APIs:
 <application-id>      ::= [a-z][a-z0-9-]*  -- application identifier
 
 -----
-
 ////
+
+
+[#226]
+== {MUST} Not contain any personal data in the URL
+ 
+Personal data such as identifiers, names, addresses, order numbers, etc. as defined by GDPR
+regulations must not be part of any URL, either as part of the path or the GET parameter.
+URLs are sometimes logged for debugging, performance analysis, and other analytics purposes,
+which must not contain personal data. Instead of putting personal data in the URL, use the
+POST method to deliver the needed personal data to the endpoint.


ePaul:
Hi,

thanks for the contribution.

I'm not quite sure that we can merge this as-is (and whether it needs to be in the guidelines at all).
We want to first discuss this in the API guild meeting. As this week half of the group is on vacation, we put it on the agenda for next time. Feel free to come to the meeting then.

chapters/security.adoc
```diff
@@ -170,5 +170,14 @@ permissions names of <<223, internal>> APIs:
 <application-id>      ::= [a-z][a-z0-9-]*  -- application identifier
 
 -----
-
 ////
+
+
+[#226]
+== {MUST} Not contain any personal data in the URL
```
ePaul, line 177: The rule number 226 needs to be changed, as we already have [a rule with this number](https://opensource.zalando.com/restful-api-guidelines/#226).

Please see the [contributing guidelines](https://github.com/zalando/restful-api-guidelines/blob/master/CONTRIBUTING.adoc#rule-id) on how to find the next rule-ID.

(But that can be done after we've discussed the rule.)

msooszalando: Thanks! I created an issue, the script that's pointed to in the guidelines no longer exists:

https://github.com/zalando/restful-api-guidelines/issues/614

Once this has been rectified, I will immediately fix the number. Thanks,

Mate

tkrop:
As far as I remember we have discussed data privacy and GDPR in the API guild before.

The arguments to not add adding a restricting rule to the guideline was as follows:

RESTful APIs are based on resources IDs using URI/URLs. In case of resources that contain private information, the resource IDs would itself be seen as private information. In our opinion, It is not practical to create a layer of pseudonym URLs around resource URI/URLs.  Instead a way to mitigate the privacy problems should be applied:

1. Requests to APIs should generally be protected by authorization and encryption unless they stay cluster-internal. In our case all cluster-external requests are required to be protected by SSL and OAuth2, while cluster-internal access may not be authorized and encrypted.

2. Logging (and tracing) of URLs and private information is discouraged (not to be confused with security relevant information which is forbidden) but may be necessary for debugging purposes. To account for logging URLs and private information, access to log-storage is generally restricted and secured itself by authorization and encrypting transmissions. Tracing needs to either follow the same access restrictions as logging or safely filter out all private information.

@ePaul @tfrauenstein should we discuss this in the guild meeting again?

msooszalando:
Hi @tkrop ,

I believe we need to discuss this in-person. See [here for details](https://docs.google.com/document/d/1NahSuMC0LRwFTTrln2m0iDLq0nHzGmF59Gv7xBhKPfQ/edit#) and please tell us in case it doesn't answer your questions, we're happy to add/fix. Alternatively, if you have any other ideas how to do this, please let us know.

Thanks for your consideration,

Mate

msooszalando:
I am closing this, as it has been discussed that this would be against the REST API guidelines themselves given that resource IDs need to be present in many cases for things to be RESTful. However, we plan to open a pull request once we had the chance to align regarding only logging the request template rather than the request data (e.g. `/order/{oder_no}`) on default logging level.

Thanks again for your thoughts and feedback,

Mate

tkrop:
@msooszalando Thanks for taking this action. We where about to suggest exactly this.

We think hope that we can extend on our API operations guideline shortly. We have agreed on providing a privacy and GDPR conform advice to restrict logging and tracing to `path-templates` instead of distinct `paths` (with `query` parameters) and/or to automatically *blank out* sensitive information from `paths`, `parameters`, and `objects`.

Since this is best solved by infrastructure or library support base on API specs, we might also consider adding a custom flag for sensitive information, e.g. `x-sensitive: <true|false>` on properties to support automation and conclusion about privacy exposure in services APIs.

msooszalando:
Ah, thank you, that's an interesting idea, re: `x-sensitive: <true|false>`. I'll write up a narrative and ask you to contribute once I feel like it's ready to share :) Thanks again for that great conversation!

Mate

PR #615, opened 2020-10-27
fix(models): use correct format for type and instance in problem.json

ercanucan:
This PR includes the following:
    
- According to their descriptions, `type` and `instance` fields
are relative URI or URI references. According to RFC 3986, the formats
should be specified as `uri-reference` instead of `uri`:
    
    https://tools.ietf.org/html/rfc3986#section-4.1
    
    Otherwise, linter tools such as spectral (https://github.com/stoplightio/spectral)
    would fail to lint these models correctly.
    
    The PR adapts these fields accordingly.
    
    As suggested by the community, keeps the file version at 1.0.1.
    
- Also includes minor language fixes to the file.

### DIFF
--- a/models/problem-1.0.1.yaml
+++ b/models/problem-1.0.1.yaml
@@ -3,19 +3,18 @@ Problem:
   properties:
     type:
       type: string
-      format: uri
+      format: uri-reference
       description: >
-        A relative URI reference that uniquely identifies the problem type
-        only in the context of the provided API. Opposed to the specification
-        in RFC-7807, it is neither recommended to be dereferencable and point
-        to a human-readable documentation nor globally unique for the problem
-        type.
+        A URI reference that uniquely identifies the problem type only in the
+        context of the provided API. Opposed to the specification in RFC-7807,
+        it is neither recommended to be dereferencable and point to a
+        human-readable documentation nor globally unique for the problem type.
       default: 'about:blank'
       example: '/problem/connection-error'
     title:
       type: string
       description: >
-        A short, summary of the problem type. Written in english and readable
+        A short summary of the problem type. Written in English and readable
         for engineers, usually not suited for non technical stakeholders and
         not localized.
       example: Service Unavailable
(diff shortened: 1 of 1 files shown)


ercanucan:
@bufferoverflow

dlouzan:
@ercanucan Should this actually be `1.0.2` itself? 😅

ercanucan:
@dlouzan 

> @ercanucan Should this actually be `1.0.2` itself? 😅

@dlouzan I guess it could be, but the semantics do not change, it's more like a fix in my opinion. However I am open to see what the community and the maintainers suggest 🙂   /cc @tfrauenstein @gewald

tkrop:
Hi @ercanucan, many thanks for the contribution.

From a service compatibility point of view it would require even a version `2.0.0`, from client `1.0.2` may be sufficient. In any case the PR should have create at least a new `problem-1.0.2.yaml`

However, I'm not sure whether we should redefine it i this way. To stick to absolute paths is just a suggestion of the rule to account for the missing practical value in providing resolvable error descriptions outside of the spec. We do not intend to enforce this suggestion by tooling, so I suggest to leave it as is.

@tfrauenstein @gewald @ePaul what do you think?

ercanucan:
@tkrop Thank you very much for your response! As you suggested, I created a new file `problem-1.0.2.yaml`.

- We actually use these guidelines in some of our projects at Siemens and we actually do enforce these using `spectral` as we believe it brings value to keep API specs consistent across various projects.
As currently the examples within `problem-1.0.1.yaml` do not fit the specified format (`uri`), we have seen that we could actually make the fix source and contribute it upstream 🙂 

An alternative I was considering could also be to adapt the format to `uri-reference`, which can be a `uri` or a `relative-reference` if we want to keep compatibility for both options as described in RFC3986:
https://tools.ietf.org/html/rfc3986#section-4.1
What would you and the rest of maintainers think about it?

tkrop:
@ercanucan Thanks again for the PR and the hint. Yes, **`uri-reference`** would actually be the correct and expected format. Since this is what we intended to express by `uri` in our initial PR, we would change **`problem-1.0.1yaml`**, since we think that this intention was broadly anticipated.

It would be great, if you can finalize your contribution in this sense. Many thanks

Tronje

ercanucan:
@tkrop Thank you very much for your response! Adapted the PR as you suggested. Please let me know if you'd like further adaptions. Thanks again!

tfrauenstein:
👍

tkrop:
:+1:

tkrop:
@ercanucan Thank you very much for your contribution.

ercanucan:
@tkrop @tfrauenstein Thank you very much for the collaboration.

Issue #607, opened 2020-08-25
Zally issue creation sometimes omits commits


ePaul:
We observed that some of the merges in this repository don't create a Zally issue, when they should have done so.

Examples:
* #579 ([build](https://travis-ci.org/github/zalando/restful-api-guidelines/builds/680531583)), #598 ([build](https://travis-ci.org/github/zalando/restful-api-guidelines/builds/719586629))

*Analysis*:

It looks like the result of https://api.github.com/repos/zalando/restful-api-guidelines/pulls/598/files (and similar) didn't return useful results (possibly an error?), which resulted in a error of jq, and then the rest of the script thought that nothing changed, and decided not to create a Zally issue.

GitHub's commit information API lags a bit behind the actual commits. So when it doesn't return anything, we might need to wait a bit and retry, when this happened.

ePaul:
Other missed (and relevant) PRs:
* #564 (compound key identifiers),
  → manually created as https://github.com/zalando/zally/issues/1182
* #560 (naming conventions for enums) 
  → manually included in https://github.com/zalando/zally/issues/1183

ePaul:
Potential (partial) solution:

* wait a moment before fetching the files for a merged PR.
* (if feasible) in the error case, wait a bit and try again
* in case of doubt, create an issue

ePaul:
We invested a bit more time to figure out what happened.

* The travis-deploy.sh script is trying to figure out which PR belongs to the merge commit. It is using jq's `capture("#(?<n>[0-9]+) ")` to find the PR number from the merge commit. This finds `Merge pull request #591 from zalando/gh-589-fix-link`, but not `Added naming convention for enums (#560)`. If it fails, no issue is created.
* After that, the script is trying to look at the files for the PR, to check whether a file in `chapters` changed.
   This sometimes fails, because the Github API doesn't return any data yet – in this case, no issue is created.

To make it more robust, we would need to:

* fix the regexp
* add a loop for the second request too
* as a fallback, if we didn't get anything, still create an issue.

tkrop:
Lets assume that #628 was finally fixing the issue.

PR #623, opened 2021-01-12
feat: improve zally issue creation (#607)

tkrop:
fixes. #607.

The new version of the script is intended to be much more robust. It will only repeat the initial request until the commit message is available and than use the result to evaluate the situation and create an issue. If the script decides to create an issue, it checks whether the commit has attached pull requests using a new experimental github api command and uses this as context information instead of the commit information. Anyhow, both ways should make it easy to review the changes.


tfrauenstein:
👍

tkrop:
:+1:

PR #625, opened 2021-01-15
Guidance for GET-with-body documentation

tfrauenstein:
Standardise how GET-with-body hint should be provided as part of the POST endpoint description.
Analog for DELETE-with-body.

### DIFF
--- a/chapters/http-requests.adoc
+++ b/chapters/http-requests.adoc
@@ -24,12 +24,12 @@ collection is empty) or {404} (if the collection is missing)
 
 
 [[get-with-body]]
-=== GET with body
+=== GET with body payload
 
 APIs sometimes face the problem, that they have to provide extensive structured
 request information with {GET}, that may conflict with the size limits of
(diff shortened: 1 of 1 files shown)


otaciliolacerda:
👍

tfrauenstein:
👍

otaciliolacerda:
👍

tkrop:
@tfrauenstein I'm wondering whether `request body payload` isn't a bit over-specified and hard to read.

tfrauenstein:
@tkrop I think it is ok -- though it is a bit over-specified. Reason: we mostly used `body` so far (and sometimes `payload`) and I think `payload` is the better term -- but I did not deleted `body` due to backward compatibility with `GET-with-Body` references.  Do you know whether acsiidoc supports multiple reference labels? -- than I can do the `body`clean-up:
```
[[get-with-body]]
[[get-with-payload]]
=== GET with payload
```

tkrop:
:+1:

tfrauenstein:
👍

PR #628, opened 2021-01-19
fix: and improve zally notification script (#607)

tkrop:
fixes #623.

It looks as if the new GitHub API does not always returning the `pull-requests`. To compensate, I added a loop and the old mechanism as fallback. In addition, I fixed also some issues with the comparison in case of errors.


tfrauenstein:
👍

ePaul:
:+1:

Issue #631, opened 2021-02-05
Remote reference rule is outdated


tkrop:
Our current rule about remote references ([#234](https://opensource.zalando.com/restful-api-guidelines/#234)) is outdated. We have created new models to reference to in https://opensource.zalando.com/restful-api-guidelines/, but have not documented this yet.

Not sure, but this may already effect Zally by making showing the message `MAY – Specify Success and Error Responses` when using the latest problem location https://opensource.zalando.com/restful-api-guidelines/problem-1.0.1.yaml#/Problem.

PR #636, opened 2021-02-15
feat: update remote reference guideline (#631)

tkrop:
fixes: #631.

### DIFF
--- a/chapters/general-guidelines.adoc
+++ b/chapters/general-guidelines.adoc
@@ -67,11 +67,17 @@ The reason is, that the content referred to is _in general_ *not durable* and
 unexpected ways.
 
 However, you may use remote references to resources accessible by the following
-service URLs.
+service URLs:
 
-* `https://infrastructure-api-repository.zalandoapis.com/` (internal repository of APIs)
-* `https://opensource.zalando.com/problem/` (see <<176>>)
-* `https://zalando.github.io/problem/` (deprecated alias for <<176>>)
+* `https://infrastructure-api-repository.zalandoapis.com/` (internal repository
+  of user defined APIs)
+* `https://opensource.zalando.com/restful-api-guidelines/` (default guideline
+  defined API fragments)
+
+For compatibility we support the following deprecated remote references:
+
+* `https://opensource.zalando.com/problem/` (alias for `Problem`, see <<176>>)
+* `https://zalando.github.io/problem/` (another alias for `Problem`, see <<176>>)
 
 As we control these URLs, we ensure that their content is *durable* and
 *immutable*. This allows to define API specifications by using fragments
--- a/chapters/general-guidelines.adoc
+++ b/chapters/general-guidelines.adoc
@@ -69,10 +69,13 @@ unexpected ways.
 However, you may use remote references to resources accessible by the following
 service URLs:
 
-* `https://infrastructure-api-repository.zalandoapis.com/` (internal repository
-  of user defined APIs)
-* `https://opensource.zalando.com/restful-api-guidelines/` (default guideline
-  defined API fragments)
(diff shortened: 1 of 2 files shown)


chapters/general-guidelines.adoc
```diff
@@ -67,11 +67,17 @@ The reason is, that the content referred to is _in general_ *not durable* and
...
 However, you may use remote references to resources accessible by the following
-service URLs.
+service URLs:
 
-* `https://infrastructure-api-repository.zalandoapis.com/` (internal repository of APIs)
-* `https://opensource.zalando.com/problem/` (see <<176>>)
-* `https://zalando.github.io/problem/` (deprecated alias for <<176>>)
+* `https://infrastructure-api-repository.zalandoapis.com/` (internal repository
+  of user defined APIs)
+* `https://opensource.zalando.com/restful-api-guidelines/` (default guideline
```
tfrauenstein, line 74: This is the link to the guidelines, not to the 'API fragments'.
Should we instead use https://github.com/zalando/restful-api-guidelines/tree/master/models

BTW: In [Rule 183](https://opensource.zalando.com/restful-api-guidelines/#183) we use 
https://opensource.zalando.com/restful-api-guidelines/models/request-headers-1.0.0.yaml#/X-Flow-ID
which cannot be resolved, can it?

chapters/general-guidelines.adoc
```diff
@@ -67,11 +67,17 @@ The reason is, that the content referred to is _in general_ *not durable* and
...
+  defined API fragments)
+
+For compatibility we support the following deprecated remote references:
+
+* `https://opensource.zalando.com/problem/` (alias for `Problem`, see <<176>>)
+* `https://zalando.github.io/problem/` (another alias for `Problem`, see <<176>>)
```
tfrauenstein, line 80: editorial --> *Hint:* The formerly used remote references to `problem` API definition fragment (aliases `https://opensource.zalando.com/problem/` and `https://zalando.github.io/problem/`) are deprecated, but still supported for compatibility.

chapters/general-guidelines.adoc
tkrop: I added the link to the published model location - which is not the location in `github.com`. This is also why `https://opensource.zalando.com/restful-api-guidelines/models/request-headers-1.0.0.yaml#/X-Flow-ID` does not work, because models are not published to `/models` but the `/`.

The latter may have been an error, but I'm not sure that we can fix this now, since the resources are already widely used.

tfrauenstein: Shouldn't it be defined like follows:
* `https://opensource.zalando.com/restful-api-guidelines/<*model*>` -- used to refer to `<*model*>` API fragment definitions provided in `https://github.com/zalando/restful-api-guidelines/tree/master/models`

chapters/general-guidelines.adoc
```diff
@@ -67,11 +67,17 @@ The reason is, that the content referred to is _in general_ *not durable* and
...
 
 However, you may use remote references to resources accessible by the following
-service URLs.
+service URLs:
 
-* `https://infrastructure-api-repository.zalandoapis.com/` (internal repository of APIs)
-* `https://opensource.zalando.com/problem/` (see <<176>>)
-* `https://zalando.github.io/problem/` (deprecated alias for <<176>>)
+* `https://infrastructure-api-repository.zalandoapis.com/` (internal repository
+  of user defined APIs)
```
tfrauenstein, line 73: * `https://infrastructure-api-repository.zalandoapis.com/` -- internal repository
  of user defined and deployed (immutable) API definition revisions

tfrauenstein:
👍

tkrop:
:+1:

ePaul:
:+1:

PR #637, opened 2021-02-15
feat: add client note and improve json problem guideline (#624)

### LINKED ISSUE #624: Robustness: clients should not break/crash when the server returns a non problem/json answer
Rule [#176](https://opensource.zalando.com/restful-api-guidelines/#176) directs servers to always return a problem/json response if there was an error.

Unfortunately, some clients are building on this, and just completely fail when they receive some error response with a different content (e.g. text/html), when they would log problem/json content (possibly when parsing the problem object from the response, or when trying to access parts of the parsed result for a log message).

Non-problem/json responses can always occur (e.g. because of some infrastructure components like the AWS load balancer returning a 502 error when the service is not reachable, and those infrastructure components often know nothing about the API guidelines).

Clients should be prepared to handle errors without a problem/json response body gracefully, relying just on the HTTP status code for error handling. (If it's just about logging, either omit this part of the log, or log e.g. the first line of the response (if it's a text/* one).)

We could either add a sentence/paragraph to Rule 176, or add a new rule for the expected client behavior.

(Thanks to @hjacobs for the suggestion.)

tkrop:
fixes #624.

Improve guideline to use Problem JSON:

* Add better guidance for consumers to setup `Accept: application/problem+json` but be prepared to receive nothing at all.
* Add better motivation why we prefer extended failure information via Problem JSON.

### DIFF
--- a/chapters/http-status-codes-and-errors.adoc
+++ b/chapters/http-status-codes-and-errors.adoc
@@ -339,17 +339,25 @@ number of requests made within a given window for each named entity.
 
 
 [#176]
-== {MUST} use problem JSON
+== {MUST} support problem JSON
 
-{RFC-7807}[RFC 7807] defines a Problem JSON object and  the media type
-`application/problem+json`. Operations should return it (together with a
-suitable status code) when any problem occurred during processing and you can
-give more details than the status code itself can supply, whether it be caused
-by the client or the server (i.e. both for {4xx} or {5xx} error codes).
+{RFC-7807}[RFC 7807] defines a Problem JSON object using the media type
+`application/problem+json` to provide an extensible human and machine readable
+failure information beyond the `status` to transports the failure kind (`type`
+/ `title`) and the failure cause and location (`instant` / `detail`). To make
+best use of this additional failure information, we expect that every endpoint
+is capable of returning a Problem JSON together with the status code, when a
+problem occurs during processing.
+
+**Note:** Clients **must not** rely on receiving a Problem JSON, since failure
+responses may be created by infrastructure components unaware of the API
+guidelines. In addition, clients usually need to setup the {Accept}-Header
+with `application/problem+json` to trigger delivery of the extended failure
+information.
 
 The Open API schema definition of the Problem JSON object can be found
-https://zalando.github.io/problem/schema.yaml[on github]. You can
-reference it by using:
+https://opensource.zalando.com/restful-api-guidelines/problem-1.0.1.yaml[on
+GitHub]. You can reference it by using:
 
 [source,yaml]
 ----
@@ -363,10 +371,11 @@ responses:
 ----
 
 You may define custom problem types as extensions of the Problem JSON object
-if your API needs to return specific, additional and detailed error information.
+if your API needs to return specific, additional, and more detailed error
+information.
 
-Problem `type` and `instance` identifiers in our APIs are not meant to be
-resolved. {RFC-7807}[RFC 7807] encourages that custom problem types are URI
+**Note:** Problem `type` and `instance` identifiers in our APIs are not meant
+to be resolved. {RFC-7807}[RFC 7807] encourages that problem types are URI
 references that point to human-readable documentation, **but** we deliberately
 decided against that, as all important parts of the API must be documented
 using <<101, OpenAPI>> anyway. In addition, URLs tend to be fragile and not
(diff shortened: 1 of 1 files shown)


chapters/http-status-codes-and-errors.adoc
```diff
@@ -339,17 +339,25 @@ number of requests made within a given window for each named entity.
...
+== {MUST} support problem JSON
+
+{RFC-7807}[RFC 7807] defines a Problem JSON object using the media type
+`application/problem+json` to provide an extensible human and machine readable
+failure information beyond the `status` to transports the failure kind (`type`
+/ `title`) and the failure cause and location (`instant` / `detail`). To make
+best use of this additional failure information, we expect that every endpoint
+is capable of returning a Problem JSON together with the status code, when a
+problem occurs during processing.
+
```
tfrauenstein, line 351: To make best use ... processing. 
-->
Endpoint responses must provide additional failure information via the 
Problem JSON object when an client usage error ( {4xx} HTTP status codes) 
or server side processing error ( ({5xx} HTTP status codes) is detected.

chapters/http-status-codes-and-errors.adoc
```diff
@@ -339,17 +339,25 @@ number of requests made within a given window for each named entity.
...
+**Note:** Clients **must not** rely on receiving a Problem JSON, since failure
+responses may be created by infrastructure components unaware of the API
+guidelines. In addition, clients usually need to setup the {Accept}-Header
+with `application/problem+json` to trigger delivery of the extended failure
+information.
```
tfrauenstein, line 356: Do we need this hint? I think the extended failure information should be always returned w/o need for client request acceptance hints.

chapters/http-status-codes-and-errors.adoc
```diff
@@ -339,17 +339,25 @@ number of requests made within a given window for each named entity.
...
+`application/problem+json` to provide an extensible human and machine readable
+failure information beyond the `status` to transports the failure kind (`type`
+/ `title`) and the failure cause and location (`instant` / `detail`). To make
+best use of this additional failure information, we expect that every endpoint
+is capable of returning a Problem JSON together with the status code, when a
+problem occurs during processing.
+
+**Note:** Clients **must not** rely on receiving a Problem JSON, since failure
+responses may be created by infrastructure components unaware of the API
+guidelines. In addition, clients usually need to setup the {Accept}-Header
```
tfrauenstein, line 354: Clients ** must not** .... guidelines.
-->
Clients must be robust and **must not rely** on Problem JSON object being returned in client or server error situations because a) failure responses may be created by infrastructure components unaware of the API
guidelines, and b) services might not always follow this guideline and not return the problem object in all HTTP error status situations.

chapters/http-status-codes-and-errors.adoc
```diff
@@ -339,17 +339,25 @@ number of requests made within a given window for each named entity.
...
-{RFC-7807}[RFC 7807] defines a Problem JSON object and  the media type
-`application/problem+json`. Operations should return it (together with a
-suitable status code) when any problem occurred during processing and you can
-give more details than the status code itself can supply, whether it be caused
-by the client or the server (i.e. both for {4xx} or {5xx} error codes).
+== {MUST} support problem JSON
+
+{RFC-7807}[RFC 7807] defines a Problem JSON object using the media type
+`application/problem+json` to provide an extensible human and machine readable
+failure information beyond the `status` to transports the failure kind (`type`
```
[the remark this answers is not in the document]
tkrop: Tried to by to smart referring to the Problem JSON `status` field, but this is definitely wrong here. I better the formal name `HTTP response status code`.

chapters/http-status-codes-and-errors.adoc
tkrop: I think, I got your point.

chapters/http-status-codes-and-errors.adoc
tkrop: Lets formulate (b) in a way that does not sound as if we approve this. I did this for (a) similar.

tfrauenstein: hmm -- errors will happen, and robustness against typical errors is a good idea. What about...
b) services might not always return the problem object with 100% guideline implementation in all failure situations

chapters/http-status-codes-and-errors.adoc
tkrop: I believe *yes*. It is often not obvious for clients, that they need to add `application/problem+json` to the `Accept`-Header, and also for service creators that their Problem is not delivered. I have improved the hint to better express this.

tkrop:
@tfrauenstein thanks for the review. I have updated based on your proposals.

chapters/http-status-codes-and-errors.adoc
tkrop: Have a look at the update I pushed.

vadeg:
👍

tkrop:
:+1:

ePaul:
:+1:

Issue #629, opened 2021-01-26
Rule 203 (provide means for event ordering) is a must for data change events


tfrauenstein:
We see event provider teams not consequently following [Rule 203](https://opensource.zalando.com/restful-api-guidelines/#203) (partially) because they are not aware that it is a MUST rule for data change events and a SHOULD rule for business events.  Only very few event types have provided this information, which makes the job hard for change data capture when creating analytical source datasets. We also have seen incidents (e.g. see [Post Mortem [internal link]](https://docs.google.com/document/d/1642eVGbq-FHLl9gBuCtQZId6N4urv7pEZCpVcPoXa50/edit#)) with the root cause of analytic consumer teams, faced with the absence of this information, made false assumptions on the (business) ordering. As  incident mitigation action we agreed to update the rule for more clarity about the must, why, and how and explicitly list the exceptions.

tfrauenstein:
done with https://github.com/zalando/restful-api-guidelines/pull/644

PR #644, opened 2021-03-09
Event ordering information is a must for data change events

tfrauenstein:
https://github.com/zalando/restful-api-guidelines/issues/629

### DIFF
--- a/chapters/events.adoc
+++ b/chapters/events.adoc
@@ -583,7 +583,7 @@ multiple event types to represent each step. For now we suggest
 assessing each option and sticking to one for a given business process.
 
 
-[#242]
+[#203]
 == {SHOULD} provide explicit event ordering for general events
 
 Event processing consumer applications need the order information to 
@@ -616,9 +616,9 @@ jump forward and backward to compensate drifts or leap-seconds. If you use anywa
 timestamps to indicate event ordering, you _must_ carefully ensure that the designated 
 event order is not messed up by these effects and use UTC time zone format.
 
-*Note:* The `occured_at` and `partition_offset` metadata of Nakadi typically is 
-different from the business event ordering, since (i) Nakadi is a distributed concurrent 
-system without atomic, ordered event creation and (ii) the application's implementation
+*Note:* The `occurred_at` and `partition_offset` metadata set by Nakadi typically is 
+different from the business event ordering, since (1) Nakadi is a distributed concurrent 
+system without atomic, ordered event creation and (2) the application's implementation
 of event publishing may not exactly reflect the business order. The business ordering 
 information is application knowledge, and implemented in the scope of event 
 partitions or specific entities, but may also comprise all events, if scaling 
@@ -631,28 +631,28 @@ requirements are low.
 You *must* use data change events to signal changes of stored entity instances
 and facilitate e.g. change data capture (CDC). Event sourced change data capture 
 is crucial for our data integration architecture as it supports the logical 
-replication (and reconstruction) of the application datastores into 
-transactional datasets used as input for data analytics and AI. 
+replication (and reconstruction) of the application datastores 
+to the data analytics and AI platform as transactional source datasets. 
 
 * Change events must be provided when publishing events that represent created, 
   updated, or deleted data.
 * Change events must provide the complete entity data including the identifier of 
   the changed instance to allow aggregation of all related events for the entity.
-* Change events <<203>>.
+* Change events <<242>>.
 * Change events must be published reliably by the service.
 
 
-[#203]
+[#242]
 == {MUST} provide explicit event ordering for data change events
 
 While the order information is recommended for business events, it *must* be 
 provided for data change events. The ordering information defines the (create, 
 update, delete) change order of the data entity instances managed via the 
 application's transactional datastore. It is needed for change data capture 
-to keep transactional data replicas in sync as source for data analytics. 
+to keep transactional dataset replicas in sync as source for data analytics. 
 
 For details about how to provide the data change ordering information, 
-please check <<242>>.
+please check <<203>>.
 
 *Exception*: In situations where the transactional data is 'append only', 
 i.e. entity instances are only created, but never updated or deleted, the 
--- a/chapters/events.adoc
+++ b/chapters/events.adoc
@@ -329,14 +329,13 @@ An _event category_ describes a generic class of event types. The
 guidelines define two such categories:
 
 * General Event: a general purpose category.
-* Data Change Event: a category used for describing changes to data
-entities used for data replication based data integration.
+* Data Change Event: a category used to inform about changes to data
+entities, and used for data replication based data integration.
 
-The set of categories is expected to evolve in the future.
-
-A category describes a predefined structure that event publishers must
-conform to along with standard information about that kind of event
-(such as the operation for a data change event).
+A category describes a predefined structure (e.g. consisting of event 
+metadata as part of the event payload) that event publishers must
+conform to along with standard information specific for the event category
+(e.g. the operation for data change events).
 
 
 === The general event category
@@ -352,8 +351,8 @@ GeneralEvent:
     custom schema payload as the top level of the document, with the
     "metadata" field being required and reserved for standard metadata. An
     instance of an event based on the event type thus conforms to both the
-    EventMetadata definition and the custom schema definition. Previously
-    this category was called the Business Category.
+    EventMetadata definition and the custom schema definition. 
+    Hint: In earlier versions this category was called the Business Category.
   required:
     - metadata
   properties:
@@ -373,14 +372,14 @@ schema:
 
 Note:
 
-* The General Event in a previous version of the guidelines was called a
-_Business Event_. Implementation experience has shown that the
+* The General Event was called a _Business Event_ in earlier versions of 
+the guidelines. Implementation experience has shown that the
 category's structure gets used for other kinds of events, hence the name
 has been generalized to reflect how teams are using it.
 * The General Event is still useful and recommended for the purpose of
 defining events that drive a business process.
 * The Nakadi broker still refers to the General Category as the Business
-Category and uses the keyword "business" for event type registration.
+Category and uses the keyword `business` for event type registration.
 Other than that, the JSON structures are identical.
 
 See <<201>> for more guidance on how to use the category.
@@ -432,17 +431,17 @@ DataChangeEvent:
 ----
 
 The Data Change Event Category is structurally different to the General
-Event Category. It defines a field called `data` for placing the custom
-payload information, as well as specific information related to data
+Event Category defining a field called `data` as container for the custom
+payload, as well as specific information related to data
 changes in the `data_type`. In the example fragment below, the fields
 `a` and `b` are part of the custom payload housed inside the `data`
 field:
 
-See the following guidelines for more guidance on how to use the Data
-Change Event Category:
+The following guidelines specifically apply to Data Change Events:
 
-* <<205>>
 * <<202>>
+* <<203>>
+* <<205>>
 * <<204>>
 
 
@@ -515,7 +514,7 @@ EventMetadata:
       example: '0'
 ----
 
-Please note than intermediaries acting between the producer of an event
+Please note that intermediaries acting between the producer of an event
 and its ultimate consumers, may perform operations like validation of
 events and enrichment of an event's <<event-metadata,metadata>>. For
 example brokers such as Nakadi, can validate and enrich events with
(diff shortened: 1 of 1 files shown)


chapters/events.adoc
```diff
@@ -588,52 +587,66 @@ assessing each option and sticking to one for a given business process.
...
-   events for an entity, or
- * a strictly monotonically increasing message counter.
-
-In the event type definition, the `ordering_key_fields` property
-should be used to indicate which field(s) contains the ordering
-key, if any.
-
-System timestamps are not necessarily a good choice, since exact
-synchronization of clocks in distributed systems is difficult, two
+== {MUST} provide explicit event ordering for data change events
```
ePaul, line 606: From the discussion in the meeting, this might be a better way to structure it into two rules, to capture SHOULD + MUST:

* {SHOULD} provide explicit event ordering (for general events)
** general reasoning why it's needed
* {MUST} provide explicit event ordering for data change events
** reasoning why it's even more needed for data change events

tkrop: For my opinion, this all is getting way to long and complicated for users to follow and understand. The section is mixing up the fundamental discussions about event ordering and the practical approach how to implement this in Zalando with very specific details of data analytics and their motivation in Zalando. More over it is turning this logical order around making it in my opinion hard to follow.

As mentioned in the meeting, I would stick to the before order of present the general motivation and discussion about event ordering in the general *should* rule. This than should be followed by the spacial data change event motivation and discussion of the impact for data analytics in the *must* rule to keep both parts understandable. Here is my suggestion using your work how this may be formulated:

```
[203]
== {SHOULD} provide explicit event ordering (for general events)

Some uses case, e.g. processing errors, require event consumers to reconstruct
an event streams or replay events outside the context of the original event
stream. To support this, events _should_ contain a way to restore the partial
order of related events.

This can be done – among other ways – by adding

 * a strictly monotonically increasing version (e.g. as created by a database)
   to an entity, or
 * a strictly monotonically increasing event counter (per partition or entity).

*Hint:* timestamps are often a bad choice, since synchronization of clocks
in distributed systems is difficult and system clocks may jump forward and
backward to compensate drifts or leap-seconds. When despite using timestamps to
indicate event ordering, you *must* carefully ensure that the designated event
order is not messed up by these effects and use UTC time zone format.

To accomplish this explicit event ordering the event type definition

* *must* specify a the `ordering_key_fields` property to indicate which
  field(s) contain the ordering key, and
* *should* specify the `ordering_instance_ids` property to define which
  field(s) represents the data instance identifier.

*Note:* The `ordering_instance_ids` can be used to restrict the scope in which
the `ordering_key_fields` provide the strict order. If undefined, the ordering
is assumed to be provided in scope of all events.

*Remark:* event odering based on data structures that can be converged upon
in a distributed system, e.g. https://en.wikipedia.org/wiki/Logical_clock[logical
clocks] or https://en.wikipedia.org/wiki/Vector_clock[vector clocks]), are
outside the scope of this guidance.

[242]
== {MUST} provide explicit event ordering for data change events

All data change events *must* provided explicit event ordering by defining the
`ordering_key_fields` and `ordering_instance_ids` as proposed in <<203>>. This
is required to enable the change data capturing of our analytics platform to
reconstruct the event order after failures, and keep the replicated data store
in sync with the source.

*Note:* The `occured_at` and `partition_offset` metadata of Nakadi provide no
event ordering, since Nakadi is a distributed concurrent system without atomic,
ordered event publication. The ordering information is business knowledge
provided by the event publisher. It is typically implemented in scope of event
partitions or specific entities, but may also comply all events, if scaling
requirements are low.
```

I must admit, the last paragraph could as well be located in the general rule.

tkrop (changes requested):
Sorry, that the review took so long. Please have a look at my comments.

chapters/events.adoc
```diff
@@ -329,14 +329,13 @@ An _event category_ describes a generic class of event types. The
 guidelines define two such categories:
 
 * General Event: a general purpose category.
-* Data Change Event: a category used for describing changes to data
-entities used for data replication based data integration.
+* Data Change Event: a category used to inform about changes to data
+entities, and used for data replication based data integration.
```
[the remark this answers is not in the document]
tfrauenstein: 👍

chapters/events.adoc
```diff
@@ -588,52 +587,72 @@ assessing each option and sticking to one for a given business process.
...
-category.
-
-* Change events must identify the changed entity to allow aggregation of
-all related events for the entity.
+You _must_ use data change events to signal changes of stored entity instances 
+and facilitate e.g. change data capture (CDC), i.e. keeping datastore replicas 
+in sync with minimal delays sourced from events.  Data change events are 
+crucial e.g. for our data analytics integration architecture: They facilitate 
+the logical replication of the microservice application datastores into 
+transactional datasets as input for data analytics and AI.
```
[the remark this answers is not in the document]
tfrauenstein: 👍

chapters/events.adoc
```diff
@@ -588,52 +587,72 @@ assessing each option and sticking to one for a given business process.
...
+The ordering information defines the transactional data change order managed 
+by the application, and is used e.g. for analytics change data capture, i.e. 
+keeping transactional data in sync as source for analytics. The ordering 
+information is business knowledge located in the application's publisher and 
+typically is implemented in scope of event partitions or event entity instances 
+(but may also be implemented in scope of the event type for all events, if 
+scaling requirements are low). The order information differs from Nakadi 
+provided `occured_at` or `partition_offset` metadata, since application's event 
+creation may not reflect the transactional order and Nakadi is a distributed 
+concurrent system without ordered atomic event publication.
```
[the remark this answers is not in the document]
tfrauenstein: 'Transactional data' is a common term used in data analytics, to differentiate from 'analytics data'. 
I'm not talking here about 'transactions' or 'atomic' operations. 
However, I will change the wording to void misunderstandings:
a) The ordering information defines the change order of the transactional data managed
by the application, and ...
b) since application's event creation may not reflect the order of the transactional data changes and Nakadi...

chapters/events.adoc
tfrauenstein: I'm not convinced, but it also makes sense -- will do it.

chapters/events.adoc
```diff
@@ -585,55 +583,80 @@ multiple event types to represent each step. For now we suggest
...
+*Hint:* timestamps are often a bad choice, since in distributed systems events 
+may occur at the same time, or clocks are not exactly synchronized, or 
+jump forward and backward to compensate drifts or leap-seconds. If you use anyway
+timestamps to indicate event ordering, you _must_ carefully ensure that the designated 
+event order is not messed up by these effects and use UTC time zone format.
+
+*Note:* The `occured_at` and `partition_offset` metadata of Nakadi typically is 
+different from the business event ordering, since (i) Nakadi is a distributed concurrent 
+system without atomic, ordered event creation and (ii) the application's implementation
+of event publishing may not exactly reflect the business order. The business ordering
```
tkrop, line 622: 1. I would like to prevent `(i)` and `(ii)` here for style reasons and use (1), (2).
2. `occured_at` -> `occurred_at` (my fault).
3. (ii) has no conjunction to `occurred_at` and `partition_offset` because these are set on the Nakadi service side.

Besides, I'm not sure whether (ii) is of much help for the reader: Programmers will argue that they can and will take care of this (using the database transactions handling). The danger here is more, that they are often not aware when their Nakadi libraries on low concurrency levels not ensuring the business order for the sake of speed, but this is just based on guessing, since we can't know the details.

chapters/events.adoc
```diff
@@ -585,55 +583,80 @@ multiple event types to represent each step. For now we suggest
...
 == {MUST} use data change events to signal mutations
 
-When publishing events that represents created, updated, or deleted
-data, change event types must be based on the Data Change Event
-category.
+You *must* use data change events to signal changes of stored entity instances
+and facilitate e.g. change data capture (CDC). Event sourced change data capture 
+is crucial for our data integration architecture as it supports the logical 
+replication (and reconstruction) of the application datastores into 
+transactional datasets used as input for data analytics and AI.
```
tkrop, line 635: Please circumvent the use of `transactional dataset` here, since (1) the type of data is of no relevance for this rule, and (2) term `transactional` can be easily interpreted out of context here. While `transactional datasets` in data mining refer to the *business transactions* captured by a data set, the connotation in data change events and event ordering can be to easily misunderstood as *database transactions*.

chapters/events.adoc
```diff
@@ -585,55 +583,80 @@ multiple event types to represent each step. For now we suggest
...
 
 
+[#242]
+== {SHOULD} provide explicit event ordering for general events
+
+Event processing consumer applications need the order information to 
+reconstruct the business event stream, for instance, in order to replay 
+events in error situations, or to execute analytical use cases 
+outside the context of the original event stream consumption. 
+All general events (fka business events) *should* be provided with the explicit
```
tkrop, line 593: Why do you want to remove the brackets around `for general events`?

I think this restricts the rule more than necessary, since it suggests that none-general events are not included, which should not be our intention. This is especially supported by defining general events as equivalent to business events `(fka business events)`. In this context here I would keep `(for general events)` in the title and stick to `All events *should* ...` to make it as general as possible.

chapters/events.adoc
```diff
@@ -585,55 +583,80 @@ multiple event types to represent each step. For now we suggest
 assessing each option and sticking to one for a given business process.
 
 
+[#242]
+== {SHOULD} provide explicit event ordering for general events
```
tkrop, line 587: Please correct rule numbering and rule order as follows:

```
[#202]
== {MUST} use data change events to signal mutations
...

[#203]
== {SHOULD} provide explicit event ordering (for general events)
...

[#242]
== {MUST} provide explicit event ordering for data change events
...
```

tkrop:
@tfrauenstein I think we are getting to a very good compromise.

chapters/events.adoc
tfrauenstein: 1+2) +1
3) I think the conjunction is given -- the statement is that it does not reflect the business order (not that it is set by Nakadi)
4) understood, but IMO the hint is helpful -- the Nakadi concurrency problem is not the only problem to solve to provide the business order.

chapters/events.adoc
tfrauenstein: In data analytics, 'transactional dataset' we use the term intensively  for 'datasets sourced from OLTP productive environment'  -- we should explicitly refer to this term. However, changed the wording a bit.

chapters/events.adoc
tfrauenstein: 'All events *should*...' is not correct, since it is a must for data change events.
We have two event categories: general vs. data change -- and a want to avoid misunderstanding wrt 'should' vs. 'must'.

chapters/events.adoc
tfrauenstein: +1

chapters/events.adoc
```diff
@@ -373,14 +372,14 @@ schema:
 
 Note:
 
-* The General Event in a previous version of the guidelines was called a
-_Business Event_. Implementation experience has shown that the
+* The General Event was called a _Business Event_ in earlier versions of 
+the guidelines. Implementation experience has shown that the
```
tkrop, line 376: Just saw this scrolling. Can we write `*Note:* ...` here for consistency?

Besides, I'm wondering about the sentence above with `:` on the end of the sentence. Either `:` is wrong, or the whole sentence is out of context - may be it can be removed at all. Have a look at the example, it contains no `a` or `b` fields.

chapters/events.adoc
```diff
@@ -432,17 +431,17 @@ DataChangeEvent:
 ----
 
 The Data Change Event Category is structurally different to the General
-Event Category. It defines a field called `data` for placing the custom
-payload information, as well as specific information related to data
+Event Category defining a field called `data` as container for the custom
+payload, as well as specific information related to data
 changes in the `data_type`. In the example fragment below, the fields
 `a` and `b` are part of the custom payload housed inside the `data`
 field:
```
tkrop, line 438: Same as in the before comment. The example is likely the one now above, but there is no `a` and `b` fields at all, so may be we can remove it.

chapters/events.adoc
tkrop: Yes, but making a sentence stronger for a special case in a second sentence is usually no perception problem, while excluding some we want to include may be. But I let this go.

chapters/events.adoc
```diff
@@ -585,55 +583,80 @@ multiple event types to represent each step. For now we suggest
 assessing each option and sticking to one for a given business process.
 
 
+[#203]
+== {SHOULD} provide explicit event ordering for general events
```
tkrop, line 587: I would keep rules in order #202, #203, #242, since #203 and #242 have a stronger on each other building context of ordering than #202 and #242 with talking about data change events.

tkrop (approved):

tkrop:
:+1:

chapters/events.adoc
tfrauenstein: +1

chapters/events.adoc
tkrop: Yes, it is definitely hard to solve, but you can't prevent engineers from trying this by challenging them. Might have the opposite effect...

chapters/events.adoc
tfrauenstein: +1

chapters/events.adoc
tkrop: My problem is the context of the word not the meaning. I acknowledge that it is known in the data-mining community - however still only 20.000 hits, which is not much. The problem is, that the term is likely to be misunderstood by 80% of the readers.

chapters/events.adoc
tfrauenstein: I prefer grouping as of now: grouping by 'category aspect', instead of by 'ordering aspect'

chapters/events.adoc
tfrauenstein: yes please :-)

tkrop:
:+1:

vadeg:
👍

Issue #651, opened 2021-04-10
Change proprietary headers format in the hint for Rule 183


vadeg:
Proprietary headers in [Rule 183](https://opensource.zalando.com/restful-api-guidelines/#183) doesn't match [Rule 132](https://opensource.zalando.com/restful-api-guidelines/#132).

tfrauenstein:
I think we should Rule 132 to the proprietary header section and include 'X-' prefixing -- Common / standard headers are anyway not in our naming responsibility. What additional concrete mismatches you see?

vadeg:
> Common / standard headers are anyway not in our naming responsibility

Then it could happen that we will see a lot of `location` instead of `Location` and `accept-encoding` instead of `Accept-Encoding` headers in APIs with arguments like "headers are case-insensitive anyways". 

Should we mention in the [Rule 133](https://opensource.zalando.com/restful-api-guidelines/#133) than _you should preserve naming convension for standard headers_?

tfrauenstein:
yes -- see my proposal https://github.com/zalando/restful-api-guidelines/pull/653

tkrop:
Replaced by #660.

PR #653, opened 2021-04-14
Editorial update of header naming convention

tfrauenstein:
https://github.com/zalando/restful-api-guidelines/issues/651

* move header naming convention to Common Headers
* make clearer that first letter upper case writing should be followed (though generic HTTP standard defines headers as case insensitive)

### DIFF
--- a/chapters/common-headers.adoc
+++ b/chapters/common-headers.adoc
@@ -1,9 +1,35 @@
 [[common-headers]]
-= Common headers
+= Standard headers
 
-This section describes a handful of headers, which we found raised the most
+This section describes a handful of standard headers, which we found raising the most
 questions in our daily usage, or which are useful in particular circumstances
 but not widely known.
+
+
+[#133]
+== {MAY} use standardized headers
+
+Use http://en.wikipedia.org/wiki/List_of_HTTP_header_fields[this list] and
+explicitly mention its support in your Open API definition.
+
+
+[#132]
+== {SHOULD} use uppercase separate words with hyphens for HTTP headers
+
+This convention is followed by most standard headers e.g. as defined in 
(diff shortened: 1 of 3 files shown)


chapters/common-headers.adoc
```diff
@@ -1,11 +1,37 @@
 [[common-headers]]
-= Common headers
+= Standard headers
```
tkrop, line 2: Changing a chapter header, I would expect that we also consistently change the file name and the anchor as well as the references. However, we would need to maintain the `[[common-header]]` to prevent external links from breaking.

tfrauenstein: I propose to keep it simple and avoid backward compatibility issue.
(Renamed 'common' to 'standard' is easy to correlate)

tkrop: This will usually will result in stale internal links, when adding new links to the chapter since it violates expectations. My suggestion is surely to keep the old anchor, but also add a new one `[[standard-headers]]`. However, may be we should remove this change at all until we have decided on the new chapter structure overall.

tfrauenstein: ok, added `[[standard-headers]]`. P.S: When doing the restructuring we also may change the file structure.

tkrop: True, that is why I suggested to keep the current heading until we have found out what to do in the future. Can you change back?

tfrauenstein: yes, but we can do this later -- pls. approve

tkrop: Okay, I will do, but if you change the name now the harm of confusion is created.

tkrop:
:+1:

vadeg:
👍

PR #654, opened 2021-04-20
Minor editorial changes

tfrauenstein:
Editorials and merging of enum rules for more clarity.

### DIFF
--- a/chapters/http-requests.adoc
+++ b/chapters/http-requests.adoc
@@ -534,7 +534,7 @@ is authorized to access.
 Implicit filtering could be done on:
 
 * the collection of resources being returned on a {GET} request
-* the fields returned for the resource's detail information
+* the fields returned for the detail information of the resource
 
 In such cases, the fact that implicit filtering is applied must be documented 
 in the API specification's endpoint description. Consider <<227,caching aspects>> 
--- a/chapters/http-requests.adoc
+++ b/chapters/http-requests.adoc
@@ -1,5 +1,5 @@
 [[http-requests]]
-= HTTP requests
+= HTTP requests and responses
 
 
 [#148]
@@ -533,15 +533,12 @@ is authorized to access.
 
 Implicit filtering could be done on:
 
-* the collection of resources being return on a parent {GET} request
-* the fields returned for the resource's detail
+* the collection of resources being returned on a {GET} request
+* the fields returned for the resource's detail information
 
-In such cases, the implicit filtering must be in the API specification (in
-its description).
-
-Consider <<227,caching considerations>> when implicitly filtering.
-
-Example:
+In such cases, the fact that implicit filtering is applied must be documented 
+in the API specification's endpoint description. Consider <<227,caching aspects>> 
+when implicit filtering is provided. Example:
 
 If an employee of the company _Foo_ accesses one of our business-to-business
 service and performs a `{GET} /business-partners`, it must, for legal reasons,
--- a/chapters/json-guidelines.adoc
+++ b/chapters/json-guidelines.adoc
@@ -11,6 +11,13 @@ The first some of the following guidelines are about property names, the later
 ones about values.
 
 
+[#120]
+== {SHOULD} pluralize array names
+
+To indicate they contain multiple values prefer to pluralize array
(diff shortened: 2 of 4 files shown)


vadeg:
👍

chapters/http-requests.adoc
```diff
@@ -533,15 +533,12 @@ is authorized to access.
 
 Implicit filtering could be done on:
 
-* the collection of resources being return on a parent {GET} request
-* the fields returned for the resource's detail
+* the collection of resources being returned on a {GET} request
+* the fields returned for the resource's detail information
```
tkrop, line 537: Not sure, but should this be `resources's details` or `resource details`?

chapters/json-guidelines.adoc
```diff
@@ -11,6 +11,13 @@ The first some of the following guidelines are about property names, the later
 ones about values.
 
 
+[#120]
+== {SHOULD} pluralize array names
+
+To indicate they contain multiple values prefer to pluralize array
+names. This implies that object names should in turn be singular.
```
tkrop, line 18: I know you just moved this, but how about getting the grammar on this straight - and I dislike that part of the sentence is in the title here? 

```
Names of arrays should be pluralized to indicate that they contain multiple values.
This implies in turn that object names should be singular.
```

chapters/json-guidelines.adoc
```diff
@@ -27,16 +34,17 @@ essential to establish a consistent look and feel such that JSON looks
 as if it came from the same hand.
 
 [#125]
-== {SHOULD} represent enumerations as strings
-
-Strings are a reasonable target for values that are by design enumerations.
-
 [#240]
-== {SHOULD} declare enum values using UPPER_SNAKE_CASE format
+== {SHOULD} declare enum values using UPPER_SNAKE_CASE string format
```
tkrop, line 38: How about just `{SHOULD} declare enum values using UPPER_SNAKE_CASE strings`?

chapters/json-guidelines.adoc
```diff
@@ -27,16 +34,17 @@ essential to establish a consistent look and feel such that JSON looks
...
 
-Enum values (using `enum` or {x-extensible-enum}) need to consistently use the upper-snake case format, e.g. `VALUE` or `YET_ANOTHER_VALUE`. This approach allows to clearly distinguish values from properties or other elements.
+Enumerations should be represented as `string` typed OpenAPI component schemas.
```
tkrop, line 40: I stumbled over *OpenAPI component schema* here. I think this is the wrong granularity for an extensible enum declaration. It should be *property* I think.

Beside, please try to stick to a common use of terms - as far as I know, we use *Open API* instead of *OpenAPI*.

chapters/json-guidelines.adoc
tfrauenstein: +1

chapters/http-requests.adoc
tfrauenstein: changed to `detail information of the resource`

chapters/json-guidelines.adoc
tfrauenstein: 👍

chapters/json-guidelines.adoc
tfrauenstein: a) changed to `string typed OpenAPI definitions of request parameters or model properties` -- see https://swagger.io/docs/specification/data-models/enums/

b) `OpenAPI` is the 'official' name/notation and we should consistently use it. However, I admit, we use both notations and created an issue to get rid of it.

tkrop: The usage also on their Web-Site was in the past very inconsistent, so we tried to used consistently *Open API*. If there is now an official consistent usage proposal I suggest to change the whole guideline consistently in a separate pull request but here stick to our current *standard*.

tkrop:
:+1:

chapters/json-guidelines.adoc
tfrauenstein: issue created: https://github.com/zalando/restful-api-guidelines/issues/656
pls. approve

vadeg:
👍

PR #665, opened 2021-05-30
Add Github Actions workflow to build and publish

### LINKED ISSUE #568: Change build process to not need a personal access token
Currently, the build process to create the web side from the repository is [running on Travis CI](https://travis-ci.org/zalando/restful-api-guidelines), and publishes the result by pushing to the gh_pages branch.

For that, it's using a `GH_TOKEN` variable configured in the [project settings](https://travis-ci.org/zalando/restful-api-guidelines/settings), which is a [personal access token](https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line).
This token was previously created by @maxim-tschumak, and after he left the organization, and thus lost the write rights, a new one was created today by me (@ePaul).
This token gives read and write rights to all repositories I have access to, in my name. Unfortunately there seems to be no way of creating a token which has access just to a single repository.

This induces those risks:
* If I ever leave the organization, my token automatically loses write access to the repository, and thus breaks the publishing of new versions.
* If something goes wrong and some attacker gets hold of the token, that person can impersonate me on all repositories – both my personal ones, and ones in organizations where I have access.

We should think about setting up the build process in a way that a personal token is not needed. A possible solution might be using a repository-specific [deployment key](https://developer.github.com/v3/guides/managing-deploy-keys/#deploy-keys).

vadeg:
* API guidelines using GitHub Action
* Create tickets in Zally only from `master` branch

- [x] GitHub App needs to be created and configured in `zalando` organisation.

Fixes #568


tkrop:
:+1:

vadeg:
👍

PR #666, opened 2021-06-02
Clearly differ between OpenAPI security scheme types and OAuth2 

tfrauenstein:
Update to avoid confusion around OAuth 2.0 and OpenAPI security scheme types:
In the OpenAPI specification, `Bearer` is defined as http and not as oauth2 typed security scheme.


chapters/security.adoc
```diff
@@ -3,23 +3,20 @@
...
-== {MUST} secure endpoints with OAuth 2.0
+== {MUST} secure endpoints 
 
-Every API endpoint *must* be secured using OAuth 2.0. Please refer to the
-https://swagger.io/docs/specification/authentication/[OpenAPI Authentication]
-section on how to specify security definitions in your API.
+Every API endpoint must be protected and armed with authentication and authorization. 
+As part of the API definition you must specify how you protect your API using 
+one of the security schemes defined in the
+https://swagger.io/docs/specification/authentication/[OpenAPI Authentication Specification].
```
tkrop, line 11: I think, we only want to support `OAuth2.0` and `Bearer Authentication` so far, so we should list this here somehow.

tfrauenstein: Agree, though I was initially not sure whether the guideline is the right place to fix aspects of our security architecture.

tkrop:
:+1:

vadeg:
👍

PR #670, opened 2021-06-07
fix: parameter name to be not id

tkrop:
Since `{id}` is predefined code fragment, we need to use another name to have it shown correctly. Also formatted source to use `80` character width.


tfrauenstein:
👍

tkrop:
:+1:

PR #672, opened 2021-06-28
fix: resource locations (638)

### LINKED ISSUE #638: External references to header resources are incorrect
In https://github.com/zalando/restful-api-guidelines/pull/636#discussion_r576111657 @tfrauenstein exposed a bug in how we publish links to our models. They are published to `/` instead of `/models` resulting in a publication path of
```
https://opensource.zalando.com/restful-api-guidelines/request-headers-1.0.0.yaml#/X-Flow-ID
```
instead of
```
https://opensource.zalando.com/restful-api-guidelines/models/request-headers-1.0.0.yaml#/X-Flow-ID
```

While the first makes the path shorter, the second prevents mistakes by guideline writers and enables readers to browse the models on the web-server. However, we have already 3 other models published before. How do we want to fix this?

tkrop:
fixes #638.

I should have done this a long time ago, but now https://github.com/zalando/zally/pull/1259 has broken Zally for incorrectly linked resources, so we have to act quickly and change the location of the resources to the models folder. To stop the bleeding for correct APIs, we first fix the guideline setup now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,4 @@
+SHELL := /bin/bash 
 DOCKER := asciidoctor/docker-asciidoctor:latest
 DIRMOUNTS := /documents
 DIRCONTENTS := chapters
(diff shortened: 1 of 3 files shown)


piyushmor:
👍

tkrop:
:+

Issue #679, opened 2021-07-14
Update for OpenAPI 3.1


scop:
Any plans to update the guidelines for OpenAPI 3.1?

For example, the standard data formats section https://opensource.zalando.com/restful-api-guidelines/#238 would receive some new formats per https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#data-types and https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-7.3. I could submit a PR for this particular part if you think it would be useful and make it to the guidelines soon, let me know.

scop:
> I could submit a PR for this particular part if you think it would be useful and make it to the guidelines soon, let me know.

#680 does that.

tkrop:
@scop Thanks for taking the initiative here, and we very much appreciate it, but we currently waiting for the swagger library supporting OpenAPI 3.1 to align our tooling before adjusting the guidelines and prominently ask our users to migrate.

tkrop:
@scop thanks for the contribution. We now support data-formats from OpenAPI 3.1 but we need to wait for updating the general support for OpenAPI 3.1 until our tooling is ready to accept this format. So for now we close the issue.

PR #680, opened 2021-07-15
Update data formats per OpenAPI 3.1

scop:
`duration`, `idn-email`, `idn-hostname`, `uri-reference`,
`iri-reference`, and `relative-json-pointer` are new, from
draft-bhutton-json-schema-validation.

`period` was additionally added from RFC 3339 per the "MAY support
additional attributes using the other production names defined anywhere
in that RFC" statement in draft-bhutton-json-schema-validation.

`email`, `hostname`, `ipv4`, `ipv6`, `uri`, `uri-template`, `uuid`,
`time`, and `json-pointer` were updated as being from OpenAPI, as
they're now specified in draft-bhutton-json-schema-validation.

Refs #679

### DIFF
--- a/chapters/data-formats.adoc
+++ b/chapters/data-formats.adoc
@@ -58,8 +58,8 @@ standard media type for JSON, and make automated processing more difficult.
 [#238]
 == {MUST} use standard data formats
 
-https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#data-types[Open API] 
-(based on https://json-schema.org/understanding-json-schema/reference/string.html#format[JSON Schema])
+https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#data-types[Open API]
+(based on https://tools.ietf.org/html/draft-bhutton-json-schema-validation-00#section-7.3[JSON Schema Validation vocabulary])
 defines formats from ISO and IETF standards for date/time, integers/numbers and binary data.
 You *must* use these formats, whenever applicable:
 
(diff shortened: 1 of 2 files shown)


tkrop:
I'm hesitant adding OpenAPI 3.1.0 support yet, since the swagger library and subsequently our tooling does not support these changes.

tfrauenstein:
Even, if we do not yet support OpenAPI 3.1.0, we can guide using the data type formats defined in [JSON Schema Specification Draft 2020-12](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-7.3[JSON%20Schema%20Validation%20vocabulary]).

ePaul:
@tkrop:

> However, I'm hesitant adding OpenAPI 3.1.0 support yet, since the swagger library and subsequently our tooling does not support these changes.

@tfrauenstein:
> Even, if we do not yet support OpenAPI 3.1.0, we can guide using the data type formats defined in JSON Schema Specification Draft 2020-12.

We just (in the API guild meeting) had a discussion about whether allowing more formats might break the Zally validation (because that might be part of the schema used for validating an API definition).
I just tried it out with our running Zally instance: adding a `format` to (non-enum) string-attributes of a valid API does not make it invalid – neither for existing formats, new formats which were added in this PR, or even completely fantasy format names like `"Hello World"`.

_Interestingly, adding a `format: "Hello World"` to a property declared as `enum` triggers the warning about rule [107](https://zalando.github.io/restful-api-guidelines/#107), which is not there without that format (which seems like a bug): https://zally.zalando.net/editor/7c39582e-5873-481e-b7f1-46ed07a83f35_

tfrauenstein:
@ePaul  ok, Zally basically works as expected: string formats are supported and not (yet) restricted, but there is a Zally bug with formats for string typed enum attributes. 
Interesting observation about string with enum and format. For some formats (date, time, data-time) it works, for most others not (e.g. decimal, byte, binary, password). It does not work for 'hello world' -- but (even stranger!) it works if I change 'Hello world' to 'hello world' in the proceeding string format definition. Obviously a bug -- I propose to open a Zally Issue here. 

I will approve the PR though we stick to OpenAPI 3.0 as long as our tooling is not yet 100% OpenAPI 3.1 ready. However, with respect to data type formats we already can go with the richer and more consistent OpenAPI 3.1 definition as proposed by the PR (as it is backward compatible to 3.0).

tfrauenstein:
👍

scop:
Rebased, conflict resolved.

tfrauenstein:
👍

tkrop:
:+1:

PR #681, opened 2021-07-22
Improve the wording when to use PUT for creation

tfrauenstein:


chapters/http-requests.adoc
```diff
@@ -94,11 +94,12 @@ the resource was created)
 
 *Important:* It is good practice to prefer {POST} over {PUT} for creation of
 (at least top-level) resources. This leaves the resource ID management under 
-control of the service and not the client, and focus {PUT} on its usage for updates.
-However, in situations where {PUT} is used for resource creation,
-the resource IDs are maintained by the client and passed as a URL path segment.
+control of the service and not the client, and focuses {PUT} on its usage for updates.
+However, in situations where all resource attributes including the resource id (!) 
+are under control of the client as input for the resource creation you should use 
+{PUT} and pass the resource id as an URL path segment.
```
tkrop, line 100: I would suggest to use either *ID* or highlight that this is a field name by using `id` (or `{id}` if we think the reference is suitable here).

tfrauenstein: @tkrop done, thx.

tkrop:
:+1:

chapters/http-requests.adoc
```diff
@@ -93,12 +93,13 @@ resource was updated – with or without actual content returned), and {201} (if
 the resource was created)
 
 *Important:* It is good practice to prefer {POST} over {PUT} for creation of
-(at least top-level) resources. This leaves the resource ID management under 
-control of the service and not the client, and focus {PUT} on its usage for updates.
-However, in situations where {PUT} is used for resource creation,
-the resource IDs are maintained by the client and passed as a URL path segment.
+(at least top-level) resources. This leaves the resource identifier management under
```
vadeg, line 96: It is confusing that in one place it is _resource identifier_ and in other places _resource `id`_. I thinks we should select one option here.

tfrauenstein: @vadeg done, thx. Pls. approve.

tkrop:
:+1:

ePaul:
:+1:

PR #682, opened 2021-08-10
Use only standard HTTP status codes registered via IANA

tfrauenstein:
Be more specific about 'standard HTTP status code' -- must be standardized via RFC and registered via IANA Code Registry.

### DIFF
--- a/chapters/http-status-codes-and-errors.adoc
+++ b/chapters/http-status-codes-and-errors.adoc
@@ -44,20 +44,18 @@ service client and provider performance.
 [#150]
 == {MUST} use standard HTTP status codes
 
-You must only use official HTTP status codes standardized by RFCs consistently 
-with their intended semantics. 
-
-RFC standards define ~60 HTTP status codes with specific semantics all registered in the 
+You must only use official HTTP status codes consistently with their intended semantics. 
+Official HTTP status codes are defined via RFC standards and registered in the 
 https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml[IANA Status Code Registry].
-Main standards are {RFC-7231}#section-6[RFC7231] and {RFC-6585}[RFC 6585] (and there are upcoming new ones, e.g. 
+Main RFC standards are {RFC-7231}#section-6[RFC7231] and {RFC-6585}[RFC 6585] (and there are upcoming new ones, e.g. 
 https://tools.ietf.org/html/draft-tbray-http-legally-restricted-status-05[draft legally-restricted-status]).
-See an overview on all official error codes e.g. on
-https://en.wikipedia.org/wiki/List_of_HTTP_status_codes[Wikipedia] 
+An overview on the official error codes e.g. provides 
+https://en.wikipedia.org/wiki/List_of_HTTP_status_codes[Wikipedia: HTTP status codes] 
 (which also lists some unofficial codes e.g. defined by popular web servers like Nginx).
 
-Below we list the most commonly used and best understood HTTP status codes,
-consistent with their semantic in the RFCs. APIs should only use these to
-prevent misconceptions that arise from less commonly used HTTP status codes.
+Below we list the subset of official codes that is _most commonly used_ and _best understood_,
+consistent with their semantic in the RFCs. You should only use these HTTP status codes to
+prevent misconceptions that often arise from less commonly used codes.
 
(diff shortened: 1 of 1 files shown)


tkrop:
:+1:

tkrop:
@tfrauenstein why did you classify this as `guideline-change`? Technically it change nothing, since we provided and require that only status codes from the list are used.

tfrauenstein:
@tkrop thx. for the review. Yes, wrt. 'should only use codes listed below' it is no change.
However, it is a change wrt. 'must not use http codes' not registered via IANA. 
Updated, to make it clearer -- pls. re-approve, thx.

tkrop:
@tfrauenstein so the change is, that you open up the below list of status code usages. You are aware, that this change makes this rule not supportable for linting, since we would have to maintain a constantly growing list of `status code - method` combinations?

In my opinion the list we maintained comprises a best practices that API designers should be restricted too. Therefore, I vote for keeping it closed and make *this* clearer in the description instead of opening the rule up.

tfrauenstein:
@tkrop 

> so the change is, that you open up the below list of status code usages. You are aware, that this change makes this rule not supportable for linting, since we would have to maintain a constantly growing list of `status code - method` combinations?

No, I did not changed it and opened it for non listed codes. The rule currently says 'you must use standard codes' and 'you should use codes listed below'. I only made the first aspect more restrictive: 'you must use IANA registered codes (all defined by RFCs)'. This aspect is not yet checked by API linter. I do not propose to be more restrictive like 'you must only use codes listed below' as only rule statement.

chapters/http-status-codes-and-errors.adoc
```diff
@@ -44,21 +44,18 @@ service client and provider performance.
...
+https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml[IANA Status Code Registry].
+Main RFC standards are {RFC-7231}#section-6[RFC7231] and {RFC-6585}[RFC 6585] (and there are upcoming new ones, e.g. 
+https://tools.ietf.org/html/draft-tbray-http-legally-restricted-status-05[draft legally-restricted-status]).
+An overview on the official error codes e.g. provides 
+https://en.wikipedia.org/wiki/List_of_HTTP_status_codes[Wikipedia: HTTP status codes] 
+(which also lists some unofficial codes e.g. defined by popular web servers like Nginx).
+
+Below we list the subset of official codes that is _most commonly used_ and _best understood_,
+consistent with their semantic in the RFCs. You should only use these HTTP status codes to
+prevent misconceptions that often arise from less commonly used codes.
```
vadeg, line 58: This is not clear to me
```
You should only use these HTTP status codes to
prevent misconceptions that often arise from less commonly used codes.
```

I should use *only* status codes listed below or only in case of misconceptions?

tfrauenstein: @vadeg updated, pls. check. Thx!

tkrop:
Okay, so we still have consent on the guidance to comply to the restricted list implemented by Zally. Then I suggest to remove the label `guideline-changed` - since, in my opinion, it does not change.

tkrop:
:+1:

vadeg:
👍

chapters/http-status-codes-and-errors.adoc
```diff
@@ -44,22 +44,25 @@ service client and provider performance.
...
+You must only use official HTTP status codes consistently with their intended
+semantics. Official HTTP status codes are defined via RFC standards and
+registered in the https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml[IANA
+Status Code Registry]. Main RFC standards are {RFC-7231}#section-6[RFC7231]
+and {RFC-6585}[RFC 6585] (and there are upcoming new ones, e.g.
+https://tools.ietf.org/html/draft-tbray-http-legally-restricted-status-05[draft
+legally-restricted-status]). An overview on the official error codes provides
+https://en.wikipedia.org/wiki/List_of_HTTP_status_codes[Wikipedia: HTTP status
+codes] (which also lists some unofficial status codes, e.g. defined by popular
+web servers like Nginx, that we do not suggest to use).
```
tkrop, line 56: Sorry, I added `that we do not suggest to use` during the merge to emphasize that our reference here does not suggest to use the list from Wikipedia.

tkrop:
:+1:

tfrauenstein:
👍

PR #683, opened 2021-08-24
PUT with server-side generated attributes

tfrauenstein:
editorial -- make clear: PUT can be used, even some resource attributes are server-side generated


vadeg:
👍

PR #684, opened 2021-08-31
rename example:  order_id --> sales_order_number

tfrauenstein:
Example contributes to promote usage of more specific `sales_order_number`(instead of `order_number` used for different resources like sales_orders, shipment_orders, purchase_orders) naming.


vadeg:
👍

ePaul:
:+1:

tkrop:
:+1:

Issue #660, opened 2021-05-18
Editorial: Uniform Layout and Naming of Hints, Remarks, Exceptions etc.


tfrauenstein:
We use different layout formats for standard guideline elements like 'Exception:', 'Hint:', 'Note:', 'Remark:', 'Example:', 'Tip', 'Reasoning:', 'Warning:', 'Important:'. Additionally, the elements are partially redundant with unclear 'didactic purpose'. 
Let us reduce the elements to a minimal set with clear purpose and uniform layout. 
It will contribute to readability and efficient consumption by API builders. Examples: 
* Hint: https://opensource.zalando.com/restful-api-guidelines/#101 
* Example: https://opensource.zalando.com/restful-api-guidelines/#116   https://opensource.zalando.com/restful-api-guidelines/#104   https://opensource.zalando.com/restful-api-guidelines/#114
* Note + Tip: https://opensource.zalando.com/restful-api-guidelines/#114
* Reasoning: https://opensource.zalando.com/restful-api-guidelines/#114

vadeg:
I propose to use several features provided by Asciidoc for formatting: 
1. [Admonitions](https://docs.asciidoctor.org/asciidoc/latest/blocks/admonitions/) for hints, exceptions and etc.
2. [Sidebars](https://docs.asciidoctor.org/asciidoc/latest/blocks/sidebars/) to highlight a block of text.

ePaul:
From the meeting:

* We agree that unifying this is useful
* @vadeg will add a proposal on how to do this (like 3-4 different things)

vadeg:
I have collected a number of occurrences of different elements and mapped them to possible replacements from AsciiDoctor
|Current element | Occurrences|AsciiDoctor Admonition element |
|-|-|-|
|Hint | 32 | TIP or NOTE |
|Example | 10 | Implement new EXAMPLE admonition |
|Note | 47 |  NOTE |
|TIP | 1 | TIP |
|Reasoning | 1 | NOTE | 
|Exception | 7 | IMPORTANT|
|Warning | 1 | WARNING |

ePaul:
@tfrauenstein and @vadeg : please coordinate whether to do this or #674 first, as there will be a lot of conflicts.

tkrop:
[...]

PR #693, opened 2021-11-10
Update introduction.adoc to use RFC 2119 terminology everywhere

sdousti:
Following RFC 2119 terminology everywhere helps in removing any ambiguity.

### DIFF
--- a/chapters/introduction.adoc
+++ b/chapters/introduction.adoc
@@ -55,12 +55,12 @@ work evolves, but teams can confidently follow and trust them.
 
 In case guidelines are changing, following rules apply:
 
-* existing APIs don't have to be changed, but we recommend it
-* clients of existing APIs have to cope with these APIs based on
+* it is RECOMMENDED that the existing APIs be changed
+* clients of existing APIs MUST cope with these APIs based on
 outdated rules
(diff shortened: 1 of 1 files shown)


chapters/introduction.adoc
```diff
@@ -55,12 +55,12 @@ work evolves, but teams can confidently follow and trust them.
...
+* clients of existing APIs MUST cope with these APIs based on
 outdated rules
-* new APIs have to respect the current guidelines
+* new APIs MUST respect the current guidelines
 
 Furthermore you should keep in mind that once an API becomes public
-externally available, it has to be re-reviewed and changed according to
-current guidelines - for sake of overall consistency.
+externally available, it MUST be re-reviewed and changed according to
+current guidelines - for the sake of overall consistency.
```
tkrop, line 65: As much as I like your rewording, the use of capital letter `MUST` and `RECOMMENDED` is currently not consistent with our usage in text - only in rule titles. If you change it to *lower case bold*, we would merge it for. Else we would need to agree on a new style before we can merge your suggestions (see also [comment on #660](https://github.com/zalando/restful-api-guidelines/issues/660#issuecomment-988037890)).

Issue #674, opened 2021-06-29
Restructuring the Guideline Sections


tfrauenstein:
Today it is not easy to find the relevant rules, because we do not have a clear structuring concept.

We should separate API Endpoint related rules, from overall design related rules.
The former should be oriented on the basic syntactic categories (analog to OpenAPI structure), like Data Formats, Resource Paths and Query Parameters, JSON Payload, HTTP requests / responses, HTTP status codes and errors, Headers. The latter should start with the basics like Hypermedia, Performance and end with the more specific design aspects like Compatibility and Deprecation (proposal see [internal link](https://docs.google.com/document/d/1zOvQHVyPyinmT6JdYk1NWYBk-yBcaWMm98_n5hISA-o/edit)).

ePaul:
@tfrauenstein and @vadeg : please coordinate whether to do this or #660 first, as there will be a lot of conflicts.

tkrop:
I think this is done, so lets close this.

PR #696, opened 2021-11-16
Restructuring the Guidelines

tfrauenstein:
implements issue: https://github.com/zalando/restful-api-guidelines/issues/674


chapters/events.adoc
```diff
@@ -322,129 +261,320 @@ with supplemental information, such as how events are partitioned in the
...
+defines a field to be of a different type that was already being
+emitted, or, is changing the type of an undefined field. Both of these
+are prevented by not using `additionalProperties`.
+
+See also rule <<111>> in the <<compatibility>> section for further guidelines 
+on the use of `additionalProperties`.
+
+
+[#246]
+== {MUST} use semantic versioning of event type schemas
```
ePaul, line 422: I'm not sure whether it makes sense to have this as a {{MUST}} rule – it's automatically done by Nakadi (as users can't set the version numbers themselves, they are set by Nakadi when updating the event type's schema).

One can argue that it would apply for other kinds of event registries, but we don't have any (and even if we had, those should also do it in a similar way).
We can keep it as-is for now, but maybe we want to introduce a different modal verb (or equivalent) for things which are done by infrastructure instead of by engineers?

chapters/events.adoc
```diff
@@ -322,129 +261,320 @@ with supplemental information, such as how events are partitioned in the
...
+        - C: Creation of an entity
+        - U: An update to an entity.
+        - D: Deletion of an entity.
+        - S: A snapshot of an entity at a point in time.
+----
+
+The Data Change Event Category is structurally different to the General
+Event Category defining a field called `data` as container for the custom
+payload, as well as specific information related to data
+changes in the `data_type`.
```
ePaul, line 557:
```suggestion
The Data Change Event Category is structurally different to the General
Event Category by defining a field called `data` as container for the custom
payload, as well as specific information related to data
changes in the `data_op`. 
```
(Unrelated to the restructuring, you can ignore if you think it needs discussion.)

chapters/events.adoc
```diff
@@ -692,6 +814,84 @@ are a "should", they can be read as "must, unless you have a very good
...
+Similar to API permission scopes, there will be event type permissions
+passed via an OAuth token supported in near future. In the meantime,
+teams are asked to note the following:
+
+* Sensitive data, such as (e-mail addresses, phone numbers, etc) are
+subject to strict access and data protection controls.
+* Event type owners *must not* publish sensitive information unless it's
+mandatory or necessary to do so. For example, events sometimes need to
+provide personal data, such as delivery addresses in shipment orders (as
+do other APIs), and this is fine.
```
ePaul, line 834: _Not new, feel free to ignore for now. I'm just writing it down so it doesn't get lost._

Nakadi decided to not go with scopes for event type authorization in 2017 (instead using application-name based authorization), and I don't think they intend to change it ever, certainly not "in near future".
And there are plenty of event types which do contain sensitive data.
Do we want to adjust this rule?

chapters/events.adoc
```diff
@@ -692,6 +814,84 @@ are a "should", they can be read as "must, unless you have a very good
...
+processing can be supported by sending the following information:
+
+* the process/resource/entity identifier,
+* a <<203,monotonically increasing ordering key>> and
+* the process/resource state after the change.
+
+A receiver that is interested in the current state can then ignore
+events that are older than the last processed event of each resource. A
+receiver interested in the history of a resource can use the ordering
+key to recreate a (partially) ordered sequence of events.
```
ePaul, line 878: _Again, ignore this for now, I'm just adding it here so I don't forget it._

A third option (which I was made aware recently in a private chat) was an event-sourcing-like one:
* store all events when they come in, and process them.
* if events are detected as out-of-order:
   * roll back your state to some older safe state
   * reprocess all events from this point in the event order on.

(The order would likely be per-entity.)

chapters/http-headers.adoc
```diff
@@ -290,7 +296,205 @@ size.
...
+when domain-specific generic context information...
+
+. needs to be passed end to end along the service call chain (even if 
+not all called services use it as input for steering service behavior 
+e.g. {X-Sales-Channel} header) and/or...
+. is provided by specific gateway components, for instance, our 
+Fashion Shop API or Merchant API gateway. 
+
+Below, we explicitly define the list of proprietary header exceptions usable for 
+all services for passing through generic context information of our fashion domain (use case a).
```
ePaul, line 327: This is also not new, but I've just noticed it: The two bullet points above are rendered in the guidelines (as well as Github's preview) as `1.` and `2.`, but the paragraph after it refers to the first of them as `a`.

index.adoc
```diff
@@ -233,20 +233,17 @@ include::chapters/design-principles.adoc[]
 include::chapters/general-guidelines.adoc[]
 include::chapters/meta-information.adoc[]
 include::chapters/security.adoc[]
-include::chapters/compatibility.adoc[]
-include::chapters/deprecation.adoc[]
-include::chapters/json-guidelines.adoc[]
 include::chapters/data-formats.adoc[]
-include::chapters/common-data-types.adoc[]
```
ePaul, line 240: The file is not included in the index anymore (the content seems to be included in the JSON guidelines and the pagination section), but I don't see it deleted in the change list. Is that intentional?

ePaul:
:+1:

chapters/events.adoc
tfrauenstein: done

chapters/events.adoc
tfrauenstein: let's discuss this outside this Restructuring PR -- created an entry in the API Guild meeting agenda to follow up

chapters/events.adoc
tfrauenstein: let's discuss this outside this Restructuring PR -- created an entry in the API Guild meeting agenda to follow up

chapters/events.adoc
tfrauenstein: let's discuss this outside this Restructuring PR -- created an entry in the API Guild meeting agenda to follow up

chapters/http-headers.adoc
tfrauenstein: done

ePaul:
:+1:

chapters/compatibility.adoc
```diff
@@ -132,17 +104,32 @@ can be removed from the API definition as long as the server ignores this
...
-backwards compatibility. See <<161>> for an example.
+Service clients should apply the robustness principle:
 
-Maps (see <<216>>), even though technically objects, are also forbidden as top
-level data structures, since they don't support compatible, future extensions.
+* Be conservative with API requests and data passed as input, e.g. avoid to
+  exploit definition deficits like passing megabytes of strings with
+  unspecified maximum length.
+* Be tolerant in processing and reading data of API responses, more
+  specifically...
```
vadeg, line 116: Is there something missing?

chapters/data-formats.adoc
```diff
@@ -207,3 +154,50 @@ to support clients specifying via the standard HTTP headers
...
+use cases. UUIDs solve this problem, as they can be generated without
+collisions in a distributed, non-coordinated way and without additional server
+round trips.
+
+However, they also come with some disadvantages:
+
+* pure technical key without meaning; not ready for naming or name scope
+  conventions that might be helpful for pragmatic reasons, e.g. we learned to
+  use names for product attributes, instead of UUIDs
+* less usable, because...
```
vadeg, line 172: The end of the sentence is missing.

chapters/compatibility.adoc
```diff
@@ -313,3 +268,49 @@ coordinating version upgrades — especially with hyperlinked service
...
+[#110]
+== {MUST} always return JSON objects as top-level data structures
+
+In a response body, you must always return a JSON object (and not e.g. an
+array) as a top level data structure to support future extensibility. JSON
+objects support compatible extension by additional attributes. This allows you
+to easily extend your response and e.g. add pagination later, without breaking
+backwards compatibility. See <<161>> for an example.
+
+Maps (see <<216>>), even though technically objects, are also forbidden as top
```
vadeg, line 282: You mean it is forbidden to have something like this?
```
MyType:
  type: object
  additionalProperties:
    anotherObject: 
      type: object
      ....
```

vadeg (commented):
Overall having `REST` prefix in every chapter, e.g. `REST Basics`, `REST Design` look redundant to me. Why not to keep chapter names simple?

chapters/compatibility.adoc
tfrauenstein: not changed by this PR.  However, indention was missing -- added.

chapters/data-formats.adoc
tfrauenstein: dito: not changed by this PR.  However, indention was missing -- added.

chapters/compatibility.adoc
tfrauenstein: Also not changed by this PR. It is not allowed, since a map object should not have other fields than the key and value. I think the description is ok, but improvement PRs are welcome.

tfrauenstein:
@vadeg Actually we need one additional section top level, like "REST API Basics" and sub sections "Metadata".
However, I wanted to avoid changing most rules (currently section level 2) as well as the toc layout, and having rules on different levels. So I flattened the new top section as prefix resulting in the redundancy you mentioned. It may look queer at first glance, but IMO looks ok and is a good compromise solution.

vadeg:
👍

chapters/compatibility.adoc
```diff
@@ -113,22 +113,20 @@ Service clients should apply the robustness principle:
   exploit definition deficits like passing megabytes of strings with
   unspecified maximum length.
 * Be tolerant in processing and reading data of API responses, more
-  specifically...
+  specifically srvice clients must be prepared for compatible API extensions
```
ePaul, line 116:
```suggestion
  specifically, service clients must be prepared for compatible API extensions 
```

ePaul: (Feel free to ignore and merge first, I can open a separate PR to fix this.)

ePaul:
:+1:

chapters/compatibility.adoc
ePaul: @vadeg You can have a map with objects as values (not as keys), though the syntax for this would be:

```yaml
MyType:
  type: object
  additionalProperties:
    type: object
    ....
```
But those maps should contain only their values, not additional fields, and as such they are not as extensible as we would like for top-level objects, and they should be wrapped in an additional object.

chapters/compatibility.adoc
tfrauenstein: I will do it via separate PR, thx!

PR #698, opened 2021-12-08
Avoid duplicate/nested assets and models dirs on multiple `make assets`

scop:
Before:
```shellsession
$ make clean assets; make assets
$ find output/assets output/models -type d
output/assets
output/assets/assets
output/models
output/models/models
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,8 +49,8 @@ next-rule-id:
 
 assets:
 	mkdir -p $(DIRBUILDS);
-	cp -r assets $(DIRBUILDS)/assets;
-	cp -r models $(DIRBUILDS)/models;
+	cp -r assets $(DIRBUILDS)/;
+	cp -r models $(DIRBUILDS)/;
 	cp -r models/{problem-1.0.{0,1},money-1.0.0}.yaml $(DIRBUILDS);
 	cp -r -n legacy/* $(DIRBUILDS);
 


Makefile
```diff
@@ -49,8 +49,8 @@ next-rule-id:
 
 assets:
 	mkdir -p $(DIRBUILDS);
-	cp -r assets $(DIRBUILDS)/assets;
-	cp -r models $(DIRBUILDS)/models;
+	cp -r assets $(DIRBUILDS)/;
+	cp -r models $(DIRBUILDS)/;
 	cp -r models/{problem-1.0.{0,1},money-1.0.0}.yaml $(DIRBUILDS);
 	cp -r -n legacy/* $(DIRBUILDS);
```
tkrop, line 55: @scop thank you for your  contribution. I'm wondering on which OS you are building?

The current code is working fine on our build hosts as well as on my local machine. Else the following link should serve the only asset we have:

https://opensource.zalando.com/restful-api-guidelines/assets/assets/api-zalando-small.jpg

scop: Ubuntu 20.04. It's not about the OS, if one runs `make assets` **twice or more** without `clean` in between as noted in the reproducer recipe, it should produce the duplicate dirs no matter the OS as long as it's a UNIXy one. No problem when `make assets` is run only once.

tkrop:
:+1:

Makefile
tkrop: May be we should use `-H "Accept: application/vnd.github.v3+json"`. I now remember, it was in test when I originally implemented this.

tfrauenstein:
👍

PR #699, opened 2021-12-08
Fix `check-rules-duplicates` recipe indentation

scop:
Makefile recipes should be indented with a leading tab, this target had
tabs vs spaces in the wrong order.


tfrauenstein:
👍

tkrop:
:+1:

tkrop:
@scop Thanks for your contribution.

PR #707, opened 2022-01-03
clean-up 'internal links'

tfrauenstein:
updated: links to new API Guild docu
editorial: make sure that all internal links are grep-able (i.e. single lined) for easy checks

### DIFF
--- a/chapters/events.adoc
+++ b/chapters/events.adoc
@@ -27,8 +27,7 @@ event type and the schema for its custom data.
 
 The basic model described above was originally developed in the
 https://github.com/zalando/nakadi[Nakadi project], which acts as a
-reference implementation (see also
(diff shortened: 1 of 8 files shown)


tkrop:
@tfrauenstein how about naming all internal links like `:internal-...` and referencing the name instead of the link. This would also make it easier to maintain and list them.

tfrauenstein:
@tkrop thx. for the proposal, but maintenance even would get harder because the internal links vary in link reference + text such that basically each internal link occurance needs to be abstracted via  `internal-..` constant without its re-usage.

tkrop:
@tfrauenstein I didn't mean to include the text in the reference - only the link.

tfrauenstein:
@tkrop ok done, pls. approve

ePaul:
:+1:

tkrop:
:+1:

PR #706, opened 2021-12-22
feat: migrate master to main (#706)

tkrop:
fixes #706.

Build is failing since we migrated from `master` to `main`, because github actions still referred to `master` branch.

PR #711, opened 2022-01-10
More master→main fixes

ePaul:
* rename master→main in one reference to this repository (leftover from #706)
* rename master→main in links to the OpenAPI spec repository
* add a comment to one URL which is broken

Ironically, the other occurrence of `master` in an URL (some lines above the fix) was already broken due to another change, and also in an example of why not to use such links, so I decided to just leave it here.

There are two `master` URLs left, they are for a Zalando-internal repository which was not yet changed.

### DIFF
--- a/chapters/general-guidelines.adoc
+++ b/chapters/general-guidelines.adoc
@@ -88,7 +88,7 @@ should not contain references to local or remote content, e.g. `../fragment.yaml
 `$ref: 'https://github.com/zalando/zally/blob/master/server/src/main/resources/api/zally-api.yaml#/schemas/LintingRequest'`.
 The reason is, that the content referred to is _in general_ *not durable* and
 *not immutable*. As a consequence, the semantic of an API may change in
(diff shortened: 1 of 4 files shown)


chapters/general-guidelines.adoc
```diff
@@ -88,7 +88,7 @@ should not contain references to local or remote content, e.g. `../fragment.yaml
 `$ref: 'https://github.com/zalando/zally/blob/master/server/src/main/resources/api/zally-api.yaml#/schemas/LintingRequest'`.
 The reason is, that the content referred to is _in general_ *not durable* and
 *not immutable*. As a consequence, the semantic of an API may change in
-unexpected ways.
+unexpected ways. (For example, the second link is already outdated due to code restructuring, it would now need to be `$ref: 'https://github.com/zalando/zally/blob/main/server/zally-server/src/main/resources/api/zally-api.yaml#/schemas/LintingRequest'`.)
```
ePaul, line 91: I'm not sure whether we actually want the new URL here, given this is just an example. We could also just leave it out:

```suggestion
unexpected ways. (For example, the second link is already outdated due to code restructuring.)
```

tfrauenstein:
👍

PR #712, opened 2022-01-11
replace dead letter queue

ePaul:
Avoid unnecessary "violent language".
In this context, what we were talking about was not even an actual dead letter queue.


ePaul:
:+1:

PR #714, opened 2022-01-24
Exclude 'type' from common field names

tfrauenstein:
Reasoning -- as discussed in the API Guild meeting: 
- its purpose is to be a discriminator for polymorphic types 
- not needed anymore since discriminators are supported by OpenAPI 3.x -- see https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/ 
- current definition and documentation is poor


tkrop:
:+1:

ePaul:
:+1:

PR #115, opened 2016-08-03
Changed assessment of Collection+JSON docs

whiskeysierra:
Fixes #112

PR #715, opened 2022-01-25
More clarity about when error spec definitions can be omitted

tfrauenstein:
I saw a couple of API defining endpoints that are not straightforward to use (e.g. different optional input arguments that only make sense in specific combinations) or with tricky error situations that have not been specified and just addressed by generic 'default' error definition. Update of [Rule #115](https://opensource.zalando.com/restful-api-guidelines/#151) should provide more clarity.

### DIFF
--- a/chapters/http-status-codes-and-errors.adoc
+++ b/chapters/http-status-codes-and-errors.adoc
@@ -25,7 +25,7 @@ service client and provider performance.
 
 **Exception:** Standard errors, especially for client side error codes 
 like 401 (unauthenticated), 403 (unauthorized) or 404 (not found) that can be 
-straightforward inferred from the specific endpoint definition need not to be 
+inferred straightforwardly from the specific endpoint definition need not to be 
 individually defined. Instead you can combine multiple error response specifications 
 with the default pattern below. However, you should not use it and explicitly 
 define the error code as soon as it provides endpoint specific indications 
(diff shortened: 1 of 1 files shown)


chapters/http-status-codes-and-errors.adoc
```diff
@@ -9,18 +9,25 @@ APIs should define the functional, business view and abstract from
 implementation aspects. Success and error responses are a vital part to
 define how an API is used correctly.
 
-Therefore, you must define **all** success and service specific error
+Therefore, you must define **all** success and service specific error situation
```
tkrop, line 12: `situations` sounds strange here. Sentence before was clearer.

tfrauenstein: 👍  
Added one additional sentence to make clearer what I had in mind.

chapters/http-status-codes-and-errors.adoc
```diff
@@ -9,18 +9,28 @@ APIs should define the functional, business view and abstract from
 implementation aspects. Success and error responses are a vital part to
 define how an API is used correctly.
 
-Therefore, you must define **all** success and service specific error
+Therefore, you must define **all** success and service specific error
```
ePaul, line 12: remove space:
```suggestion
Therefore, you must define **all** success and service specific error
```

PR #719, opened 2022-03-17
Problem definition examples conflict with application contexts

tfrauenstein:
Deleted or substituted the examples with generic values.
The problem definition is generic and we guide to use it in all error return situations -- see https://opensource.zalando.com/restful-api-guidelines/#176 
Concrete example values in the problem definition lead to confusion because they conflict with the values valid in the application context of specific http status response definitions.

### DIFF
--- a/models/problem-1.0.1.yaml
+++ b/models/problem-1.0.1.yaml
@@ -10,14 +10,14 @@ Problem:
         it is neither recommended to be dereferenceable and point to a
         human-readable documentation nor globally unique for the problem type.
       default: 'about:blank'
-      example: '/problem/connection-error'
+      example: '/some/uri-reference'   
     title:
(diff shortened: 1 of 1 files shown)


models/problem-1.0.1.yaml
```diff
@@ -10,14 +10,14 @@ Problem:
         it is neither recommended to be dereferenceable and point to a
         human-readable documentation nor globally unique for the problem type.
       default: 'about:blank'
-      example: '/problem/connection-error'
+      example: '/some/uri-reference'
```
ePaul, line 13:
```suggestion
      example: '/some/uri-reference'
```

ePaul:
:+1:

tfrauenstein:
👍

PR #726, opened 2022-05-24
Define `next` page link as optional

tfrauenstein:
Minor update: next (and prev) pagination attributes are optional in case of last (or first) page returned.


tkrop:
:+1:

PR #730, opened 2022-05-31
fix(data-formats): reference to IPv6 RFC (4291, not 2673)

scop:
https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-7.3.4


ePaul:
So it turns out that there is no proper definition of the 4-digit IPv4 address in one of the IPv4 RFCs.
The Internet draft [draft-main-ipaddr-text-rep](https://datatracker.ietf.org/doc/html/draft-main-ipaddr-text-rep-02) tried to remediate this, but it never got into the RFC stage.

It copied the definition from [RFC 3986 (URI syntax), section 3.2.2 Host](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2). So I guess we can also just refer to that one.

ePaul:
Though arguably, we should just use the same list of format definitions as JSON schema.

ePaul:
:+1:

tfrauenstein:
👍

PR #736, opened 2022-07-25
Feature/peter/enable 2816 api first principle

caliginouspierson:
Prefer NZ Standards Language:
"Should be designed before implementation"


caliginouspierson:
Apologies.

Issue #709, opened 2022-01-10
some formatting can't be converted to PDF


ePaul:
*Problem*
Pages 10 (durable references) and 44/45 (common method properties table) of the PDF show some pieces of text as plain HTML instead of formatted properly.

*Analysis*

When running the `make all` locally, I get some errors:
```
...
docker run -v /home/pebermann/projektoj/restful-api-guidelines:/documents/ asciidoctor/docker-asciidoctor:latest asciidoctor \
  -D /documents/output index.adoc;
docker run -v /home/pebermann/projektoj/restful-api-guidelines:/documents/ asciidoctor/docker-asciidoctor:latest asciidoctor-pdf \
  -D /documents/output index.adoc;
asciidoctor: ERROR: failed to parse formatted text: ​<code><a href="https://infrastructure-api-repository.zalandoapis.com/" class="bare">https://infrastructure-api-repository.zalandoapis.com/</a></code> <b>–</b> used to refer to user defined, immutable API specification revisions published via the internal API repository.
asciidoctor: ERROR: failed to parse formatted text: ​<code><a href="https://opensource.zalando.com/restful-api-guidelines/{model.yaml}" class="bare">https://opensource.zalando.com/​restful-api-guidelines/​{model.yaml}</a></code> <b>–</b> used to refer to guideline defined re-usable API fragments (see <code>{model.yaml}</code> files in <a href="https://github.com/zalando/restful-api-guidelines/tree/master/models">restful-api-guidelines/models</a> for details).
asciidoctor: ERROR: failed to parse formatted text: <span class="should-keyword">&#x26a0;&#xFE0F;</span> No, but <a anchor="229"><span class="should-keyword"><strong>SHOULD</strong></span> consider to design <code>POST</code> and <code>PATCH</code> idempotent</a>
asciidoctor: ERROR: failed to parse formatted text: <span class="should-keyword">&#x26a0;&#xFE0F;</span> May, but only if specific <a href="#post"><code>POST</code></a> endpoint is <a anchor="safe">safe</a>. <strong>Hint:</strong> not supported by most caches.
asciidoctor: ERROR: failed to parse formatted text: <span class="should-keyword">&#x26a0;&#xFE0F;</span> No, but <a anchor="229"><span class="should-keyword"><strong>SHOULD</strong></span> consider to design <code>POST</code> and <code>PATCH</code> idempotent</a>
[...]

PR #739, opened 2022-09-09
Add separate fonts and theme for PDF generation

vadeg:
Fix symbols rendering in PDF generation. 

* Add custom theme for PDF
* Add custom fonts for PDF
* Change symbols to similar, but which are supported by both fonts.
* Add system files to `.gitignore`

The result will look like this:

### PDF

Screenshot 2022-09-09 at 11 10 46

### HTML

Screenshot 2022-09-09 at 11 11 01

I have to replace symbols in HTML in order to make it work correct in both PDF and HTML. Let me know what do you think.

Let me know what do you think about this.
Refers to #709


tfrauenstein:
great -- PDF version now looks fine and HTML unchanged. Thx!

tkrop:
:+1:

tfrauenstein:
👍

PR #751, opened 2023-01-30
Link proprietary headers to 184

tor-vs-floki:
Proprietary headers are specified above this section, but that doesn't mean it is impossible to miss 183 when you are navigating directly to 184.


tfrauenstein:
👍

tor-vs-floki:
👍

Issue #742, opened 2022-10-18
Re-review the status codes


ePaul:
With [RFC 9110 (HTTP semantics)](https://www.rfc-editor.org/rfc/rfc9110.html), some status codes are now more clearly specified.

We should look into whether to update our status code list.

Candidates:
* 307, 308, 422, 417

ePaul:
Also (noted by @tkrop) 409 → 412 seems often better.

ePaul:
Fixed by #781.

PR #756, opened 2023-04-12
Reformat status codes into sections

SmilyOrg:
* This does not change any status code body text (apart from removing the "name" part at the beginning)
* It does not add any new status codes (e.g. from #742) 
* It _does_ add more usage and documentation hints based on internal discussions
* Links to RFCs have been added for all status codes with RFC 9110 taking precedence, filled out by RFC 4918 and RFC 6585 for the missing codes

The format, terminology used, and the usage/documentation hints are all up for discussion. The status code text bodies we can review after cleaning up the internal working doc, perhaps as a follow-up PR.

Bonus: add `make watch` for fun and profit

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ DIRWORK := $(shell pwd -P)
 
 .PHONY: all clean install lint format pull assets rules html pdf epub force
 .PHONY: check check-rules check-rules-duplicates check-rules-incorrects
-.PHONY: next-rule-id
+.PHONY: next-rule-id watch
 
 all: clean html pdf epub rules
 clean:
@@ -61,6 +61,9 @@ rules: check-rules
 html: check assets pull
 	docker run -v $(DIRWORK):$(DIRMOUNTS)/ ${DOCKER} asciidoctor \
 	  -D $(DIRMOUNTS)/$(DIRBUILDS) index.adoc;
+
+watch:
+	watchexec --exts adoc,css --ignore output -r make html
 
 pdf: check pull
 	docker run -v $(DIRWORK):$(DIRMOUNTS)/ ${DOCKER} asciidoctor-pdf -v \
--- a/index.adoc
+++ b/index.adoc
@@ -5,6 +5,7 @@
 :leveloffset: +1
 :sectlinks:
 :sectnumlevels: 1
+:icons: font
 
 :creator: Zalando SE
 :producer: Asciidoctor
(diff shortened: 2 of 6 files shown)


index.adoc
```diff
@@ -140,6 +140,88 @@
...
+:doc-may: pass:quotes[link:#status-code-doc-may[[.status-code-hint.may]#✍️ MAY document#]]
+
+// RFC Links
+// 
+// You can regenerate them by pasting the adjacent JavaScript one-liner into the browser console while having the RFC website open
+
+// 
+// RFC 9110 - https://www.rfc-editor.org/rfc/rfc9110.html
+// 
+// var u = new URL(document.URL); u.hash = ""; console.log(Array.from(document.querySelectorAll('.section-name[href^="#name-"]')).map(s => ({ m: s.innerText.match(/^(\d{3}) (.*)/), href: s.attributes["href"].value })).filter(c => c.m).map(c => `:rfc-status-${c.m[1]}: pass:quotes[${u}${c.href}[^RFC^^]]`).join("\n"))
```
tkrop, line 157: Interesting. How is this executed? Can you add a documentation to this script snippet?

index.adoc
```diff
@@ -140,6 +140,88 @@
...
+:rfc-status-100: pass:quotes[https://www.rfc-editor.org/rfc/rfc9110.html#name-100-continue[^RFC^^]]
```
tkrop, line 158: I'm wondering whether it is possible to make use of the `:RFC-9110: https://tools.ietf.org/html/rfc9110` pattern that is missing in line `55` of this file.

index.adoc
```diff
@@ -105,7 +105,7 @@
 :HEAD: pass:[<a href="#head"><code>HEAD</code></a>]
 :OPTIONS: pass:[<a href="#options"><code>OPTIONS</code></a>]
 :TRACE: pass:[<a href="#trace"><code>TRACE</code></a>]
-:ALL: pass:[<code>&lt;all&gt;</code>]
+:ALL: pass:[<code title="Applicable to all HTTP methods.">&lt;all&gt;</code>]
```
tkrop, line 108: Can you make sure that this is exactly looking like `POST`, etc.

index.adoc
```diff
@@ -140,6 +140,88 @@
 :501: pass:[<a href="#status-code-501" class="status-code">501</a>]
 :503: pass:[<a href="#status-code-503" class="status-code">503</a>]
 
+// See http-status-codes-and-errors.adoc for explanation
+:use-mustnot: pass:quotes[link:#status-code-use-mustnot[[.status-code-hint.mustnot]#❌ MUST NOT use#]]
+:use-should: pass:quotes[link:#status-code-use-should[[.status-code-hint.should]#✅ SHOULD use#]]
+:doc-should: pass:quotes[link:#status-code-doc-should[[.status-code-hint.should]#✍️ SHOULD document#]]
+:doc-shouldnot: pass:quotes[link:#status-code-doc-shouldnot[[.status-code-hint.shouldnot]#✍️ SHOULD NOT document#]]
+:doc-may: pass:quotes[link:#status-code-doc-may[[.status-code-hint.may]#✍️ MAY document#]]
```
tkrop, line 148: I'm not convinced about the `MUST`, `SHOULD`, `MAY` usage as well as the icons here. I would suggest to simplify the badges, `use` and `document` and rely on colors (`red=do not`, `green=do`) to represent the suggestion. I would follow Yoda an try to stay away from unclear suggestions like `MAY`, however, if we really need a weaker abstraction, I would suggest to use `orange=weak do not` and `gree-yellow=weak do`.

ePaul: Color-only is bad for people with color vision weaknesses. It needs at least an icon.

index.adoc
SmilyOrg: The `<all>` has existed already, I just added a `title`. Maybe a better way to go would be to make it look like `ALL` and then have a linked section somewhere that describes what it means? 🤔

index.adoc
SmilyOrg: You know, I'm also not convinced on them. Especially `SHOULD document` vs `MAY document`, they're kind-of the same, so maybe we can just stick with two states - do/don't. I'm with @ePaul though that we cannot just have color (it also doesn't print well). We need to represent two things:
1. **Type**: what the badge is referring to, currently that's "use" or "document" & icon
2. **Guidance**: what we recommend, currently that's SHOULD/MUST/... & icon (for use only) & color

Each should be represented in a way that's not just color, but it can be a combination of words, icons, and color. It doesn't have to be _too_ explicit as long as it's easily recognizable, as we link to the legend with explanation anyway.

index.adoc
SmilyOrg: With the page open, you can press Ctrl+Shift+I or F12 to open the browser dev tools, then paste the onliner in the Console there and it'll spit out the definitions below.

I wasn't sure how much instruction to leave here or if I should even include it, but there's a hint on line 152. I can make it more detailed and/or add the instruction before each one-liner, wdyt?

index.adoc
SmilyOrg: Ah, good point, lemme try that!

SmilyOrg: It works! Fixed in latest commit.

Makefile
```diff
@@ -62,6 +62,9 @@ html: check assets pull
 	docker run -v $(DIRWORK):$(DIRMOUNTS)/ ${DOCKER} asciidoctor \
 	  -D $(DIRMOUNTS)/$(DIRBUILDS) index.adoc;
 
+watch:
+	watchexec --exts adoc,css --ignore output -r make html
```
tkrop, line 66: What is this for?

index.adoc
SmilyOrg: How about this?

image
image

tkrop: Quite good - makes sense -, but I think we do not need upper case letters for `do not use` - the red is giving enough attention. I would also abstain to put a red `do not use` and a yellow `document` in one line. This sends wrong signals. We better should add the documentation necessity to the `do not use` explanation I think.

Makefile
tkrop: That would be great. Interesting extension.

SmilyOrg:
Added some more fixes based on the comments above. Here are a few screenshots of what it looks like right now. I've also rebased the branch to `status-codes` (which is the new feature branch based on current main).

image
image

tkrop:
`<all>` should still be fixed to look like `POST` etc, i.e. gray background, fat, and black. Besides today, I would suggest to use the mark `{all}` to signal the `set`-character.

SmilyOrg:
Linked `<all>` to [#http-requests](https://opensource.zalando.com/restful-api-guidelines/#http-requests) now, which also makes styling more consistent.
image

> Besides today, I would suggest to use the mark `{all}` to signal the `set`-character.

I don't understand this part, @tkrop could you elaborate?

SmilyOrg:
After checking in with @tkrop I'm merging this into the feature branch. Will continue with content changes in follow-up PR.

PR #758, opened 2023-04-18
feat: disable pdf and epub (#753)

### LINKED ISSUE #753: Remove PDF + epub version
We don't have the capacity to maintain the PDF + epub versions of the guidelines.
There are long-outstanding issues we can't solve (e.g. #709), and nobody really seems to care.

Suggestion: just remove the links to these two formats (and the code generating them).

If someone needs a PDF, most browsers offer this functionality from their print dialog.

tkrop:
This pull request removes the build of the `pdf` and the `epub` versions.

fixes #753.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ DIRWORK := $(shell pwd -P)
 .PHONY: check check-rules check-rules-duplicates check-rules-incorrects
 .PHONY: next-rule-id
 
-all: clean html pdf epub rules
+all: clean html rules
 clean:
 	rm -rf $(DIRBUILDS);
 
@@ -62,6 +62,7 @@ html: check assets pull
 	docker run -v $(DIRWORK):$(DIRMOUNTS)/ ${DOCKER} asciidoctor \
 	  -D $(DIRMOUNTS)/$(DIRBUILDS) index.adoc;
 
+# Not used any longer.
 pdf: check pull
 	docker run -v $(DIRWORK):$(DIRMOUNTS)/ ${DOCKER} asciidoctor-pdf -v \
 		-a pdf-fontsdir=$(DIRMOUNTS)/resources/fonts \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -62,13 +62,15 @@ html: check assets pull
...
 	docker run -v $(DIRWORK):$(DIRMOUNTS)/ ${DOCKER} asciidoctor-pdf -v \
 		-a pdf-fontsdir=$(DIRMOUNTS)/resources/fonts \
 		-a pdf-theme=$(DIRMOUNTS)/resources/themes/pdf-theme.yml \
 	  -D $(DIRMOUNTS)/$(DIRBUILDS) index.adoc;
 	mv -f $(DIRBUILDS)/index.pdf $(DIRBUILDS)/zalando-guidelines.pdf;
 
+# Not used any longer.
 epub: check pull
 	docker run -v $(DIRWORK):$(DIRMOUNTS)/ ${DOCKER} asciidoctor-epub3 \
 	  -D $(DIRMOUNTS)/$(DIRBUILDS) index.adoc;
```
ePaul, line 76: Why not just removing these?

Makefile
```diff
@@ -10,7 +10,7 @@ DIRWORK := $(shell pwd -P)
 .PHONY: check check-rules check-rules-duplicates check-rules-incorrects
 .PHONY: next-rule-id
 
-all: clean html pdf epub rules
+all: clean html rules
 clean:
 	rm -rf $(DIRBUILDS);
```
ePaul, line 16: As we are looking at this file ... the next few lines are installing and using a markdown linter. As we are not longer using markdown, they are also not needed anymore, right?

Makefile
tkrop: Currently, this code is not broken and keeping it allows us locally to experiment with building `epub` and `pdf`. If I remove it, we would loose the knowledge and the ability to experiment.

Makefile
tkrop: Nice catch but intentionally left since I considered to configure `markdownlint` to work for the majorly similar `asciidoc` until a linter becomes available. Unfortunately, this is still not the case.

tkrop: Did some investigation. It actually seems as if we at least could apply a couple of markdown linter rules to improve on style, however, we currently start with 415 violations we need to fix in one go.

tkrop: I have created #761 to publish my current investigation result for further testing.

ePaul:
:+1:

tkrop:
👍

PR #760, opened 2023-04-18
feat: improve headers

tkrop:
This pull request improves the definition of default headers.

As fix for the wrong naming of `X-device-Type` and `X-device-OS` I added absolute references to the corrected version. To check effectiveness after merging we need to look at https://apis.zalando.net/apis/af983b04-fc6f-11ec-b939-0242ac120002/ui.

### DIFF
--- a/models/request-headers-1.0.0.yaml
+++ b/models/request-headers-1.0.0.yaml
@@ -1,7 +1,59 @@
+# Standard Headers
+
+If-Match:
+  name: If-Match
+  in: header
+  required: false
+  description: |-
+    The `If-Match` header field is used to declare a list of identifiers that
+    are required to match the current resource version identifier in at least
+    one position as a pre-condition for executing the request on the server
+    side. This behavior is used to validate and reject optimistic updates, by
+    checking if the resource version a consumer has based his changes on is
+    outdated on arrival of the change request to prevent lost updates.
+
+    If the pre-condition fails the server will respond with status code `412`
+    (Precondition Failed). For further details see [RFC 7232 Section
+    3.1](https://tools.ietf.org/html/rfc7232#section-3.1).
+  schema:
+    type: array
+    items:
+      type: string
+  style: simple
+  explode: false
+  example: W/"xy", "5", "5db68c06-1a68-11e9-8341-68f728c1ba70"
+
+If-None-Match:
+  name: If-None-Match
+  in: header
+  required: false
+  description: |-
+    The `If-None-Match header` field is used to declare a list of identifiers
+    that are required to fail matching all the current resource version
+    identifiers as a pre-condition for executing the request on the server
+    side. This is especially used in conjunction with an `*` (asterix) that is
+    matching all possible resource identifiers to ensure the initial creation
+    of a resource. Other use cases are possible but rare.
+
+    If the pre-condition fails the server will respond with status code `412`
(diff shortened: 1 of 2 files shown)


models/request-headers-1.0.0.yaml
```diff
@@ -10,68 +62,78 @@ X-Flow-ID:
...
+    `chat-app`, and `email`. The `X-Frontend-Type` should be passed-through as
+    generic aspect - there are diverse concerns, e.g. pushing mobiles with
+    specific coupons, that make use of it.
   in: header
   schema:
     type: string
     example: mobile-app
 
 X-device-Type:
   name: X-device-Type
```
tkrop, line 118: This should be `X-Device-Type`, but changing may break referencing APIs - we need to investigate, whether this is referenced at all.

models/request-headers-1.0.0.yaml
```diff
@@ -10,68 +62,78 @@ X-Flow-ID:
...
+    for steering customer experience (incl. features and content) depending on
+    device types, e.g. `smartphone`, `tablet`, `desktop`, and `other`. The
+    `X-Device-Type` should be passed-through as generic aspect.
   in: header
   schema:
     type: string
     example: tablet
 
 X-device-OS:
   name: X-device-OS
```
tkrop, line 130: This should be `X-Device-OS`, but changing may break referencing APIs - we need to investigate, whether this is referenced at all.

models/request-headers-1.0.0.yaml
ePaul: I guess you could change it in `name:`, but leave it as the map key.

tkrop: Yes, I agree, should be possible. Shall I?

tfrauenstein: Changing only the name is compatible and should be done.
Can you simply check, whether `https://opensource.zalando.com/restful-api-guidelines/models/request-headers-1.0.0.yaml#/X-device-Type` is used in active API definitions? If it is not used, I would also change it consistently.

tkrop: I think this should be easy possible by doing a grep over all API specifications. May have some false positives but it at least gives a first idea.

tkrop: The simple search confirmed that this item is referenced at least one API.

models/request-headers-1.0.0.yaml
tkrop: The simple search confirmed that this item is referenced at least one API.

tfrauenstein: Than I would propose to change it consistently and create PR for the API currently using it

models/request-headers-1.0.0.yaml
ePaul: Discussed in the meeting:
Let's try to keep it compatible by using
```suggestion
  $ref: X-Device-Type

X-Device-Type:
    name: X-device-Type
```

models/request-headers-1.0.0.yaml
tkrop: Solved this by adding a reference from the old name to the new name within the headers.

tkrop:
👍

SmilyOrg:
:+1:

ePaul:
:+1:

PR #761, opened 2023-04-18
feat: suggest markdown linter config

tkrop:
This pull request suggest to apply a markdown linter configuration that is majorly working also for asciidoc. To put this into actions, the linter step needs to be added to the pipeline again, however, currently the documentation has over `400` linter violations that need to be fixed again.

More rules can be activated, if we can self constrain us to using markdown based code blocks, however, we would need investigate the transformation of some currently used features of asciidoc-based code blocks first.


ePaul:
Hmm, "Pretend to be markdown, and then disable the rules which don't work for AsciiDoc" sounds somewhat strange here.

tkrop:
Strange, but looking on the origin and the compatibility it provides a possible solution.

At the moment I see the problem that the code is degenerating from its original consistent, easy readable style created 6 years ago  with length restricted lines (max 80) with empty separating lines between sections and headings, code blocks etc. into a more and more inconsistent format partially hard to read.

I think we should prepare to add linting to counter-act this.

ePaul:
From the meeting: adjust the build.adoc to show how to use it (and make sure it actually works with the makefile).

tkrop:
👍

BUILD.adoc
```diff
@@ -1,16 +1,40 @@
...
+[source,bash]
+----
+make lint
+----
+
+If linting with the suggested rules proves helpful we will integrate this
+target into the a `pre-commit`-hook to enforce a common style and append it
+to the `all`-target.
+
+**Note:** We have currently not tested the capability of `makrdownlint` (or
```
SmilyOrg, line 27:
```suggestion
**Note:** We have currently not tested the capability of `markdownlint` (or
```

SmilyOrg (commented):
Tiny typo fix, otherwise LGTM

tkrop:
👍

ePaul:
:+1:

PR #769, opened 2023-07-12
feat: improve headers guideline (#769)

tkrop:
This pull request picks up a couple of header related tasks and left overs from previous pull request, e.g. #760, and from the rolling agenda to complete default header definition support via the guidelines:

* Add general introduction on how to reference default header definitions from the guideline.
* Add script to generate guideline definition of headers from provided default models to use models as a single source of truth.
* Add support for `Idempotency-Key`, `Prefer`, `Accept-Encoding`, `Content-Encoding`, `Sunset`, and `Deprecation` headers.
* Improve support and correct definitions of `If-Match`, `If-None-Match`, `ETag` , `Cache-Control`, and `Vary` headers.
* Replace references to RFCs obsoleted by [RFC-9110](https://tools.ietf.org/html/rfc9110) and [RFC-9111](https://tools.ietf.org/html/rfc9111) - except of status code and error codes section.

Generally, this pull request is editorial and does not require Zally to be changed, however, it would be good to add a hint about compression as a Zally rule, since we now can identify endpoints not supporting compression.


chapters/http-headers.adoc
```diff
@@ -2,29 +2,70 @@
...
+Therefore APIs should use the `parameters` and `headers` definition of the API
+endpoint and response to clarify what is supported in a certain context.
+
+[[using--headers]]
+== Using Standard Header definitions
+
+Usually, you can the standard HTTP request and response header definition
+provided by the guideline to simplify API by using well recognized patterns.
+The best practice importing headers providing the highest readability is as
+follows:
```
tkrop, line 27: May be we should introduce this as a new rule to reference headers - if possible - this way.

chapters/http-headers.adoc
```diff
@@ -183,52 +205,29 @@ operation is supposed a to respond with status code {412} - precondition
...
+    If-None-Match:
+      $ref: 'https://opensource.zalando.com/restful-api-guidelines/models/headers-1.0.0.yaml#/If-None-Match'
+----
+
+[source,yaml]
+----
+include::../includes/etag.yaml[]
+include::../includes/if-match.yaml[lines=3..-1]
+include::../includes/if-none-match.yaml[lines=3..-1]
 ----
```
tkrop, line 227: In the long run, I want this represented as collapsible block, that can be unfolded on demand. Since this currently does not work with our current `asciidoc` stile, I think, we need to keep this visible by default.

tfrauenstein:
👍

tfrauenstein (approved):
Thank you! Looks good!

chapters/performance.adoc
```diff
@@ -31,16 +31,46 @@ Each of these items is described in greater detail below.
...
+1. The content is already compressed, or
+2. The server has not enough resources to support compression.
+
+While `gzip` content encoding should be the default, servers must also support
+unencoded content for testing. This is ensured by content negotiation using the
+{Accept-Encoding} request header (see {RFC-9110}#section-12.5.3[RFC 9110
+Section 12.5.3]). Successful compression is signaled via the {Content-Encoding}
+response header (see {RFC-9110}#section-8.4[RFC 9110 Section 8.4]). Clients and
+servers may support other compression algorithms as `compress`, `deflate`, or
+`br`.
```
ePaul, line 47: I'd use "compression formats" here, given that deflate is using the same algorithm as gzip (just with a slightly different wrapper).
```suggestion
servers may support other compression formats as `compress`, `deflate`, or
`br`.
```

models/headers-1.0.0.yaml
```diff
@@ -0,0 +1,475 @@
...
+  in: header
+  required: false
+  description: |-
+    The **Content-Encoding** header lists any encoding (usually a compression
+    algorithm) applied to the message payload - in execution order. This allows
+    the recipient to decode the content in order to obtain the original payload
+    format.
+
+    Supported compression algorithms of the server are [**gzip**][gzip] and
+    **identity**, but other encoding from list below may be support too:
```
ePaul, line 187:
```suggestion
    Always supported compression algorithms of the server are [**gzip**][gzip] and
    **identity**, but other encoding from list below may be support too:
```

models/headers-1.0.0.yaml
```diff
@@ -0,0 +1,475 @@
...
+    * **gzip** - a compression format that uses [Lempel-Ziv coding][gzip]
+      (LZ77) with a 32-bit CRC.
+    * **compress** - a format using the [Lempel-Ziv-Welch][lzw] (LZW)
+      algorithm.
+    * **deflate** - A compression format that uses the [zlib][zlib] structure
+      with the [deflate][deflate] compression algorithm.
+    * **br** - a compression format that uses the [Brotli][brotli] algorithm.
+    * **identity** - the identity function without modification or compression
+      used to indicate unchanged responses. This value is always considered as
+      acceptable, even if omitted.
```
ePaul, line 198:
```suggestion
    * **gzip** - a[compression format with the [gzip file format][gzip]
        (including a 32 bit CRC)  around the [deflate][deflate] compression algorithm.
    * **deflate** - A compression format that uses the [zlib][zlib] structure
      with the [deflate][deflate] compression algorithm.
    * **compress** - a format using the [Lempel-Ziv-Welch][lzw] (LZW)
      algorithm.
    * **br** - a compression format that uses the [Brotli][brotli] algorithm.
    * **identity** - the identity function without modification or compression
      used to indicate unchanged responses. This value is always considered as
      acceptable, even if omitted.
```
Both gzip and "deflate" (aka zlib) are using the same algorithm (deflate), just with different wrappers.

models/headers-1.0.0.yaml
```diff
@@ -0,0 +1,475 @@
...
+    representation of the data. If the original media is encoded in some way,
+    e.g. as a zip file, then this information is not be included in the
+    **Content-Encoding** header.
+
+    See [RFC 9110 Section 8.4][rfc-9110-8.4] as well as [API Guideline Rule
+    #156][api-156] for further details.
+
+    [rfc-9110-8.4]: <https://tools.ietf.org/html/rfc9110#section-8.4>
+    [api-156]: <https://opensource.zalando.com/restful-api-guidelines/#156>
+    [gzip]: <https://en.wikipedia.org/wiki/LZ77_and_LZ78#LZ77>
```
ePaul, line 211:
```suggestion
    [gzip]: <https://en.wikipedia.org/wiki/gzip>
```

models/headers-1.0.0.yaml
```diff
@@ -0,0 +1,475 @@
...
+    * **gzip** - a compression format that uses [Lempel-Ziv coding][gzip]
+      (LZ77) with a 32-bit CRC.
+    * **compress** - a format using the [Lempel-Ziv-Welch][lzw] (LZW)
+      algorithm.
+    * **deflate** - A compression format that uses the [zlib][zlib] structure
+      with the [deflate][deflate] compression algorithm.
+    * **br** - a compression format that uses the [Brotli][brotli] algorithm.
+    * **identity** - the identity function without modification or compression
+      used to indicate unchanged responses. This value is always considered as
+      acceptable, even if omitted.
```
ePaul, line 25:
```suggestion
    * **gzip** - a[compression format with the [gzip file format][gzip]
        (including a 32 bit CRC)  around the [deflate][deflate] compression algorithm.
    * **deflate** - A compression format that uses the [zlib][zlib] structure
      with the [deflate][deflate] compression algorithm.
    * **compress** - a format using the [Lempel-Ziv-Welch][lzw] (LZW)
      algorithm.
    * **br** - a compression format that uses the [Brotli][brotli] algorithm.
    * **identity** - the identity function without modification or compression
      used to indicate unchanged responses. This value is always considered as
      acceptable, even if omitted.
```
Both gzip and "deflate" (aka zlib) are using the same algorithm (deflate), just with different wrappers.

models/headers-1.0.0.yaml
```diff
@@ -0,0 +1,475 @@
...
+    **Note:** A client is allowed to explicitly forbid the identity value by
+    setting **identity;q=0** or **\*;q=0**, however, the server in return may
+    respond with a **406** (Not Acceptable) error.
+
+    See [RFC 9110 Section 12.5.3][rfc-9110-12.5.3] as well as [API Guideline
+    Rule #156][api-156] for further details.
+
+    [rfc-9110-12.5.3]: <https://tools.ietf.org/html/rfc9110#section-12.5.3>
+    [api-156]: <https://opensource.zalando.com/restful-api-guidelines/#156>
+    [gzip]: <https://en.wikipedia.org/wiki/LZ77_and_LZ78#LZ77>
```
ePaul, line 41:
```suggestion
    [gzip]: <https://en.wikipedia.org/wiki/gzip>
```

scripts/generate-includes.sh
```diff
@@ -0,0 +1,44 @@
...
+    }
+    ($0 ~ "^#.*request headers$") {
+        mode = "parameters"; next
+    }
+    ($0 ~ "^#.*common headers$") {
+        mode = "parameters"; next
+    }
+    ($0 ~ "^[A-Z][A-Za-z-]*:$") {
+        print $0 mode
+    }'))
```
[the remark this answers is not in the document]
tkrop: Thanks for the hint. I love linters, but have not used shellcheck before. Will do from now on. Howver: shellcheck is not matching the right code format. This is an inline `awk`-script, where this is the native usage pattern. So I'm not sure that this error applies here.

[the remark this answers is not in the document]
tkrop: Seems to be a correct note, but following the default pattern does not work with `readonly`. Will see whether there is a proper option to change this.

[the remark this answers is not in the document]
tkrop: I cannot come up with an alternate pattern, however, as the advise states, I have already taken care in the `awk` part that the input pattern is as required with separators/white spaces where they ought to be.

tfrauenstein:
👍

tkrop:
👍

Issue #763, opened 2023-05-30
Improvements for Problem guideline (176)


ePaul:
We discovered that our current [guideline about problem/json](https://opensource.zalando.com/restful-api-guidelines/#176) is interpreted somewhat differently (both between guild members, and by other developers in the company).

Some things to clarify (most related to each other):

- [ ] 1. Is the problem object meant just for the client developers (for debugging purposes), or is it also intended to be used by client applications?
     - [ ]  1a. If the latter, only for display to a human (e.g. in an UI), or also for automatic processing decisions?
- [ ] 2. What is the intended meaning of the `type` (+ `title`) member, compared to the status code?
    - [ ] 2a. Should the problem types be defined in your API? (How?) Should they be standardized across APIs?
    - [ ] 2b. Can clients rely on the type attribute?
    - [ ] 2c. Is this the same as an "error code"?
- [ ] 3. What is the meaning of the `instance` field?
   - [ ] 3a. Does pointing to a specific occurrence of an error is meant to point to a code location, or to a set of input data or time when this error happened (e.g. via flow-ID)?
   - [ ] 3b. When is this needed/useful in addition to the type?
- [ ] 4. For cases where a server application doesn't have more detailed error information, should fields (e.g. `type` + `title`) be filled with default values replicating the status code information, or just be left empty/missing?
- [ ] 5. What are compatibility concerns for problem objects? Is it an incompatible change if in some specific situation now a different `type` value is returned?
- [ ] 6. How can/should the problem object be extended?
    - [ ] 6a. Use the `type` as a discriminator for subtyping?
    - [ ] 6b. How can extended problems and generic ones coexist?
- [ ] 7. Client robustness: Can clients rely on anything in the error object? If not, can they still use something?

(Also, if we are looking at this guideline: Fix the typo `instant` → `instance` in the first paragraph.)

SmilyOrg:
I tried to dump a few thoughts on this without going through the whole discussion, so feel free to just skip over this and treat it as a brain dump. I'll try to go through the discussion and reconcile it with my thoughts in a bit :)

> 1. Is the problem object meant just for the client developers (for debugging purposes), or is it also intended to be used by client applications?

Currently the spec heavily leans towards it being for developers. I'd tend to mostly agree there. The errors the API returns likely have to be interpreted and provided with the right messaging in the right context for them to be useful when embedded into a frontend.

[...]

PR #773, opened 2023-08-14
refactor: RFC 9457 obsoletes 7807, refer to it

scop:
No substantial changes that would run at odds with the elaboration in the guidelines that I can see.

### DIFF
--- a/chapters/http-status-codes-and-errors.adoc
+++ b/chapters/http-status-codes-and-errors.adoc
@@ -357,7 +357,7 @@ number of requests made within a given window for each named entity.
 [#176]
 == {MUST} support problem JSON
 
-{RFC-7807}[RFC 7807] defines a Problem JSON object using the media type
+{RFC-9457}[RFC 9457] defines a Problem JSON object using the media type
 `application/problem+json` to provide an extensible human and machine readable
 failure information beyond the HTTP response status code to transports the
 failure kind (`type` / `title`) and the failure cause and location (`instant` /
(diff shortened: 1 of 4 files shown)


tfrauenstein:
👍

ePaul:
Thanks for the pointer, I missed this new RFC.

There are these changes from RFC-7807 to RFC-9457 (mentioned [in the Appendix D](https://datatracker.ietf.org/doc/html/rfc9457#changes)) :

* [Section 4.2](https://datatracker.ietf.org/doc/html/rfc9457#registry) introduces a registry of common problem type URIs

   The [registry](https://www.iana.org/assignments/http-problem-types/http-problem-types.xhtml) so far doesn't have any types which seem to be generally useful (except the default `about:blank`).

* [Section 3](https://datatracker.ietf.org/doc/html/rfc9457#problem-json) clarifies how multiple problems should be treated

  * For multiple problems of the same type, use an extension with more problem objects in it (only the relevant fields).
  * For multiple problems of different types, just keep the most important/urgent one.

* [Section 3.1.1](https://datatracker.ietf.org/doc/html/rfc9457#type) provides guidance for using type URIs that cannot be dereferenced

  It also contains this recommendation:
  >  As a result, it is **RECOMMENDED** that absolute URIs be used in "type" when possible and that when relative URIs are used, they include the full path (e.g., "/types/123").

 This goes contrary to our existing recommendation to not use absolute problem types (though it fits the "use full path" one).
I suggest we discuss in our next API guild meeting whether it makes sense to adjust our guideline here. (Related: #763)

tkrop:
@epaul I don't think that the difference is so big compared to the short sentences in https://www.rfc-editor.org/rfc/rfc7807#section-3.1. We deliberately decided to not follow the convention of absolute URIs to relief our API designers from providing a registry of error types.

tkrop:
👍

PR #776, opened 2023-09-15
fix: header examples (#769)

tkrop:
Fixes examples in header templates that must either escape commas or provide a list of examples using array brackets. Non-observing this rule leads to a null pointer exceptions in the swagger library and an error in the swagger editor.

fixes #769.


tkrop:
👍

tfrauenstein:
👍

PR #785, opened 2023-10-25
feat: add new rule to prefer single resources (#764)

### LINKED ISSUE #764: Add guideline for common request/response schemas
In the guild meeting we have discussed that we want to provide additional guidance on how to utilize a common schema definition for requests and responses. JSON Schema draft is providing support for `readOnly` and `writeOnly` flags to differentiate between properties only present in a request and properties only present in a response. The guideline has the following goals:

1. It encourages the use of a common schema for requests and responses instead of writing to separate models where possible.
2. It defines the default unspecified server behavior for `readOnly` that a server should ignore the field (opposing to rule [#109](https://opensource.zalando.com/restful-api-guidelines/#109) that suggest to reject fields unknown in the request object). APIs rejecting `readOnly` fields consequently need to specify their non-default-like behavior explicitly.

tkrop:
This pull request adds a new single resource preference rule according to proposal defined in #764.

fixes #764.

### DIFF
--- a/chapters/json-guidelines.adoc
+++ b/chapters/json-guidelines.adoc
@@ -11,15 +11,14 @@ an idiomatic form across teams and services.
 [#167]
 == {MUST} use JSON as payload data interchange format
 
-Use JSON ({RFC-7159}[RFC 7159]) to represent structured (resource) data
-passed with HTTP requests and responses as body payload.
-The JSON payload must use a JSON object as top-level data
-structure (if possible) to allow for future extension. This also applies to
-collection resources, where you ad-hoc would use an array -- see also
-<<110>>.
+Use JSON ({RFC-7159}[RFC 7159]) to represent structured (resource) data passed
+with HTTP requests and responses as body payload. The JSON payload must use a
+JSON object as top-level data structure (if possible) to allow for future
+extension. This also applies to collection resources, where you ad-hoc would
+use an array -- see also <<110>>.
 
-Additionally, the JSON payload must comply to the more restrictive Internet JSON ({RFC-7493}[RFC 7493]),
-particularly
+Additionally, the JSON payload must comply to the more restrictive Internet
+JSON ({RFC-7493}[RFC 7493]), particularly
 
 * {RFC-7493}#section-2.1[Section 2.1] on encoding of characters, and
 * {RFC-7493}#section-2.3[Section 2.3] on object constraints.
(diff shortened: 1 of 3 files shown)


chapters/json-guidelines.adoc
```diff
@@ -32,6 +31,27 @@ As a consequence, a JSON payload must
 * contain only {RFC-7493}#section-2.3[unique member names] (no duplicate
   names).
 
+
+[#252]
+== {SHOULD} design single resource for reading and writing
```
SmilyOrg, line 36: Resource might be ambiguous here, how about [schema](https://swagger.io/specification/#schema-object)? That seems to be what OpenAPI calls it and also where you'd usually define these (under `components/schemas`)
```suggestion
== {SHOULD} design single schema for reading and writing
```

chapters/json-guidelines.adoc
```diff
@@ -32,6 +31,27 @@ As a consequence, a JSON payload must
...
+
+To simplify API specs and API maintenance, as well as the cognitive load of
+developers, endpoints should  utilize a common model for reading and writing
+the same resource type. The usualy differences between read and write models
+are best addressed by
```
SmilyOrg, line 41:
```suggestion
To simplify API specs and API maintenance, as well as the cognitive load of
developers, endpoints should utilize a common model for reading and writing
the same resource type. The differences between read and write models
are best addressed by
```

chapters/json-guidelines.adoc
```diff
@@ -32,6 +31,27 @@ As a consequence, a JSON payload must
...
+* marking properties `writeOnly` that are only available in a request, e.g.
+  passwords, and
+* marking properties `readOnly` that are only available in a response, e.g.
+  resource identifiers.
+
+**Note:** A resource with property marked `readOnly` in the API according to
+the {json-schema-validation}#rfc.section.9.4[JSON schema validation] may be
+ignored or rejected by a server. Both is inline with the idea behind <<#109,
+the compatibility guidance>>, however, we suggest to document rejection to
+prevent client side surprises.
```
SmilyOrg, line 52:
```suggestion
**Note:** A resource with property marked `readOnly` in the API according to
the {json-schema-validation}#rfc.section.9.4[JSON schema validation] may be
ignored or rejected by a server. Both are in line with the idea behind
<<#109, the compatibility guidance>>, however, we suggest to document
rejection to prevent client side surprises.
```

SmilyOrg (commented):
A few comments :)

chapters/json-guidelines.adoc
tkrop: How about `resource schema`?

SmilyOrg: Yeah that also works!

SmilyOrg:
:+1:

tkrop:
👍

ePaul:
:+1:

chapters/json-guidelines.adoc
```diff
@@ -32,6 +31,27 @@ As a consequence, a JSON payload must
...
+addressed by
+
+* marking properties `writeOnly` that are only available in a request, e.g.
+  passwords, and
+* marking properties `readOnly` that are only available in a response, e.g.
+  resource identifiers.
+
+**Note:** A resource with property marked `readOnly` in the API according to
+the {json-schema-validation}#rfc.section.9.4[JSON schema validation] may be
+ignored or rejected by a server. Both are in line with the idea behind <<#109,
```
ktsypkina, line 50:
```suggestion
ignored or rejected by a server. Both are in line with the idea behind 
<<#109,the compatibility guidance>>, however, we suggest to document when
rejecting to prevent client side surprises.
```

PR #111, opened 2016-07-21
removed duplicate deprecation paragraph

codekultur:
Not sure how this happened: By accident the merge for the API deprecation as a separate chapter didn't have the commit in it which removed the API deprecation paragraph from the Compatibility chapter... This PR should fix the duplicate deprecation paragraph...

PR #793, opened 2024-01-29
fix discrepancy between 109 and 111

ktsypkina:
change "may" to stronger "should" for additional (unknown) properties described in #111 to be compliant with requirements from server described in #109

### DIFF
--- a/chapters/compatibility.adoc
+++ b/chapters/compatibility.adoc
@@ -149,11 +149,12 @@ is not required to make an object definition extensible:
   ignore fields sent by the server they cannot process. This allows API
   servers to evolve their data formats.
 * For API servers receiving unexpected data, the situation is slightly
-  different. Instead of ignoring fields, servers _may_ reject requests whose
-  entities contain undefined fields in order to signal to clients that those
-  fields would not be stored on behalf of the client. API designers must
-  document clearly how unexpected fields are handled for {PUT}, {POST}, and
-  {PATCH} requests.
+  different. According to <<109>> instead of ignoring fields,
+  servers _should_ reject requests whose entities contain undefined fields
+  in order to signal to clients that those fields would not be stored on behalf
+  of the client.
+  API designers must document clearly how unexpected fields are handled
+  for {PUT}, {POST}, and {PATCH} requests.
 
 API formats must not declare `additionalProperties` to be false, as this
 prevents objects being extended in the future.


chapters/compatibility.adoc
```diff
@@ -149,11 +149,12 @@ is not required to make an object definition extensible:
   ignore fields sent by the server they cannot process. This allows API
   servers to evolve their data formats.
 * For API servers receiving unexpected data, the situation is slightly
-  different. Instead of ignoring fields, servers _may_ reject requests whose
-  entities contain undefined fields in order to signal to clients that those
-  fields would not be stored on behalf of the client. API designers must
-  document clearly how unexpected fields are handled for {PUT}, {POST}, and
-  {PATCH} requests.
+  different. According to <<109>> instead of ignoring fields,
```
ePaul, line 152:
```suggestion
  different. According to <<109>>, instead of ignoring fields,
```

tfrauenstein:
👍

chapters/compatibility.adoc
```diff
@@ -149,11 +149,12 @@ is not required to make an object definition extensible:
...
-  entities contain undefined fields in order to signal to clients that those
-  fields would not be stored on behalf of the client. API designers must
-  document clearly how unexpected fields are handled for {PUT}, {POST}, and
-  {PATCH} requests.
+  different. According to <<109>>, instead of ignoring fields,
+  servers _should_ reject requests whose entities contain undefined fields
+  in order to signal to clients that those fields would not be stored on behalf
+  of the client.
+  API designers must document clearly how unexpected fields are handled
+  for {PUT}, {POST}, and {PATCH} requests.
```
tkrop, line 157: As written, this sentence is impractical and counterproductive, since it does not restrict when to document. I suggest to change it as follows 

```suggestion
  API designers must document clearly how unexpected fields are handled
  for {PUT}, {POST}, and {PATCH} requests, if they deviate from the default
  rejection strategy.
```

to be in line with the recommendation in rule `109`.

tkrop (commented):
@ktsypkina I think we should also improve on rule `109`. At the moment it is missing a similar sentence as in `111` about that APIs must document, if they deviates from the default rejection strategy. Can you add a sentence about this after the strict recommendation too?

chapters/compatibility.adoc
ktsypkina: good point, I'll update the sentence

tkrop:
👍

ePaul:
:+1:

ktsypkina:
👍

tkrop:
:+1:

PR #794, opened 2024-02-07
Update security section about uid scope

oporkka:
Add more visible note that `uid` scope is also a valid scope for publicly available data, as this is often misunderstood, even if there is text about this below in the same section.

### DIFF
--- a/chapters/security.adoc
+++ b/chapters/security.adoc
@@ -55,11 +55,11 @@ because it exposes authentication server address details and may make use of red
 
 APIs must define permissions to protect their resources. Thus, at least one
 permission must be assigned to each API endpoint. You should use the `uid`
-pseudo-scope to allow read-only (?) access to public and employee-only data
-(classified as `green` and `yellow` respectively). For sensitive data (`orange`
-or `red`), the `uid` scope based authorization must be either accompanied by
-individual object level authorization or use role based permissions through
-specific scopes.
+pseudo-scope to allow access to public and employee-only data (classified as
+`green` and `yellow` respectively). For sensitive data (`orange` or `red`),
+the `uid` scope based authorization must be either accompanied by individual
+object level authorization or use role based permissions through specific
+scopes.
 
 The naming schema for permissions corresponds to the naming schema for <<224,
 hostnames>> and <<213, event type names>>. Please refer to <<225>> for
--- a/chapters/security.adoc
+++ b/chapters/security.adoc
@@ -35,9 +35,7 @@ also <<105>>):
 [source,yaml]
 ----
 security:
-
-- BearerAuth: [ api-repository.read ]
-
+  - BearerAuth: [ api-repository.read ]
 ----
 
 
--- a/chapters/security.adoc
+++ b/chapters/security.adoc
@@ -54,8 +54,12 @@ because it exposes authentication server address details and may make use of red
 == {MUST} define and assign permissions (scopes)
 
 APIs must define permissions to protect their resources. Thus, at least one
-permission must be assigned to each API endpoint. The `uid` default scope can be used
-for publicly available data, but should be declared explicitly (see details below).
+permission must be assigned to each API endpoint. You should use the `uid`
+pseudo-scope to allow read-only (?) access to public and employee-only data
+(classified as `green` and `yellow` respectively). For sensitive data (`orange`
+or `red`), the `uid` scope based authorization must be either accompanied by
+individual object level authorization or use role based permissions through
+specific scopes.
 
 The naming schema for permissions corresponds to the naming schema for <<224,
 hostnames>> and <<213, event type names>>. Please refer to <<225>> for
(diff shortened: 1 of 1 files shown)


chapters/security.adoc
```diff
@@ -35,7 +35,9 @@ also <<105>>):
 [source,yaml]
 ----
 security:
-  - BearerAuth: [ api-repository.read ]
+
+- BearerAuth: [ api-repository.read ]
```
oporkka, line 39: This is from running 
```
make format
```

tkrop: `make format` is experimental and the before format is more aligned with our expectations - I think. @SmilyOrg  what do you think.

oporkka: I restored the original format

tkrop: Thanks :smiley:

chapters/security.adoc
```diff
@@ -54,7 +54,8 @@ because it exposes authentication server address details and may make use of red
 == {MUST} define and assign permissions (scopes)
 
 APIs must define permissions to protect their resources. Thus, at least one
-permission must be assigned to each API endpoint.
+permission must be assigned to each API endpoint. The `uid` default scope can be used
+for publicly available data, but should be declared explicitly (see details below).
```
tkrop, line 58: @oporkka I'm not sure about the connection of `uid` scope and `public available data` here. I think the context is much more complex and the simple formulation can create an incorrect understanding. May be the following suggestion is can improve on your extension:

```suggestion
permission must be assigned to each API endpoint. Often, the pseudo-scope `uid`
can be used to expose date allowed to be shared within a realm, e.g. `users`
(data classified as `gree`), or `employees` (data classified as `yellow`). For
sensitive data (`orange` or `red`), the `uid` scope based authorization must be
either accompanied by individual object level authorization or use role based
permissions through specific scopes.
```

SmilyOrg: @tkrop I think that's a good addition. I would even more directly say that green and yellow data -> uid.
```suggestion
permission must be assigned to each API endpoint. You should use the `uid`
pseudo-scope to allow read-only (?) access to public and employee-only data
(classified as `green` and `yellow` respectively). For sensitive data (`orange`
or `red`), the `uid` scope based authorization must be either accompanied by
individual object level authorization or use role based through specific scopes.
```

tkrop: 👍  - re-added the missing `permissions` to `role based permissions through specific scopes`.

chapters/security.adoc
```diff
@@ -54,7 +54,12 @@ because it exposes authentication server address details and may make use of red
 == {MUST} define and assign permissions (scopes)
 
 APIs must define permissions to protect their resources. Thus, at least one
-permission must be assigned to each API endpoint.
+permission must be assigned to each API endpoint. You should use the `uid`
+pseudo-scope to allow read-only (?) access to public and employee-only data
```
oporkka, line 58: Looks good, but what is the. `(?)` referring to here?

chapters/security.adoc
```diff
@@ -54,7 +54,12 @@ because it exposes authentication server address details and may make use of red
 == {MUST} define and assign permissions (scopes)
 
 APIs must define permissions to protect their resources. Thus, at least one
-permission must be assigned to each API endpoint.
+permission must be assigned to each API endpoint. You should use the `uid`
```
oporkka, line 57: Would typically be good to add here, or does "should" kind of include it as terminology?
```suggestion
permission must be assigned to each API endpoint. You should typically use the `uid`
```

tkrop: There is no alternative to `uid` for the use case, so I hope without `typically` this is much clearer.

chapters/security.adoc
tkrop: Ja, that's a good question. I missed to remove this before merging. @SmilyOrg any reason for keeping this in?

tfrauenstein:
Thank you for the PR -- I have some more change proposals and therefore created a new PR that anticipates this PR and the comments -- see https://github.com/zalando/restful-api-guidelines/pull/798 
Please, let us close this PR and check the new PR -- thank you!

chapters/security.adoc
oporkka: There is no alternative to `uid` as a default scope. But there is alternative to use additional scopes for reason XYZ. "SHOULD" would possibly include that you "MAY" use other scopes. I do not know what is the clearest way to represent that there might be exceptions.

tfrauenstein: It is a must rule -- the PR I proposed makes it a bit clearer https://github.com/zalando/restful-api-guidelines/pull/798/files#diff-5dc5d5e7874b03c87880339a829d18faa591365191f05caa2e9911b3419282e5R66-R68

ePaul:
:+1:

tkrop:
👍

tfrauenstein:
👍

PR #802, opened 2024-03-21
fix: asciidoc build to run as user (#802)

tkrop:
This pull request fixes the issue that asciidoctor output (i.e. the produced index.html) is owned by root through running the docker container using root permissions (which some browsers running in a snap refuse to open).

The pull request changes this to run asciidoctor with user permissions.

In addition, it allows running with podman instead of docker if no docker is installed.


ePaul:
Works on my Ubuntu, thanks.

ePaul:
:+1:

tkrop:
👍

PR #804, opened 2024-04-11
Remove sort example from simple query language (236)

ePaul:
The `sort=id:desc` example doesn't fit the "equality comparison for a logical property".

It also leads to confusion on how to indicate how the result should be sorted (rule 137 says `sort=-id`).

This replaces it with another "logical property", the age (which is often calculated from the birthday and current date).

_(Triggered by an [internal chat message](https://chat.google.com/room/AAAAAqZDJ28/zr5609abqFs/zr5609abqFs?cls=10).)_

### DIFF
--- a/chapters/http-requests.adoc
+++ b/chapters/http-requests.adoc
@@ -541,7 +541,7 @@ The following examples should serve as ideas for simple query language:
 
 * `name=Zalando`, `creation_year=2023`, `updated_by=user1` (query elements
   based on property equality)
-* `created_at=2023-09-18T12:12:00.000Z`, `sort=id:desc` (query elements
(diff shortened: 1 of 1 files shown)


rishiraj88:
For sort order, 'asc' and 'dec' may make good sense. I am not sure whether we are curious about 'desc' to mean "decreasing". Is it a case here?

ePaul:
@rishiraj88 
> For sort order, 'asc' and 'dec' may make good sense. I am not sure whether we are curious about 'desc' to mean "decreasing". Is it a case here?

`desc` (for "descending") might make sense, but it doesn't fit into the query language rule (as it's not a filter), and it also contradicts the earlier rule 137 about the `+` or `-` prefix (and we want to be consistent).

tfrauenstein:
👍

PR #815, opened 2024-08-24
change the image of the assets

AliMohammadnezhad:


ePaul:
Hi,

thank you for your contribution.
We don't think that this seemingly AI-generated image works well as an icon – it is a bit too information-dense.
(Also, we like our current logo.)

Therefore we won't merge this change.

PR #830, opened 2025-04-15
Update RFC links in rule 243

### LINKED ISSUE #828: Is this rendering intentional? Link looks broken
Is this behavior intended, or should the whole thing render as one continuous link?

To me, it looks like the fragment #section-6[...] was meant to be part of the link, but the browser doesn't recognize it that way — probably due to the inline // comment breaking the parser?

[link](https://opensource.zalando.com/restful-api-guidelines/#243)
Image

ePaul:
RFCs 7231 (HTTP 1.1 Semantics) and 7235 (HTTP 1.1: Authentication) were both replaced by RFC 9119 (HTTP Semantics).

The draft for 451 is now RFC 7725, but I don't see a point linking to it, as we are not using it.

Fixes #828.

### DIFF
--- a/chapters/http-status-codes-and-errors.adoc
+++ b/chapters/http-status-codes-and-errors.adoc
@@ -9,15 +9,12 @@ You must only use official HTTP status codes consistently with their intended
 semantics. Official HTTP status codes are defined via RFC standards and
 registered in the
 https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml[IANA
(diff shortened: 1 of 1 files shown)


tfrauenstein:
👍

ePaul:
So the issue was that the build didn't run as the PR was from an outside contributor, and nobody approved the check run.
And now I can't retrigger it because it's too old.

I created a new PR #833 from it, hopefully that works better.

PR #832, opened 2025-04-15
Update RFC links in request and response header models

ePaul:
When doing the change for #830, I noticed that the model YAML files also still pointed to the old RFCs in the descriptions.
This changes the links (and a bit text around it) to point to the current generation of HTTP RFCs.


tfrauenstein:
👍

PR #838, opened 2025-04-30
Fix missing link to cache-support-patterns

jiheon2234:
Previously, there was only `(see)` without a reference.
Added anchor to cache-support-patterns section.


ePaul:
Thanks for the contribution, and sorry for taking so long.

Issue #840, opened 2025-05-12
Clarify interaction of get-with-body with pagination


ePaul:
The "Note" at the end of the [Get with body](https://opensource.zalando.com/restful-api-guidelines/#get-with-body) section of [MUST use HTTP methods correctly [148]](https://opensource.zalando.com/restful-api-guidelines/#148) says:

> **Hint:** As [GET with body](https://opensource.zalando.com/restful-api-guidelines/#get-with-body) is used to transport extensive query parameters, the [`cursor`](https://opensource.zalando.com/restful-api-guidelines/#cursor) cannot any longer be used to encode the query filters in case of [cursor-based pagination](https://opensource.zalando.com/restful-api-guidelines/#160). As a consequence, it is best practice to transport the query filters in the body payload, while using [pagination links](https://opensource.zalando.com/restful-api-guidelines/#161) containing the [cursor](https://opensource.zalando.com/restful-api-guidelines/#cursor) that is only encoding the page position and direction. To protect the pagination sequence the [cursor](https://opensource.zalando.com/restful-api-guidelines/#cursor) may contain a hash over all applied query filters (See also [SHOULD use pagination links](https://opensource.zalando.com/restful-api-guidelines/#161)).

On the other hand, the [SHOULD use pagination links](https://opensource.zalando.com/restful-api-guidelines/#161) rule has no mention of get-with-body, and when missing this hint, it's not immediately clear how this should work together.

One potential interpretation might be that the "get with body" POST creates a filtered view on the server, and the returned pointer URLs might directly work as GET (pointing to this filtered view). This is not a best practice (as it stores client session state on the server), instead it should be made clear that those URLs are meant to be used together with the body containing the filter parameters (the same one as in the original query, and the one in the `query` property of the response page).

I suggest adjusting rules 161 and 284 for making this clearer.

tkrop:
@ktsypkina as far as I can see from the related pull request, this is already sufficiently addressed.

PR #841, opened 2025-05-13
Clarify interaction between get-with-body and pagination

ePaul:
Main points:

* following the links will need another POST with the same filter parameters in the body
* paginated responses to get-with-body should have a `query` object
* the `query` object in the response can be directly used as a request body

In addition, a few editorial improvements around these areas.

Related issue: #840


tfrauenstein:
👍

PR #848, opened 2025-09-17
fix: deprecation and sunset header (#848)

### LINKED ISSUE #844: Deprecation header syntax
Current guidance ([Rule 189](https://opensource.zalando.com/restful-api-guidelines/#189)) says to use a date-time or set it to true. That seems to have been the case in a [previous draft](https://datatracker.ietf.org/doc/html/draft-ietf-httpapi-deprecation-header-02#section-2.1) but  subsequent versions, and the resulting [RFC 9745](https://datatracker.ietf.org/doc/html/rfc9745#section-2.1) expect it to be structured (date) header field with a value like `@1688169599`.

tkrop:
This pull request fixes the errors made when defining the deprecation and sunset headers and advances the documentation to refer to the latest RFCs.

fixes #844.


tkrop:
👍

tfrauenstein (commented):
Good catch! -- two minor change proposals.

tfrauenstein:
👍

PR #781, opened 2023-10-17
Status codes publish to main

SmilyOrg:
Publishes the `status-codes` branch used by #756 and #762 to main.

PR #762, opened 2023-05-04
Apply working draft content to status codes

SmilyOrg:
* Revise the body text of existing codes
* Add and clarify missing and new codes
* Add no-redirection as rule 250
* Gray-out "do-not-use" to make it even more obvious

PR #759, opened 2023-04-18
feat: must sanitize json payload (#734)

tkrop:
Creates a new rule to require `JSON`-payload to be sanitized.

fixes #734.

PR #852, opened 2025-10-07
fix mixup of rules 250 and 251.

ePaul:
Rule 251 was originally added as 250 in #762 to a feature branch, but then #759 adding 250 was merged first to main before #781 merged that feature branch to master (and renamed it).

This fixes a reference in rule 107 and in the changelog.


tfrauenstein:
👍

ePaul:
:+1:

PR #865, opened 2026-05-05
feat: best practice implementing intervals (#805)

### LINKED ISSUE #805: How to implement intervals?
[Rule 127](https://opensource.zalando.com/restful-api-guidelines/#127) describes intervals (i.e. a range of date-times, with some special open-ended ones), and suggests to use these with `..._between` query parameters instead of a pair of `..._before` and `..._after` query parameters.

In the API guild meeting of 2024-04-16, we found that this is not quite easy to implement, e.g. the standard java.time API doesn't even has a type representing it and Joda time's Interval doesn't support open-ended intervals. Guava's Range\<Instant\> could work, but needs a manual parser. (Same issue likely exists for other languages.)

We should get a section in the best practices chapter giving some ideas of how to implement this (@tkrop mentioned [an example](https://github.bus.zalan.do/builder-knowledge/api-repository/blob/baab70488602ad9749f99190af7cb8a3fee164d1/src/main/java/org/zalando/infrastructure/api/repository/domain/service/ApiRevisionHelper.kt#L193) (internal link)), or alternatively rethink this recommendation.

tkrop:
This pull request provides a best practice section on how to implement open ended intervals in Java(Kotlin/Scala), Go, and Python as an example for other languages.

fixes #805.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ check-rules-duplicates:
 	fi;
 
 check-rules-incorrects:
-	@INCORRECT="$$(grep -rnE "(\[[0-9]{1,4}\]|\[ +#?[0-9]+ +\])" $(DIRCONTENTS))"; \
+	@INCORRECT="$$(grep -rnE "(^\[[0-9]{1,4}\]|\[ +#?[0-9]+ +\])" $(DIRCONTENTS))"; \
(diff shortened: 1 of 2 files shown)


tfrauenstein (approved):
LGTM - Thank you!

tkrop:
👍

PR #866, opened 2026-05-15
Add note on endpoint URL consistency for pagination

fghibellini:
# Description

We have services that render pagination links with statically preconfigured base URLs.
The same service will be exposed under multiple hostnames for compatibility reasons.

Users of such endpoints will then put effort into configuring access-list rules in client systems (e.g. ZMON) based on the behavior of the requests. During development, a single page of results is often returned.

Unbeknownst to them, the access-list rules will not apply to the pagination links since they are not guaranteed to target the same base URLs they used to fetch the first page. As soon as the number of entities grows large enough not to fit in the first page, the integration breaks and the developer has to put extra effort into investigating the problem and configuring more access-list rules.

This PR amends the [suggestion for pagination links](https://opensource.zalando.com/restful-api-guidelines/#161) with a paragraph encouraging the use of consistent base URLs.

### DIFF
--- a/chapters/pagination.adoc
+++ b/chapters/pagination.adoc
@@ -173,6 +173,12 @@ For {GET-with-body} operations, the `query` object can to be used as a request
 body with either of these links. (It should be equivalent to just resend
 the original body.)
 
+An endpoint might be exposed under different URLs (the protocols, hostnames and ports may vary).
+The protocol, hostame and port of the pagination links should match the ones used in the request
+this response is for. This avoids scenarios where one would configure access-list rules that
+allow the first request to succeed only for the integration to fail the moment the number of entities
+overflows to the second page.
+
 See also <<248>> for details on the pagination fields and page result object.
 
 
--- a/chapters/pagination.adoc
+++ b/chapters/pagination.adoc
@@ -174,8 +174,7 @@ body with either of these links. (It should be equivalent to just resend
 the original body.)
 
 An endpoint might be exposed under different URLs (the protocols, hostnames and ports may vary).
-The protocol, hostame and port of the pagination links should match the ones used in the request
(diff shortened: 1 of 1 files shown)


chapters/pagination.adoc
```diff
@@ -173,6 +173,12 @@ For {GET-with-body} operations, the `query` object can to be used as a request
 body with either of these links. (It should be equivalent to just resend
 the original body.)
 
+An endpoint might be exposed under different URLs (the protocols, hostnames and ports may vary).
+The protocol, hostame and port of the pagination links should match the ones used in the request
+this response is for. This avoids scenarios where one would configure access-list rules that
```
fghibellini:
```suggestion
The protocol, hostame and port of the pagination links should match the ones used in the first request. This avoids scenarios where one would configure access-list rules that
```

fghibellini:
Alternatively/Additionally the guidelines may suggest an implementation that treats one of the origins as the "canonical one" but then any request to any other origin (including a request for the first page) should result in a redirect. This would ensure that the target of the redirect is included in the access list early on (without it even the first page fetch is impossible).

ePaul:
From the meeting:

* This doesn't apply just to pagination links, but all kinds of links.
* Thus it might best fit into rule [217](https://opensource.zalando.com/restful-api-guidelines/#217):
    * Whenever you create a link which is intended to your own service, it should use the same scheme + host + port as the original request URL.
* Still, we should link from the pagination rule to that one.

ePaul:
@fghibellini Hi, sorry for taking that long to review your PR. We've discussed this in the meeting today, and got to the result that this should be tackled more generally, for all kinds of links. I've opened #868 for this.

PR #868, opened 2026-06-09
Require host name consistency

ePaul:
Replaces #866.

As discussed in today's meeting, the valid concern addressed in #866 applies to all kinds of links, not just the pagination links. As such, I've added it to rule 217, with a link at rule 161.

### DIFF
--- a/chapters/hyper-media.adoc
+++ b/chapters/hyper-media.adoc
@@ -147,7 +147,7 @@ See <<164>> and <<161>> for more information and examples.
 
 
 [#217]
-== {MUST} use full, absolute URI for resource identification
+== {MUST} use full, absolute URI for resource identification (with same host, if applicable)
 
 Links to other resource must always use full, absolute URI.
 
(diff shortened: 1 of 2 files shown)


chapters/hyper-media.adoc
```diff
@@ -158,6 +158,11 @@ when using features like embedding subresources. The primary advantage of
 non-absolute URI is reduction of the payload size, which is better achievable
 by following the recommendation to use <<156,gzip compression>>
 
+If the linked resource belongs to the same API, the link should use the same scheme + host + port combination as the original request URI. (This only comes into play when the API is available at e.g. multiple host names.)
```
tfrauenstein:
```suggestion
If the linked resource belongs to the same [API](https://opensource.zalando.com/restful-api-guidelines/#terminology) i.e. running service, the link should use the same scheme + host + port combination as the original request URI. (This only comes into play when the API is available at e.g. multiple host names.)
```

chapters/hyper-media.adoc
```diff
@@ -158,6 +158,11 @@ when using features like embedding subresources. The primary advantage of
...
+
+*Motivation:* This avoids cases where authorization is configured for one host
```
tfrauenstein:
```suggestion
*Motivation:* This avoids inconsistent client experience situations, for instance, where authorization is configured for one host
```

tfrauenstein (commented):
thx! -- added two editorial change proposals

ePaul:
:+1:

tfrauenstein:
👍

Issue #864, opened 2026-04-29
Distinguishing the source of a 404 response


bocytko:
Recently, we observed an incident where an ingress had 0 routes and returned 404 for every single call. The calling applications reacted to the 404 as if it would have been a response from the applications behind the ingress.

The question is whether it's correct behavior that 404 is returned both for:
a) no matching route returned by the ingress (direct call)
b) application returned 404 explicitly because the entity was not found (proxied call)

We'd benefit from consistent guidelines how to handle the two mentioned scenarios.

The question is relevant for:
- https://opensource.zalando.com/restful-api-guidelines/#148
- https://opensource.zalando.com/restful-api-guidelines/#151

--

Envoy has clear distinction for these 404s: https://oneuptime.com/blog/post/2026-02-24-how-to-debug-envoy-route-not-found-issues-in-istio/view

Same can be done in the response by setting a header. This way, teams who interpret 404s can choose to react differently on ingress (direct) or application errors (proxied).

tfrauenstein:
Thanks for raising this issue -- it relates to a common architectural problem where you have application web-services implementing the business resources and additional infrastructure components like (ingress) router or load balancer. 

The HTTP status web standard does not clearly differentiate between application and infrastructure services. The system is basically a black-box from the client point of view. Purists argue that 404 (Not Found) is the standard response when the server (which includes the ingress router acting as a gateway) cannot find a representation for the requested URI. The client should not necessarily care if the error happened at the ingress layer or inside the app. If there is no route, the resource is "not found". However, the purist's view is not consistently adopted in the infrastructure industry: AFAIK most ingress providers (e.g. IGINX, HAProxy) return 404 and most cloud load balancer (e.g. AWS ALB, GCP LB) return 503 (Service Unavailable). 

Of course, the black-box view is not very helpful for clients that need to decide how to handle the error to provide the best user experience and care to restore service availability. They need to differentiate client errors (e.g. typo in the resource id - application resource is not available) and server systems error (e.g. route not defined). In our RESTful API thinking, 404 relates to the business resource provided by the application web-service and not to the API infrastructure state. Whenever we can define return codes of infrastructure components we should use 503 instead of 404. 

[...]

PR #869, opened 2026-07-01
feat: clarify end-to-end semantics of http status codes (#864)

tkrop:
This pull request addresses the issue raised in #864 by emphasizing the importance of the end-to-end principle in the HTTP protocol design and its implications for the RESTful API desgin.

I'm currently missing a bit the idea from where to link to this guideline.


chapters/http-status-codes-and-errors.adoc
```diff
@@ -58,6 +58,38 @@ responses:
           $ref: 'https://opensource.zalando.com/restful-api-guidelines/models/problem-1.0.1.yaml#/Problem'
 ----
 
+[#256]
+== {MUST} observe end-to-end semantics of HTTP status codes
+
+When implementing services and middleware (see {RFC-3234}[RFC-3234]), designers
```
tfrauenstein:
```suggestion
Implementers of middleware services (see {RFC-3234}[RFC-3234])
```

chapters/http-status-codes-and-errors.adoc
```diff
@@ -58,6 +58,38 @@ responses:
...
+and implementers must ensure that the end-to-end semantics of HTTP status codes
```
tfrauenstein:
```suggestion
must ensure that the end-to-end semantics of HTTP status codes
```

chapters/http-status-codes-and-errors.adoc
```diff
@@ -58,6 +58,38 @@ responses:
...
+are preserved (see also {RFC-3439}#section-2.1[RFC-3439 - Section 2.1 - The
+End-to-End Argument and Simplicity]). This means that any intermediate
```
tfrauenstein:
```suggestion
End-to-End Argument and Simplicity]). Any intermediate middleware
```

chapters/http-status-codes-and-errors.adoc
```diff
@@ -58,6 +58,38 @@ responses:
...
+component (e.g., sidecar, proxies, gateways, load balancers etc) must not alter
+the meaning of a status code returned by the service, nor must it return status
```
tfrauenstein:
```suggestion
the meaning of a status code returned by the application service, nor must it return status
```

chapters/http-status-codes-and-errors.adoc
```diff
@@ -58,6 +58,38 @@ responses:
...
+codes that could be interpreted as an authoritative answer on behalf of the
```
tfrauenstein:
```suggestion
codes that could be interpreted as an authoritative answer on behalf of and
```

chapters/http-status-codes-and-errors.adoc
```diff
@@ -58,6 +58,38 @@ responses:
...
+service without clear consent of the service.
```
tfrauenstein:
```suggestion
without clear consent of the application service.
```

chapters/http-status-codes-and-errors.adoc
```diff
@@ -58,6 +58,38 @@ responses:
...
+
+Particular, a middleware must not return a status code {404}, if the service
+would have returned a different status code, as this could mislead clients into
+thinking that the resource does not exist when it actually does. Without clear
```
tfrauenstein:
```suggestion
thinking that the resource does not exist when it actually does. Without
```

chapters/http-status-codes-and-errors.adoc
```diff
@@ -58,6 +58,38 @@ responses:
...
+contract and explicit indication by the service a middleware must always return
```
tfrauenstein:
```suggestion
explicit indication by the application service a middleware must always return
```

chapters/http-status-codes-and-errors.adoc
```diff
@@ -58,6 +58,38 @@ responses:
...
+a temporary, non-authoritative status code, such as:
```
tfrauenstein:
```suggestion
a temporary, non-authoritative server-side status code, such as:
```

chapters/http-status-codes-and-errors.adoc
```diff
@@ -58,6 +58,38 @@ responses:
...
+
+* **{502} — Bad Gateway**,
+* **{503} — Service Unavailable**, or
+* **{504} — Gateway Timeout**.
+
+This also applies to coupled middleware components, e.g. sidecars deployed with
+the service, whenever they pose the risk of returning misleading status codes
+to clients.
```
tfrauenstein:
```suggestion
For instance, a Kubernetes ingress router (like Skipper) must be configured to return status code {503} 
instead of {404} in case no route for the application service endpoint is available.
```

chapters/http-status-codes-and-errors.adoc
```diff
@@ -58,6 +58,38 @@ responses:
...
+
+**Note:** It is important for API designers and consumers to understand that
```
tfrauenstein:
```suggestion
**Note:** It is important for API designers and consumers to rely on
```

chapters/http-status-codes-and-errors.adoc
```diff
@@ -58,6 +58,38 @@ responses:
...
+the end-to-end semantics of HTTP status codes is a fundamental part of the API
```
tfrauenstein:
```suggestion
the end-to-end semantics of HTTP status codes being a fundamental part of the API contract.
```

chapters/http-status-codes-and-errors.adoc
```diff
@@ -58,6 +58,38 @@ responses:
...
+contract that consumers rely on. Therefore, any middleware that is deployed in
+the path of an API must comply with this guideline and must not violate the
+end-to-end semantics of HTTP status codes.
```
tfrauenstein:
```suggestion
There is the high risk that status codes mislead API consumers leading to incidents if the middleware does not comply to this guideline and not preserve the end-to-end semantics.
```

chapters/http-status-codes-and-errors.adoc
```diff
@@ -58,6 +58,38 @@ responses:
           $ref: 'https://opensource.zalando.com/restful-api-guidelines/models/problem-1.0.1.yaml#/Problem'
 ----
 
+[#256]
+== {MUST} observe end-to-end semantics of HTTP status codes
```
tfrauenstein:
```suggestion
== {MUST} preserve the end-to-end semantics of HTTP status codes
```

tfrauenstein (commented):
Thanks for the PR!
I provided a couple of editorial change proposals as discussed in the API Guild JF.
Let's also add a general hint and link as intro to the [Client side error codes](https://opensource.zalando.com/restful-api-guidelines/#client-side-error-codes) section:

API consumers can rely on the client-side code semantics being provided by the application since middleware components <<256>> and return server-side codes. (This includes e.g. authorization errors possibly being written by the middleware on behalf of the application service.)

chapters/http-status-codes-and-errors.adoc
```diff
@@ -217,6 +249,13 @@ being implicitly kept in some state based on a previous redirect. See also
 [[client-side-error-codes]]
 === Client side error codes
 
+Clients can safely assume that {4xx} status codes are always authoritative and
+comply with the end-to-end principle (see <<256>>). While a middleware is not
+can authenticate a request and return {401} or {403} on behalf of the service,
+it must never return a misleading status code, e.g. {404} or {410}, if the
+service is not configured or not responding.
```
ePaul:
```suggestion
Clients can safely assume that {4xx} status codes are always authoritative and
comply with the end-to-end principle (see <<256>>). While a middleware
can authenticate a request and return {401} or {403} on behalf of the service,
it must never return a misleading status code, e.g. {404} or {410}, if the
service is not configured or not responding.
```

Alternatively, the "is not" phrase needs to be completed and a new one started, but I don't know how.

ePaul: Fun fact: While trying to submit this comment, I got a "not found" response from Github as I was logged in with the enterprise user account. Almost what this paragraph wants to forbid.

chapters/http-status-codes-and-errors.adoc
```diff
@@ -58,6 +58,38 @@ responses:
...
+When implementing services and middleware (see {RFC-3234}[RFC-3234]), designers
+and implementers must ensure that the end-to-end semantics of HTTP status codes
+are preserved (see also {RFC-3439}#section-2.1[RFC-3439 - Section 2.1 - The
+End-to-End Argument and Simplicity]). This means that any intermediate
+component (e.g., sidecar, proxies, gateways, load balancers etc) must not alter
+the meaning of a status code returned by the service, nor must it return status
+codes that could be interpreted as an authoritative answer on behalf of the
+service without clear consent of the service.
+
+Particular, a middleware must not return a status code {404}, if the service
```
ePaul:
```suggestion
In particular, a middleware must not return a status code {404}, if the service
```

chapters/http-status-codes-and-errors.adoc
tkrop: I take the liberty to reformulate this sentence.

tfrauenstein:
👍

tfrauenstein:
@bocytko and @szuecs We aligned on this PR addressing the issue [Distinguishing the source of a 404 response](https://github.com/zalando/restful-api-guidelines/issues/864#top). Change proposals / comments are appreciated, thx.

tkrop:
👍

szuecs:
> @bocytko and @szuecs We aligned on this PR addressing the issue [Distinguishing the source of a 404 response](https://github.com/zalando/restful-api-guidelines/issues/864#top). Change proposals / comments are appreciated, thx.

Of course I disagree on the outcome, but that was clear. 
I don't think we will change production like this, because it's a too big change. 
And finally there is OTel tracing as tool so you can see where the response is coming from, not sure if your PR and issue even makes sense with this kind of visibility technology.
A 404 from the proxy will never have a backend call, so it's clearly visible.

szuecs:
There is no agreement at hyperscalers, but no one returns a 5xx on a clear L7 routing error:

```
# amazon returns a 403 for github.com
% curl -v --connect-to www.github.com:443:www.amazon.com:443 https://www.github.com/ -k G -E '(Host:|HTTP/|< Location:)'
> GET / HTTP/1.1
> Host: www.github.com
< HTTP/1.1 403 Forbidden

# amazon returns a 400 for google.com
% curl -v --connect-to www.google.com:443:www.amazon.com:443 https://www.google.com/ -k  G -E '(Host:|HTTP/|< Location:)'     
> GET / HTTP/1.1
> Host: www.google.com
< HTTP/1.1 400 Bad Request

# google returns a 404 as we do today
% curl -v --connect-to www.github.com:443:www.google.com:443 https://www.github.com/ -k G -E '(Host:|HTTP/|< Location:)'     
> GET / HTTP/1.1
> Host: www.github.com
< HTTP/1.1 404 Not Found

# github.com returns an old permanent redirect to itself
% curl -v --connect-to www.google.com:443:www.github.com:443 https://www.google.com/ -k G -E '(Host:|HTTP/|< Location:)'     > GET / HTTP/1.1
> Host: www.google.com
< HTTP/1.1 301 Moved Permanently
< Location: https://github.com/

# facbook returns 400
% curl -v --connect-to www.google.com:443:www.facebook.com:443 https://www.google.com/ -k G -E '(Host:|HTTP/|< Location:)'    
> GET / HTTP/1.1
> Host: www.google.com
< HTTP/1.1 400 Bad Request

# netflix returns 403
% curl -v --connect-to www.google.com:443:www.netflix.com:443 https://www.google.com/ -k G -E '(Host:|HTTP/|< Location:)'     
> GET / HTTP/1.1
> Host: www.google.com
< HTTP/1.1 403 Forbidden

```

bocytko:
Thanks for the PR. Do you have an overview which ingress proxies or http routers are compliant with your proposed phrasing of the guidelines?

We should not put up guidelines that are impractical and cannot be easily configured in existing software.

tfrauenstein:
a) We have not done a systematic analysis of all ingress proxies and routers used in our environment.
But Skipper is by far the most used, and its config could be changed in favor of the e2e argument.
At minimum we should take care that error source information is provided via a defined response header.

b) I think the guideline is helpful even if it is not realistic to ensure it in all our infrastructure. However, if we keep it, we should add a hint along the following lines:

Warning: Some middleware components in your environment — such as ingress proxies or routers (e.g. Skipper) — may not be configured according to the e2e argument and e.g. return the 404 status code instead of a 503 when routes are missing. In these situations API clients cannot rely on 404 representing the application service state and have to inspect middleware response headers or log information to identify the error cause.

bocytko:
> a) We have not done a systematic analysis of all ingress proxies and routers used in our environment.

@tfrauenstein Even our environment would be too small of a footprint to validate the proposed rule.
See below for an LLM-assisted research if interested.

> But Skipper is by far the most used, and its config could be changed in favor of the e2e argument.

Based on what @szuecs said as the maintainer in the thread, there is more work needed to get an alignment (also, there are simpler means to address the behavior we observed in the incident through validation during rollout of new versions).

The tricky part is the 404 being returned for a scenario where a route already existed before. There are better means to signal routes that have been superseded (410 Gone or redirects). An empty routing table scenario for existing deployments could be considered erroneous by default.

> And finally there is OTel tracing as tool so you can see where the response is coming from, not sure if your PR and issue even makes sense with this kind of visibility technology.
> A 404 from the proxy will never have a backend call, so it's clearly visible.

@szuecs It's clear that manual inspection during an incident can reveal which layer returns the 404. However, you first need to know that you actually have an incident. A backend service experiencing the 404 will not magically start inspecting telemetry to check if it received a 404 from the proxy or from the backend. Inferring from response time is also error-prone. Example clients that can take decisions upon 404:
- a data synchronization backend treating 404 as deleted, resulting in data loss client-side
- a web crawler treating the target as non-existent and delisting it from its index

For clients relying on a route being consistently configured in the ingress, the failure will be either catastrophic or silent and potentially undetected for longer time. 404 for misspelled routes or non-existing backends is not controversial at all.

---

@tkrop @tfrauenstein I would suggest to soften the guideline to SHOULD and include clear recommendation for infrastructure components to be configured in a way that explicitly handles the 'no route' behavior with short description why. Many HTTP proxies support configuring catch-all routes that can override the response code with 503. I believe such route can also be easily configured in skipper (correct me if I'm wrong @szuecs). We can also acknowledge other mechanisms, like Envoy making 404 explicit with information that it was a *no route matched* scenario as mentioned in the original ticket or Kubernetes Gateway API explicitly demanding 404 as a strict standard.

The SHOULD framing also supports cases where 3rd party systems are integrated as this behavior cannot be typically influenced by a single client. Therefore this consideration must be solved client-side where clients are to be careful on how they interpret the 404, accounting for the inconsistent configurations that @szuecs pointed out above.

If we reach agreement with @szuecs & team to configure skipper in a way that enforces the 503 in Zalando clusters as catch-all route, a case for an upgrade to MUST can be considered. I believe this is a pragmatic approach allowing us to have guidance on the nuanced 404 case in the guidelines, which is overall helpful.

---

<details>
<summary>
Summary on http proxy behavior done by Opus 4.8 (based on Research output that crawled the proxy docs)
</summary>

Summary prompt:

> Imagine you have an ingress and its new version is deployed. The previously existing routes are removed, even though the target backends are up. Therefore one ingress pod with old version returns responses from the backends while the new version detects missing route and reacts appropriately. Outline how the analyzed technologies will react. Outline compliance with the proposed rule. Show output as a table.

<html><head></head><body><p>The scenario isolates exactly the case the guideline cares about most: the backend is healthy, but a config rollout has (wrongly or transiently) removed the route. The old pod still has the route and serves 200s from the backend; the new pod has no route and must decide what to return. That decision is precisely where "don't fake a 404 for a service that's actually up" bites.</p>
<p>One framing note before the table: in this scenario there is <strong>no invalid backendRef</strong> — the backend is up and reachable. The new pod simply has <em>no matching route</em> for the request. So this is squarely the <strong>no-route</strong> bucket, not the bad-backend or backend-down bucket. That matters because, as we established, the no-route bucket is where almost every proxy returns 404 by design — which is the <em>least</em> compliant case for the proposed rule.</p>
<p>The "proposed rule" I'm scoring against is the two-tier reframing: <em>reached the backend, no good response → 502/503/504; never found a route → 404 is acceptable, but return 503 if route-flapping during deploys is a real risk.</em> Since this scenario <strong>is</strong> that risk materializing, I'm scoring compliance strictly — i.e. "does the tool let you make the new pod emit a temporary 5xx instead of a misleading 404 during the rollout gap?"Here's how each technology behaves in your rollout-gap scenario — old pod still has the route and serves the backend's 200s, new pod has no matching route and must respond.</p>
<p>The key columns: what the new (routeless) pod returns, whether that misleads the client into thinking the resource is gone, and whether the tool <em>can</em> be made to emit a temporary 5xx instead — which is what compliance requires here.</p>

Technology | New pod's response (no route) | Misleading? | Can emit 5xx for no-route? | Compliance with proposed rule
-- | -- | -- | -- | --
NGINX (reverse proxy/LB) | Depends on config — often default_server returning 404, or connection reset | Yes, if 404 | Yes — return 503 in a catch-all server/location | High — trivial to configure a 503 default
ingress-nginx | 404 from the default backend | Yes | Yes — replace default backend with a 503-returning one | High (with config) — the guideline's own example case
HAProxy | 503 if no default_backend matches the ACL | No | Native — already 503 | High (out of the box)
Envoy | 404 (NR — no route configured) | Yes | Yes — catch-all vhost with direct_response: 503 | High (with config)
Traefik | 404 (no router matches) | Yes | Yes — documented low-priority catch-all router → serverless service → 503 | High (with config) — officially supported pattern
Kong | 404 "no Route matched" | Yes | Yes — custom plugin rewrites 404→503 | Medium-High — needs a plugin
Kubernetes Gateway API | 404 (no route match — conformance-mandated) | Yes | Only via response-override/direct-response layered on top | Low-Medium — spec MUST is 404; overriding fights conformance
AWS ALB | Default action fixed-response (whatever you set — 404 or 503) | Only if set to 404 | Yes — set the listener default action to fixed-response 503 | High (with config)
AWS API Gateway | 403 "Missing Authentication Token" | Yes (worse — obscures existence) | Via Gateway Responses remap, but 503 is semantically odd for "unauthenticated" | Low-Medium
GCP LB / GKE Ingress | 404 reaches the default backend | Yes | Partly — URL-map default routing to a 503 backend | Medium
Azure App Gateway | 404 (no listener match) | Yes | Limited — custom error pages, less flexible than software proxies | Medium
Cloudflare | N/A — route-flapping is an origin-config concern, not Cloudflare's | — | Origin errors already 5xx (52x) | N/A for this scenario
Apache mod_proxy | 404 from local handler (no ProxyPass match) | Yes | Yes — ErrorDocument 404 → 503, or a catch-all ProxyPass | High (with config)
Spring Cloud Gateway | 404 (no route predicate matches) | Yes | Yes — a catch-all route to a serverless/lb:// target yields 503; error routes remap | High — 503 is already default for the no-instance sibling case
Istio (Envoy) | 404 (NR) | Yes | Yes — catch-all VirtualService with direct_response: 503, or EnvoyFilter | High (with config)


<h2>The insight this scenario exposes</h2>
<p>The scenario is engineered to hit the <em>worst</em> bucket for the proposed rule. There's no dead backend and no invalid <code>backendRef</code> — the backend is up, and the <em>only</em> thing missing is the route on the new pod. That is the pure no-route case, and the no-route case is exactly where nearly every proxy returns <strong>404 by design</strong>. So out of the box, most of these technologies <strong>fail</strong> the rule during the rollout gap: a client (or crawler, or downstream service) hitting the new pod sees a 404 for a resource that demonstrably still exists on the old pod. For a search crawler or a cache, that 404 can be acted on as "permanently gone" — the exact data-integrity hazard the guideline warns about.</p>
<p>The good news is that "can be made compliant" is <strong>High</strong> for most: HAProxy is already compliant, and NGINX, ingress-nginx, Envoy/Istio, Traefik, ALB, Apache, and Spring Cloud Gateway all support a documented catch-all → 503 pattern. The genuinely hard cases are the two governed by external constraints rather than config: <strong>Gateway API</strong> (404 is a conformance MUST, so a compliant 503 must be layered on via response-override and technically deviates from spec behavior) and <strong>AWS API Gateway</strong> (defaults to 403, and remapping to 503 is semantically strange).</p>
<p>There's also a subtlety the table can't fully capture: during this specific transient, a routeless pod emitting <strong>503 with <code>Retry-After</code></strong> is <em>more</em> correct than 404, because the condition genuinely is temporary — the route will exist again once the rollout settles or rolls back. This is the one no-route situation where the guideline's "prefer 503" instinct is unambiguously right, precisely because the missing route is a deploy artifact, not a real absence. That argues for the operational refinement from the reframing: enable the 503 catch-all specifically in environments where route-flapping during deploys is a known pattern — which is every rolling Kubernetes ingress upgrade.</p></body></html>
<details>

chapters/http-status-codes-and-errors.adoc
```diff
@@ -58,6 +58,40 @@ responses:
           $ref: 'https://opensource.zalando.com/restful-api-guidelines/models/problem-1.0.1.yaml#/Problem'
 ----
 
+[#256]
+== {MUST} preserve end-to-end semantics of HTTP status codes
```
bocytko, line 62: I suggest to downgrade to SHOULD. See comment for more context.

chapters/http-status-codes-and-errors.adoc
```diff
@@ -58,6 +58,40 @@ responses:
...
+{RFC-3439}#section-2.1[RFC-3439 - Section 2.1 - The End-to-End Argument and
+Simplicity]). Any intermediate middleware component (e.g., sidecar, proxies,
+gateways, load balancers etc) must not alter the meaning of a status code
+returned by the application service, nor must it return status codes that
+could be interpreted as an authoritative answer on behalf and without clear
+consent of the application service.
+
+In particular, a middleware must not return a status code {404}, if the
+service would have returned a different status code, as this could mislead
+clients into thinking that the resource does not exist when it actually does.
```
bocytko, line 75: The thing is that this never happens by design, otherwise you'd have a malicious proxy.

I would make a case here for the scenario where a 404 from the proxy cannot be distinguished by a client as to whether it's a (temporary) routing error due to missing routing table or an actual response from the backend. Then say, that catch-all routes can be configured in proxies to return 503 if routes are not found. This way, one ensures that 404s are only returned from the backends, never decided directly by the proxy.

chapters/http-status-codes-and-errors.adoc
```diff
@@ -217,6 +251,13 @@ being implicitly kept in some state based on a previous redirect. See also
 [[client-side-error-codes]]
 === Client side error codes
 
+Clients can safely assume that {4xx} status codes are always authoritative and
```
bocytko, line 254: Due to 3rd party systems being hosted in infrastructure that cannot be influenced by a single client, I would be against advising that one can safely rely on 404. This is only valid in infrastructure that is fully controlled by a single actor.

tfrauenstein:
@bocytko Thank you for the detailed feedback. I'm ok with anticipating infrastructure reality by keeping the guideline, but as SHOULD infrastructure guideline and supplemented with config hints to ensure routing errors being reported as 503 (e.g. via catch-all routes). Additionally, I propose to add a guideline for the clients to be careful with 404. It is not very satisfactory for the clients and would look roughly as follows:

Guideline - Clients must be careful with how to interpret 404: Infrastructure components such as ingress proxies or routers (e.g. Skipper) may be configured inconsistently not following the e2e argument and return 404 status code instead of a 503 when routes are missing. Clients cannot generally rely on 404 representing the backend application state, and have to exclude 404 infrastructure routing error source e.g. via tests and inspection of infrastructure response headers or log information.


I'm curious about @szuecs responses on how we may be more concrete and improve the situation for the Skipper clients.

tfrauenstein:
I provided an update as discussed, though for the time being without specific Skipper notes from Sandor, and created a PR [UPDATE - feat: clarify end-to-end semantics of http status codes- #872](https://github.com/zalando/restful-api-guidelines/pull/872) as update of this original PR.

tkrop:
@szuecs with respect to the above - I'm not sure that this proves anything with respect to the status code case we are talking about: The request is in the first place *invalid* by using an invalid `Host`-header when requesting a service/gateway. In this  case it is with out question expected that this is marked as a user error and returns `4xx` - even if this decision is done by a middle-box.

A real world test, e.g. for `nginx`, looks as follows:
```bash
cat >nginx.conf <<'EOF'
events {}
http {
  server {
    listen 8080;
    location /refused {
      proxy_connect_timeout 1s;
      proxy_read_timeout 1s;
      proxy_send_timeout 1s;
      proxy_pass http://127.0.0.1:65534;
    }
    location /timeout {
      proxy_connect_timeout 1s;
      proxy_read_timeout 1s;
      proxy_send_timeout 1s;
      proxy_pass http://10.255.255.1:81;
    }
  }
}
EOF

docker rm -f middlebox-nginx-test >/dev/null 2>&1 || true
docker run -d --name middlebox-nginx-test -p 8080:8080 \
  -v $(pwd)/nginx.conf:/etc/nginx/nginx.conf:ro \
  nginx:1.27-alpine
  
sleep 3

curl -sS -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8080/refused
curl -sS -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8080/timeout
curl -sS -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8080/missing
```

or `haproxy`:

```bash
cat > "haproxy.cfg" <<'EOF'
global
    maxconn 100

defaults
    mode http
    timeout connect 1000ms
    timeout client  5000ms
    timeout server  5000ms

frontend fe_http
    bind *:8080
    acl is_refused path -i /refused
    acl is_timeout path -i /timeout
    use_backend be_refused if is_refused
    use_backend be_timeout if is_timeout

backend be_refused
    server dead_refused 127.0.0.1:65534 check

backend be_timeout
    server dead_timeout 10.255.255.1:81 check
EOF

docker rm -f middlebox-haproxy-test >/dev/null 2>&1 || true
docker run -d --name middlebox-haproxy-test -p 8081:8080 \
    -v $(pwd)/haproxy.cfg:/usr/local/etc/haproxy/haproxy.cfg:ro \
    haproxy:2.9-alpine

sleep 3

curl -sS -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8081/refused
curl -sS -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8081/timeout
curl -sS -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8081/missing
```

and `skipper`:

```bash
cat > routes.eskip <<'EOF'
r_refused: Path("/refused") -> "http://127.0.0.1:65534";
r_timeout: Path("/timeout") -> "http://10.255.255.1:81";
EOF

docker rm -f middlebox-skipper-test >/dev/null 2>&1 || true
docker run -d --name middlebox-skipper-test -p 9090:9090 \
  -v $(pwd)/routes.eskip:/etc/skipper/routes.eskip:ro \
  ghcr.io/zalando/skipper:latest \
  skipper -address :9090 -routes-file /etc/skipper/routes.eskip

sleep 3

curl -sS -o /dev/null -w "%{http_code}\n" http://127.0.0.1:9090/refused
curl -sS -o /dev/null -w "%{http_code}\n" --max-time 40 http://127.0.0.1:9090/timeout
curl -sS -o /dev/null -w "%{http_code}\n" http://127.0.0.1:9090/missing
```

Nearly all return `5xx` by default, with the exception of `nginx` and `skipper`, that by default need manual adjustment to return `503` on non-configured end-points. Similar `5xx` is also the default behavior for all major load-balancers, i.e. AWS Application Load Balancer,
Google Cloud HTTP(S) Load Balancer, F5 BIG-IP, and Citrix ADC / NetScaler to my best knowledge. Only middleboxes, that usually are installed close or as part of the end-systems like `nginx`, `envoy`, etc. are usually returning `404` (by default) assuming that misconfiguration of these is part of the service.

As a consequence, I would say it is the standard to expect middleboxes and loadbalancers in the role of a router by default to return `5xx` to comply with the `end-to-end` principle of the HTTP protocol, while it is possible that middleboxes in an authorized service gateway roles can return also `4xx`. We have already accounted for exactly this differences in the current formulation of the `must` rule by adding this as an exception, to make the clear as strong and understandable as possible.

Besides, I would be very much interested in knowing what we would actually gain by changing the rule to `should`. Is the consequence that we are allowing our routing infrastructure to stay broken returning `404` until we experience further incidents? Do we intend to require all clients to implement an additional source check?

szuecs:
@bocytko the change for skipper is trivial, yes. The impact is that all CBOs will have a higher error rate if scrapers find their target. Right now we just ignore 404s (no route found cases) in skipper-ingress logs and alerts because it is just too noisy for everyone. If we do change it to 503 then everyone will have pages all over the place unpredictable and I will have the meetings and not @tfrauenstein nor @tkrop so I think I should ignore this request even if we change it to SHOULD for the good of every team that have alerting on 5xx rates.

@tkrop it's funny that you show 3 example and 2 out of 3 are showing the 404 and rest is an "assumption" of experience without any data. AWS ALB highly depends on the configuration and in our case all requests will be passed to skipper-ingress. Of course you can change by rules, but that's similar to create a specific route with status() and shunt. 

f5 does a 404 for route not found, see https://my.f5.com/manage/s/article/K000146828

